PR #172, opened 2014-07-30
Provide ``to_common`` method

catherinedevlin:
Feel free to ignore if this convenience function doesn't fit in your vision, but it's useful for me.

I wasn't sure where (or whether) it would fit into your docs structure, so I didn't write any.  At least it's got a docstring!


hgrecco:
I am closing this for now. If you want to pursue, please open an issue providing a use case.

PR #222, opened 2015-01-13
Parse unicode micro prefix and angstrom 

ryanpdwyer:
Allows the unit registry to recognize `µ` as micro, and `Å` as Angstrom in parsing from strings. Similar to #117. I'd also propose changing the shortened pretty print format (`{:P~}`) for ångström to `Å`, but I can do that in a separate PR.

### DIFF
--- a/pint/default_en.txt
+++ b/pint/default_en.txt
@@ -138,7 +138,7 @@ byte = 8 * bit = Bo = octet
 baud = bit / second = Bd = bps
 
 # Length
-angstrom = 1e-10 * meter = ångström = Å
+angstrom = 1e-10 * meter = Å = ångström = Å
 inch = 2.54 * centimeter = in = international_inch = inches = international_inches 
 foot = 12 * inch = ft = international_foot = feet = international_feet
 mile = 5280 * foot = mi = international_mile
--- a/pint/testsuite/test_issues.py
+++ b/pint/testsuite/test_issues.py
@@ -9,6 +9,7 @@ from pint.unit import UnitsContainer
 from pint.util import ParserHelper
 
 from pint.compat import np, unittest, long_type
+from pint.errors import UndefinedUnitError
 from pint.testsuite import QuantityTestCase, helpers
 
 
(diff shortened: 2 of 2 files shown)


hgrecco:
Great! Do the pretty print in another PR. I will merge them both into master soon.

ryanpdwyer:
So I looked into it a little more, and I think it is two separate issues; one the two different unicode symbols for the angstrom symbol, and one a problem with parsing unicode units from strings in python2. On python2.7, the error looks like,


    >>> ureg('µm')
      File "<string>", line 1
        µQ_ (1 ,U_ (meter=1 ))
        ^
    SyntaxError: invalid syntax

I think it comes from the `eval` call in `pint/util.py` here:

        ret = eval(untokenize(result),
                   {'__builtins__': None},
                   {'L_': cls.from_word})

which is trying to `eval` something with unicode in it.

I added tests for this, which should show failures on python2 in Travis.

hgrecco:
I think that the root of the problem is actually earlier. Taking the important parts from `util.py`
(In Python 2)

```python
>>> import tokenize
>>> from StringIO import StringIO
>>> tokenizer = lambda input_string: tokenize.generate_tokens(StringIO(input_string).readline)
```

The result is:
```python
>>> list(tokenizer('umeter'))
[(1, 'umeter', (1, 0), (1, 6), 'umeter'), (0, '', (2, 0), (2, 0), '')]
>>> list(tokenizer(u'µmeter'))
[(52, u'\xb5', (1, 0), (1, 1), u'\xb5meter'), (1, u'meter', (1, 1), (1, 6), u'\xb5meter'), (0, '', (2, 0), (2, 0), '')]
```

The micro prefix in unicode is:
- interpreted as an ERRORTOKEN (code 52)
- splitted from the unit name.

May be we can make a workaround in the parserhelper or write another tokenizer.

hgrecco:
I think we can generate a new tokenizer that does the job by improving Python 2 `tokenize.generate_tokens`.

Issue #151, opened 2014-05-29
Support for normalization


infraredgirl:
It would be nice to have support for conversion of values to the human readable/normalized form. Let me explain what I mean by giving a few examples.

```python
>>> normalize(1, 'B')
(1, 'B')
>>> normalize(1024, 'B')
(1, 'KB')
>>> normalize(10*1024**3, 'B')
(10, 'GB')
>>> normalize(0.5, 'MB')
(512, 'KB')
>>> normalize(60, 's')
(1, 'min')
>>> normalize(0.5, 'day')
(12, 'hr')
```
You get the idea. I think this kind of utility function would be extremely useful. (If this is already supported, and I have just missed it, please do let me know.) Thanks for considering this!

ryanpdwyer:
Here is a functional implementation for the standard SI prefixes.

https://gist.github.com/ryanpdwyer/bcb4dc3d2a2a6b6734bb

hgrecco:
I was toying with the idea to get a compact representation in #15. It basically used and OrthogonalMatchingPursuit to find the shortest combination of units (base and derived) with the same dimensionality of the input units. It was working for some cases, but not giving good results in some cases. 

I like @ryanpdwyer because it is like a *guided* compact methods. And then we might work on the infer_base_unit in a second step. Lets merge it **after** we finish with the _systems branch and #202.

hgrecco:
I am closing this now. Feel free to reopen.

PR #224, opened 2015-01-13
Compact

ryanpdwyer:
- Adds a `to_compact` method to Quantity, as discussed in #151.
- Adds `infer_base_units`, a helper method used by `to_compact`, to `utils`.
- Adds tests of `to_compact`

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -274,8 +274,8 @@ class _Quantity(SharedRegistryObject):
 
         magnitude = q_base.magnitude
         # Only changes the prefix on the first unit in the UnitContainer
-        unit_str = q_base._units.items()[0][0]
-        unit_power = q_base._units.items()[0][1]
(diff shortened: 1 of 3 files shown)


hgrecco:
While I am definitely **+1** on adding the `compact` method that you propose I am not happy about hardcoding the prefixes (and in 2 places). You can access all the prefixes from the registry. To get the unprefixed name we could modify get_name to have an extra parameter (`strip_prefix`?) or similar defaulting to the current behavior.

This is not just a aesthetic choice. Is part of Pint design philosophy to avoid hardcoded values and making everything part of the registry. 

Additionally, this will also help when we try implement this compact for non SI units, but this are future plans when `systems` are fully implemented.

hgrecco:
Sorry but only now this came again to my attention. It looks really good. I will merge it next week.

hgrecco:
thanks!

PR #229, opened 2015-02-04
Add "hectare" as a supported unit of area

edschofield:
The usual "hecto" prefix for 100* (hectopascal, hectometre, etc.) is
shortened to "hect" for "hectares". This seems to be a unique case:
https://en.wikipedia.org/wiki/Hecto-.

Adding "hectare" as an explicit unit of area therefore seems to be cleaner than
adding "hect" as a supported prefix for all units. Adding "hect" instead would
cause Pint to recognize many strange new Quantities such as ureg.hectsecond.


hgrecco:
Thanks a lot!

PR #230, opened 2015-02-05
Add photometric units : lumen & lux

janpipek:
Hello,

I just thought adding these two units to the default registry might be useful. However, I don't know if the "angular part" in the dimensions provides some problem. 

Cheers,
Jan


hgrecco:
Everything related to angle is problematic as they are dimensionless, but it should work fine as long as you do a few things explicitely. Thanks for the PR.

janpipek:
Glad I could help.

Issue #160, opened 2014-06-15
UnitsContainer print


romarro:
Hi I want to print only the units of a quantity (for a label in a plot). But it doesn't print the abbreviated units  even if I am formatting it with ~
here is the code:
>>> from __init__ import um,Q_
>>> x=125.3*um['l/s']
>>> print x
125.3 l/s
>>> x.units
<UnitsContainer({'liter': 1.0, 'second': -1.0})>
>>> print x.units
liter / second
>>> print '{:~}'.format(x.units)
liter / second
>>> print '{:~L}'.format(x.units)
\frac{liter}{second}

But with latex formatting works
is there a workaround?

Thanks for help

hgrecco:
I think the current is fixed in develop. Please reopen if you find any problem.

Issue #226, opened 2015-01-24
New tokenizer and maybe evaluator


hgrecco:
This is an issue to discuss the new tokenizer and evaluator. The main goal is to make unicode indentifier accepted in Python 2. In 8756ed727963f4115cdee1b3a24601bba3d52e4c I have copied the Python 3.4 tokenize module to pint and tweaked to make it work in Python 2.7. In b32c2ec3bbfe4b777871396f2ce0c89fe9ae9ce0 I have changed the `isindentifier` function to allow some unicode characters in Python 2. 

Pending:
1.- Accepting identifiers is more restrictive than accepting probable unit names. For example (`"`, `'` and `%`) are not valid identifiers but they are valid unit names. We might need to tweak even more the tokenizer. What to do?
2.- Removing some redundancy. Because the way we copied from Python3.4, we now change the string to bytes, detect the encoding and then change it back to a unicode string. This needs to be discussed.
3.- Do we need a custom evaluator?
4.- Performance

cheezman34:
Regarding a custom evaluator & security:

As far as I can tell, the current use of eval only happens after the input has been modified during tokenization.  This *seems* like it should make eval safe to use on unverified input, but I'm not entirely convinced a sufficiently clever person could't figure a way around it.  Custom eval would, if nothing else, help provide some peace of mind for some of us.

I might be willing to take a crack at it if the requirements are concise.

What all would a custom evaluator have to do?
1) Parse quantities & units
2) Perform math operations on units
3) Compare units?
4) Anything else?

hgrecco:
Evaluating an expression can be divided in two different aspects:
1.- Evaluating a multiplicative expressions: This just needs to understand product, division, power, numbers and units as strings, returning a ParserHelper-like instance. Notices that this is registry independent. 
2.- Evaluating an additive expression: This is more complex as it needs to know how to convert one unit to another. This requires the registry.

For most internal operations only **1** is required. To allow a generic unit calculator we also need the second..

hgrecco:
We have a new parser!

nfearnley:
I'm still trying to parse something like "5'" or "6\"" or "5'6\"" as per #192. Pint doesn't seem to parse the 5'2" units properly, and when I attempt to add it as a custom unit, it just throws an error. Any luck handling this?

Issue #215, opened 2014-12-26
Systems: A feature to change base units


hgrecco:
See **Version 2** below

----

### Introduction

In Pint every unit is defined in terms of other units. We call these **derived units**. Notable exception are **base units** which are linked to dimensions (enclosed in square brackets `[]`). For example, in the [units definition file](https://github.com/hgrecco/pint/blob/master/pint/default_en.txt):

```
angstrom = 1e-10 * meter # derived unit
meter = [length] # base unit
```

This also creates the concept of **base dimensions**, i.e. the dimensions linked to base units (e.g. `[length]`, `[time]`) and **derived dimensions** (e.g. `[speed] = [length] / [time]`) 

Of course the system could be changed easily right now changing the definition file. 

Some possibilities are:

A redefinition of base units with the default base dimensions, for example:
```
meter = 1e10 * angstrom # derived unit
angstrom = [length] # base unit
```

A redefinition of base units with other base dimensions, for example:
```
hour = [time]  # base unit
mph = [speed] # base unit
mile = mph * hour  # derived unit
```

This creates a Directed Acyclic Graph relating each unit to one or more *basic* ones until base units are reached. In current version of Pint (0.6), base units are unprefixed SI units.

Additionally, all units in Pint are exposed to the user in the same way. There is no way to indicate that a certain unit belongs to certain family. For example, a user might want to obtain all compatible metric units but not imperial.

All together, base units, base dimensions and groups of units are what we call a **system**. 

### Goal

Being able to define and use a new system based on the definition file or programmatically as a change to an existing system.

### Defining a new system

Defining a new system involves substituting one or more base units with others.

```
@system <name> = <rule 1>, <rule 2>, ..., <rule N>
```

```python
>>> ureg.define_system(<name>, <rule 1>, <rule 2>, ..., <rule N>)
```

A rule is a string with two parts separated by a semicolon: `<new unit>:<old unit>`. The first part indicates a unit to be considered a base unit. The second is a unit not to be used as a base unit any more.

The `<old unit>` part is unnecessary if `<new unit>` has the same dimension of the base unit to be replaced. In other words: `inch` is interpreted as `inch:meter`.

A `ValueError` is raised if:
- two or more units with the same dimension are provided.
- the old unit part is not provided and new unit is not a base unit.
- the new unit and old unit are unrelated.

This syntax allows changing the system keeping or changing the same base dimensions.

### Using a named system

```
>>> ureg.set_system(<name>)
```

[...]

Issue #197, opened 2014-10-14
Rules in wrapping functions


hgrecco:
It would be convenient to be able to specify rules for units such as given

```
def func(a, b):
    # check that a/b is in seconds.
```

Questions
1.- Define API. Do it in `wraps` or create a specialized function (eg. `wraps_rules`)
2.- What shall we do with a/b
2.a.- Check
2.b.- insert the value in the function. Where?
2.c.- change the value of `b` so that `a/b` is in seconds. Does this scale for more complicated rules?

PR #231, opened 2015-02-05
Introduce 'u' to format only a Quantity's units

jpgrayson:
It is currently not possible to correctly format units (UnitsContainer
objects) using the '~' format spec.

    >>> format(q, '~')
    '17.0 km / s ** 2'
    >>> format(q.units, '~')
    'kilometer / second ** 2'

Notice how the units are different when formatted on their own.

In order for units to be properly formatted with '~', the UnitsRegistry
must be available so that the long names in the units container can be
mapped to their short, symbolic names. I.e. 'kilo' -> 'k' and 'meter' ->
'm'. A key problem is that UnitsContainer objects do not have access to
the UnitRegistry. When quantities are formatted, we cheat (a little) and
use the quantity's UnitRegistry (_REGISTRY) to perform the _get_symbol()
mapping in the case of '~'.

So we introduce 'u' as a formatting spec for Quantity that causes only
the unit, and not the magnitude, to be returned from format().

    >>> format(q, '~u')
    'km / s ** 2'
    >>> format(q, '~Pu')
    'km/s²'

-----

I believe this is a viable fix for #160.

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -134,8 +134,11 @@ class _Quantity(object):
         else:
             units = self.units
 
-        return '%s %s' % (format(self.magnitude, remove_custom_flags(spec)),
-                          format(units, spec))
+        if 'u' in spec:
+            return format(units, spec.replace('u', ''))
+        else:
+            return '%s %s' % (format(self.magnitude, remove_custom_flags(spec)),
+                              format(units, spec))
 
     # IPython related code
     def _repr_html_(self):


hgrecco:
Thanks for looking into this. +1 on doing this, but using `U` (capital letter) in accordance with other format specifiers.

Please notice that in the current Pint, Units container is a leaky abstraction which knows nothing about the registry and therefore cannot do certain type of formatting (ie. symbol representaiton of a unit). The good news is that this will change for the next version of Pint. The develop branch already has the new behavior. Please make your patch against develop which is where we include new features.

jpgrayson:
Thanks for taking a look at this patch. I think I jumped the gun with this changeset--I had not yet looked at the development branch. I see that with the new _Unit class the '~' formatting flag applies correctly in, for example, format(q.units, '~'). Insofar as that use case works, I do not believe that a new 'U' flag is required. If we do want to go ahead with the 'U' flag, I think perhaps we might also want a complimentary 'Q' flag. I.e.:

    >>> 'A magnitude of {q:Q} with units of {q:U}'.format(q=q)

Without these new 'Q' and 'U' flags, the above would alternatively be written:

    >>> 'A magnitude of {q.magnitude} with units of {q.units}'.format(q=q)

This latter form is arguably nicer, but can at least be considered good enough such that we can say no to the feature creep of new formatter flags.

So I am going to respectfully withdraw this pull request and anxiously await a pint release from the develop branch.

hgrecco:
Thanks for your feedback. The plan is to stabilize and test the Systems (#215), tokenizer (#226) and compact (#224)

Hopefully we would also be able to reach a conclusion and some code about #197, #196 and #195

If everything goes fine, Pint 0.7 will be out in late March.

PR #235, opened 2015-02-10
Additional unit definitions: 'ha' symbol (hectare) and 'quad' (10**15 Btu)

edschofield:
'Quad' is a unit of energy used when discussing world energy budgets. See:

http://www.aps.org/policy/reports/popa-reports/energy/units.cfm
https://en.wikipedia.org/wiki/Quad_(unit)


hgrecco:
Merged in develop. Thanks!

Issue #238, opened 2015-02-13
Add Travis, Coverage, Pypi badge


scls19fr:
It will be nice to add Travis, Coverage, Pypi badge to frontpage

scls19fr:
For example:

    [![Build Status](https://travis-ci.org/hgrecco/pint.svg?branch=master)](https://travis-ci.org/hgrecco/pint)

Some other example for Pypi https://pypip.in/

jonemo:
Given the discussion in https://github.com/hgrecco/pint/pull/241, this can probably be closed as #wontfix.

bilderbuchi:
This issue can be closed, those badges are there now.

hgrecco:
thanks

PR #241, opened 2015-02-13
Update README

scls19fr:
Add some badges (PyPi, Travis, Coveralls).
Unfortunately I'm doing this "blindly" because file extension of `README` should be `.md` to have preview in GitHub.


scls19fr:
see https://github.com/hgrecco/pint/issues/238

MatthieuDartiailh:
I would be + 1 on adding badges and if you want you can keep a README without badges that you distribute when bundling code.

hgrecco:
A lot has changed in 4 years and the large amount of projects supporting this and the experience using them has convinced me that I was wrong

PR #246, opened 2015-02-27
modified the Bo to B

a358003542:
modified the Bo to B , am I right?


hgrecco:
yes. We just need to make sure that is does not collide with others.

hgrecco:
Merged. Thanks

PR #247, opened 2015-03-06
Custom eval

cheezman34:
I've created a module pint_eval.py that implements a custom evaluator so we can avoid using python's built-in eval.  All the changes are currently contained within pint_eval.py, so that it's easy to conceptualize.  Once we're satisfied with functionality, we can integrate it into the repo as a whole.

For now, take a look and let me know if there are any bugs, missing functionality, or any other issues.  This has only been tested on python 2.7, so let me know if I need to make any changes for other python versions.

### DIFF
--- a/pint/pint_eval.py
+++ b/pint/pint_eval.py
@@ -8,13 +8,13 @@ from decimal import Decimal
 import math
 import operator
 
-from pint.util import ParserHelper, UnitsContainer
-
 import token as tokenlib
 
 #for controlling order of operations
 _OP_PRIORITY = {
-    '**' : 2,
+    '**' : 3,
+    '^' : 3,
+    'unary' : 2,
     '*' : 1,
     '' : 1, #operator for implicit ops
     '/' : 1,
@@ -63,7 +63,7 @@ class EvalTreeNode(object):
             return self.left[1]
         return '(%s)' % ' '.join(comps)
     
-    def evaluate(self, define_op, bin_op, un_op):
+    def evaluate(self, define_op, bin_op=_BINARY_OPERATOR_MAP, un_op=_UNARY_OPERATOR_MAP):
         '''
         define_op is a callable that translates tokens into objects
         bin_op and un_op provide functions for performing binary and unary operations
@@ -87,7 +87,7 @@ class EvalTreeNode(object):
             return define_op(self.left)
         
 
-def build_eval_tree(tokens, op_priority, index=0, depth=0, prev_op=None, ):
+def build_eval_tree(tokens, op_priority=_OP_PRIORITY, index=0, depth=0, prev_op=None, ):
     '''
     Params:
     Index, depth, and prev_op used recursively, so don't touch.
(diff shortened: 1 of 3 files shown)


kdeldycke:
:+1: as the use of Python's `eval` leads to quite bad behaviour. See #267 for instance.

hgrecco:
Everything that I have tested seems to work. Shall we do a few benchmarks to see that the performance in common operations is not reduced?

MatthieuDartiailh:
I would say yes for the benchmarks !

rec:
Very cool pull request!

Delurking for a second, I'd say that, while benchmarks are always great, if
someone's using eval then performance probably isn't one of their concerns.
 :-)

If users want to make repeated calls to eval efficient, then they usually
pre-compile the code being evaluated with a trick like this:
http://stackoverflow.com/questions/12467570/python-way-to-speed-up-a-repeatedly-executed-eval-statement

Your existing build_eval_tree function would be really useful for that.  If
it could be tweaked so external users could "compile, then evaluate" and
not have to redo the code starting around here:
https://github.com/hgrecco/pint/pull/247/files#diff-9950937a1468c4bc5804fee4595ef11dR214,
that might be even better!  You'd probably wrap it in some sort of class
with just two methods...

Back into my cave... :-)

On Sat, Apr 18, 2015 at 3:59 AM, Matthieu Dartiailh <
notifications@github.com> wrote:

> I would say yes for the benchmarks !
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/hgrecco/pint/pull/247#issuecomment-94141703>.
>



-- 
     /t

http://radio.swirly.com - art music radio 24/7 366/1000

cheezman34:
By all means, benchmarks sound like a good idea.  Without digging more into it, I think there's a decent chance this approach is already faster than the original.  

I've been kinda swamped at work, so if anybody wants to run some benchmarks, feel free to do so.  Otherwise I might be able to get to it later this week (or weekend).

cheezman34:
So I ran the existing benchmarking (comparing develop branch to master branch) and they look pretty similar. 


MASTER:
bench_base.yaml
---------------

7.24e-07   importing
2.95e-04   empty registry
1.12e-01   default registry
5.03e-06   finding meter (attr)
1.51e-04   finding meter (item)
5.96e-05   base units meter
6.04e-05   base units yard
3.07e-05   base units meter / second
4.30e-05   base units yard / minute
5.28e-02   build cache

bench_numpy.yaml
----------------

5.94e-07   NumPy cosine (base)
7.50e+01   NumPy cosine radian (normalized)
8.27e+01   NumPy cosine dimensionless (normalized)
1.98e+02   NumPy cosine degree (normalized)


DEVELOP
bench_base.yaml
---------------

6.81e-07   importing
1.93e-04   empty registry
1.30e-01   default registry
7.10e-06   finding meter (attr)
1.39e-04   finding meter (item)
2.12e-05   base units meter
2.14e-05   base units yard
3.59e-05   base units meter / second
5.31e-05   base units yard / minute
3.46e-02   build cache

bench_numpy.yaml
----------------

6.21e-07   NumPy cosine (base)
8.07e+01   NumPy cosine radian (normalized)
9.14e+01   NumPy cosine dimensionless (normalized)
1.41e+02   NumPy cosine degree (normalized)


Everything seems to more or less be within an order of magnitude.  I took the liberty of running cProfile on some basic unit parsing, and it looks like significant amounts of time are spent in pint.unit.UnitRegistry.parse_unit_name and pint.compat._tokenize.  If I were worried about performance, I'd start with caching results from parse_unit_name, or possibly working out a faster algorithm.

cheezman34:
So what do we have left to do before this can be merged?  I've been poking around with profilers and such, and there are some performance tweaks that can be made here and there, but it seems like most of the running time is in other parts of pint instead of in the tree-building and evaluation steps.

hgrecco:
Sorry for the delay. This is a big change (the parser is at the heart of pint) so I wanted to be sure. Therefore I was using this in my daily work for a while and I am very happy with it. So I am pretty confident that we should merge it. I will do it on the weekend. Thanks for your great work!

cheezman34:
Fantastic, thanks!

hgrecco:
Congratulations! Again, great work!

PR #258, opened 2015-03-24
Change troy ounce from 480 grams to 480 grains in default_en.txt

efabens:
A troy ounce is 480 grains not 480 grams
http://en.wikipedia.org/wiki/Ounce

I also realize that my text editor auto cleaned up some whitespace. This is my first pull request so I am not positive if I did this correctly.


hgrecco:
Thanks! It was an ugly typo!

hgrecco:
Merged in develop. Thanks

PR #277, opened 2015-06-17
added wavenumber to the spectroscopy context

antonl:
Wavenumber conversion is missing from the spectroscopy context. This request adds it to ```default_en.txt```.


antonl:
Looks like it works for simple cases, but breaks on reverse transformation because it allows ```1/[length] <-> [length]``` conversions. I'll play with it some more.

Issue #282, opened 2015-07-09
Decorator wrapper to check the dimensions on function input args


kaidokert:
There is a ureg.wraps decorator that converts from builtins to pint quantities. Is anyone interested in decorator that checks correct dimensionality on input variables ? I found i needed this in my own library that uses Pint a lot.
The gist is here, i can make this into a pull request if there is interest

https://gist.github.com/kaidokert/7e23f4593cc7b4b69c92

hgrecco:
Yes! I would like a PR in relation to this. Take a look at #198 #196 and #195

kaidokert:
Couple of small questions. I'm not sure about naming the API calls properly for a straightforward and self-explanatory use, so @ureg.require_units ? @ureg.require_dimensions ? 

Second, in my usage and the gist i posted, i do not enforce or check return value there ( different from @ureg.wraps ). My reasoning is this : if i'm writing a library that manipulates quantities and it expects inputs from unknown code, checking their dimensionality attribute is probably the least offensive type checking, as exact type checking in kind of un-pythonic. If the duck has a dimension of length, you can reasonably try to calculate how many of them fit in the phone booth.

However, returning correct type from the same library call should be ensured by the unit tests on my code, rather than at run-time by the wrapper function. 

The reasoning is different for @ureg.wraps having a return type spec, as it is intended to actually perform conversions on a wrapped function. Does that make sense ?

hgrecco:
Thanks for looking into this. 

Indeed this is different from `wraps`. `wraps` is not a checker, but rather a way to encapsulate non-pint aware code. That is why it takes quantities, pass their (converted) magnitude into the function and then use the return value as the magnitude of a new quantity.  Your line of reasoning  make sense. If we follow this idea, we need to make it clear in the docs as it will have a different API compared to `wraps`. (again, it is not a problem just need to be clear). IF we want to have the same API, we could allow  `None` as a way to skip it (this is what we do in `wraps`)

Regarding the naming, another possibility will be `check`. It is small and clear. I think we should have just one that works for both units and dimensions.

Regarding the implementation, in `wraps` we parse the input variables during decorations. This is better as it avoid an overhead when calling the wrapped function. Take a look at the `wraps` code.

... and make the PR against develop

hgrecco:
By the way, what do you think about #195 and #196 ?

hgrecco:
Merged

PR #283, opened 2015-07-09
Add initial draft of @wrap.check decorator

kaidokert:
Proposed way of doing  #282

### DIFF
--- a/docs/wrapping.rst
+++ b/docs/wrapping.rst
@@ -147,5 +147,24 @@ To avoid the conversion of an argument or return value, use None
     >>> mypp3 = ureg.wraps((ureg.second, None), ureg.meter)(pendulum_period_error)
 
 
+Checking units
+==============
 
+<to be written>
+When you want pint quantities to be used as inputs to your functions, pint provides a wrapper to ensure units are of
+correct type - or more precisely, they match the expected dimensionality of the physical quantity.
+
+Similar to wraps(), you can pass None to skip checking of some parameters, but the return parameter type is not checked.
+
+.. doctest::
+
+    >>> mypp = ureg.check('[length]')(pendulum_period)
+
+In the decorator format:
+
+.. doctest::
+
+    >>>@ureg.wraps('[length]')
+    ... def pendulum_period(length):
+    ...     return 2*math.pi*math.sqrt(length/G)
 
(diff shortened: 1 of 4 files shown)


docs/wrapping.rst
```diff
@@ -147,5 +147,24 @@ To avoid the conversion of an argument or return value, use None
...
+
+.. doctest::
+
+    >>> mypp = ureg.check('[length]')(pendulum_period)
+
+In the decorator format:
+
+.. doctest::
+
+    >>>@ureg.wraps('[length]')
```
hgrecco: this should be `check`

kaidokert:
Whoops i failed python3 really bad here. Will go back and fix

pint/unit.py
```diff
@@ -1145,6 +1145,36 @@ def wrapper(*values, **kw):
...
+        """
+        dimensions = [self.get_dimensionality(dim) for dim in args]
+
+        def decorator(func):
+            assigned = tuple(attr for attr in functools.WRAPPER_ASSIGNMENTS if hasattr(func, attr))
+            updated = tuple(attr for attr in functools.WRAPPER_UPDATES if hasattr(func, attr))
+
+            @functools.wraps(func, assigned=assigned, updated=updated)
+            def wrapper(*values, **kwargs):
+                for dim, value in itertools.izip_longest(dimensions, values):
```
hgrecco: Two things about this.
1) izip_longest does not exist in Python 3+ (it is called zip_longest). Please add it via the compat module
2) In `wraps` we used zip, but actually I think that zip_longest is better. Would you mind submitting another PR fixing that.

pint/unit.py
```diff
@@ -1145,6 +1145,36 @@ def wrapper(*values, **kw):
...
+                    if dim and value.dimensionality != dim:
+                        raise TypeError(
```
hgrecco: I think that a `DimensionlityError` would be a better choice

kaidokert: Yeah i thought about this too, wasnt sure.

kaidokert:
Addressed most of the comments. I'll PR use of zip_longest as well after this. 
DimensionalityError that is thrown looks kinda funny now..
DimensionalityError: Cannot convert from '12 parsec' ([length]) to 'a quantity of' ([length] / [time])
But i think that would be mostly OK and understandable.

hgrecco:
This is merged but somehow GitHub is not recognizing it.

PR #285, opened 2015-07-16
Properly support deepcopy of Quantity objects.

nevsan:
deepcopy of quantities will currently drop the unit information (just returning the magnitude).


hgrecco:
Thanks but this is already implemented in develop.

PR #286, opened 2015-07-16
Make Quantity and UnitContainer objects hashable.

nevsan:


hgrecco:
This has been discussed in the past and the answer has been: "no, a quantity is a mutable object when using a ndarray as magnitude and therefore it cannot be hashable". For UnitsContainers the situation is different. Take a look at develop where there has been considerable rework and see #202

MatthieuDartiailh:
It might be the right time for a new release.

nevsan:
Understood. But when a Quantity is used with an immutable object, shouldn't it be allowed to be hashable? As implemented in the PR, the hash function will throw an exception if the value is an unhashable type (like an ndarray)--an exception very similar to the one that it throws now--but remain hashable if the underlying object is.

The tuple type behaves exactly the same way: if the items in the tuple are hashable, then the tuple itself is hashable. If the items are not (e.g. ndarrays), then the tuple is not hashable. I view Quantity in a similar light--a container capable of hashability, but only if contained objects are themselves hashable.

pint/quantity.py
```diff
@@ -124,6 +124,9 @@ def __str__(self):
     def __repr__(self):
         return "<Quantity({0}, '{1}')>".format(self._magnitude, self._units)
 
+    def __hash__(self):
+        return hash(self.__class__) + hash(self.magnitude) + hash(self.units)
```
eric-wieser: This should probably be `hash((self.__class__, self.magnitude, self.units))`, to allow `tuple.__hash__` to combine the values in a better way that addition

nevsan:
I want to revisit this issue. With the new release, I still don't see hashable Quantities and I really think that the hashability of a Quantity should be determined by the object being wrapped. I understand ndarrays are not hashable, but there's no good reason to make a scalar unhashable as well.

This approach would throw an exception if an ndarray is being wrapped, but work fine if it's a scalar--the same behavior as a tuple.

eric-wieser:
Would be good to have some more tests here for when hashes are equal. Also, does `1*ureg.m == 1000*ureg.mm`? Because if so, you need to convert to base units before hashing

nevsan:
All very good suggestions.

pint/testsuite/test_quantity.py
```diff
@@ -89,6 +89,13 @@ def test_quantity_repr(self):
         self.assertEqual(str(x), '4.2 meter')
         self.assertEqual(repr(x), "<Quantity(4.2, 'meter')>")
 
+    def test_quantity_hash(self):
+        x = self.Q_(4.2, 'meter')
+        x2 = self.Q_(4200, 'millimeter')
+        y = self.Q_(4.2, 'second')
+        self.assertEqual(hash(x), hash(x2))
+        self.assertNotEqual(hash(y), hash(x))
```
eric-wieser: This isn't actually a requirement of `hash`, so I'm not sure it should be tested here

eric-wieser:
I think you might need to special case `ureg.dimensionless` too. If it is the case that

    1 == 1 * ureg.dimensionless

Then python requires that

    hash(1) == hast(1 * ureg.dimensionless)

pint/testsuite/test_quantity.py
nevsan: I know that hashes technically can collide. It's more of a check against incomplete hashing. I can remove.

nevsan:
How do we feel about this now?

pint/testsuite/test_quantity.py
```diff
@@ -89,6 +89,14 @@ def test_quantity_repr(self):
...
         self.assertEqual(repr(x), "<Quantity(4.2, 'meter')>")
 
+    def test_quantity_hash(self):
+        x = self.Q_(4.2, 'meter')
+        x2 = self.Q_(4200, 'millimeter')
+        y = self.Q_(2, 'second')
+        z = self.Q_(0.5, 'hertz')
+        self.assertEqual(hash(x), hash(x2))
+        self.assertEqual(hash(y * z), hash(1.0))
+
```
eric-wieser: Needs a test for `dimensionless` equality from different uregs

nevsan: Is the latest what you had in mind?

nevsan:
How do we feel about this now?

eric-wieser:
I think that you should update the date when you amend a commit that is over 4 months old!

Otherwise, looks good. Would be good if the test captured the "equality => hash equality" requirement, to prevent regressions in future. Something like:

```python
if a == b:
    self.assertEqual(hash(a), hash(b))
elif hash(a) == hash(b):
    warnings.warn("Unequal values have the same hash - perhaps equality has been made more strict")
```

But I'd say this doesn't need that to be merged

vascotenner:
Any updates? I like to use quantities as keys in a dictionary, which works fine with python2 but not yet with python3

hgrecco:
Sorry for the very long delay. I am merging this now. I find the implementation very pragmatic and the resulting code that can be written is convenient.

hgrecco:
bors +r

mmcdermott:
Any update on this? This would be really convenient.

hgrecco:
I think it would be a nice addition to improve the "unhashable" error message.

hgrecco:
Thanks!. Anybody wants to provide a PR for the error message. I think that we need to point out that the magnitude is unhashable.

jgkatz:
This seems to have been applied without resolution of the inability to hash numpy arrays. Doesn't this break most compatibility with numpy?

nevsan:
@jgkatz Users of `numpy` arrays shouldn't expect hashibility, so it does not break compatibility with `numpy` (because none of their existing code should ever try to hash a `Quantity`). This change merely enables hashibility for objects that are hashable (like `int` or `float`).

jgkatz:
OK, understood. It looks like in my application I had incorrectly assumed hashability (quantities as arguments in an LRU cache). Thanks for the clarification!

PR #295, opened 2015-09-30
Fix python2.6 incompatibility

cpascual:
pint fails to inititialize the UnitRegistry with python2.6:

~~~~
~/src/pint> python -c 'from pint import UnitRegistry; UR=UnitRegistry()'
(...)
ValueError: While opening default_en.txt
'yard' is not defined in the unit registry
~~~~

The problem is due to a known python2.6 bug triggered in `ParserHelper.__copy__()`. 
See: http://bugs.python.org/issue2646

This patch fixes at least one trigger of this bug (the unit registry can be initialized now) but I haven't checked for other similar triggers

### DIFF
--- a/pint/util.py
+++ b/pint/util.py
@@ -449,7 +449,10 @@ class ParserHelper(UnitsContainer):
                                  for key, value in ret.items()))
 
     def __copy__(self):
-        return ParserHelper(scale=self.scale, **self)
+        # workaround for python2.6 compatibility (cast kwarg keys to str)
+        # see http://bugs.python.org/issue2646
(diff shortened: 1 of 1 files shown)


cpascual:
Hi, I see that travis found problems with my proposed change in Python 3.2 (the rest seem ok)
I am afraid I do not know how to solve those issues. Does this mean that this pull request is not going to progress further?

hgrecco:
This is not your fault. Coverage dropped Python 3.2 support recently. I will fix this with travis.

hgrecco:
This is fixed

PR #298, opened 2015-10-28
Fix TypeError in formatter method

cmft:
To reproduce the error the pull request #295 have to be applied.

In Python 2.6 format_unit method does not manage properly unicode
string and gives a TypeError: "formatter() keywords must be strings".

Fix the error, doing a cast of the _FORMATS[spec].items() to str.

You can reproduce:

<pre>
In [2]: UR = UnitRegistry()

In [3]: UR.parse_units('m')
Out[3]: ---------------------------------------------------------------------------

...
/pint/formatting.py in format_unit(unit, spec)
    183     spec = _parse_spec(spec)
    184     fmt = _FORMATS[spec].items()
--> 185     result = formatter(unit.items(), **fmt)
    186     if spec == 'L':
    187         result = result.replace('[', '{').replace(']', '}')

TypeError: formatter() argument after ** must be a mapping, not list

</pre>

### DIFF
--- a/pint/formatting.py
+++ b/pint/formatting.py
@@ -182,7 +182,10 @@ def format_unit(unit, spec):
         return 'dimensionless'
 
     spec = _parse_spec(spec)
-    fmt = _FORMATS[spec]
+    fmt = {}
+    # Convert keys and values to str to avoid a bug in Python2.6
+    for k, v in _FORMATS[spec].items():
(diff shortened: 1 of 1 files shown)


hgrecco:
See #358

hgrecco:
This is fixed already! Thanks!

PR #321, opened 2016-01-06
ENH: Add support for wavenumber/kayser in spectroscopy context

gerritholl:
In spectroscopy, it is (regrettably) common to express the wavelength in
units of wavenumber (a.k.a. kayser) or cm^{-1}.  This commit adds support
to the default spectroscopy context, so that users can convert from
frequency and wavelength directly to wavenumber units such as kayser.

### DIFF
--- a/pint/default_en.txt
+++ b/pint/default_en.txt
@@ -267,6 +267,8 @@ stere = meter ** 3
     [length] <-> [frequency]: speed_of_light / n / value
     [frequency] -> [energy]: planck_constant * value
     [energy] -> [frequency]: value / planck_constant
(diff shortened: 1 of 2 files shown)


hgrecco:
You need to fix the test in order for this to get accepted. See Travis, but basically you need to add in TestCompatibleUnits:test_context_sp that kayser is now a 'valid' unit in line 474

gerritholl:
Ok, I'll look into this.

gerritholl:
@hgrecco Is anything more needed to get this accepted?

hgrecco:
Sorry for the long delay. This is now merged. Thanks a lot

Issue #323, opened 2016-01-12
Magnitude types and transformations


emilienkofman:
Hi,

I'm not completely sure if this is a bug or if this is the expected behavior. Please let me know if I don't use pint correctly:
```python
>>> from pint import UnitRegistry
>>> ur = UnitRegistry()
>>> Q_ = ur.Quantity
>>> from fractions import Fraction
>>> Q_(Fraction(2,3),ur.s)
<Quantity(2/3, 'second')>
>>> Q_(Fraction(2,3),ur.s).to(ur.ms)
<Quantity(666.666666667, 'millisecond')>
```
As I used the type Fraction, I expected to get a Fraction, namely `Fraction(2000,3)` in this example. If this is a bug, how could I try to fix it?

hgrecco:
It would be very nice to fix it but I do not think is possible without hitting a severe performance hit. The reason of this "bug" is shown in the following snippet:
```python
>>> from fractions import Fraction
Fraction(3, 2)
>>> Fraction(3,2)*.1
0.15000000000000002
```

emilienkofman:
Ok what do you mean with "severe performance hit"? Do you mean that specializing the conversion in the case of `Fraction` magnitude would cause a performance hit in the case of `Float` magnitude? Generally speaking is performance a big issue in pint? If I cannot fix it in pint, is there a trick that I can use to achieve this?

hgrecco:
You can try to fix it in Pint but I think that the calculation will be efficient enough for every day use (I might be wrong here). Additionally, it will be annoying to implement. Pint calculates the conversion factor and then multiplies it by the magnitude. The conversion factor is calculated as float by multiplying floats (which are obtained from parsing the definitions). So yo will need to change the parser works to output fractions.

There is simpler, fast (both to run and implement), not so accurate alternative which might be good enough for you. In the UnitRegistry code locate the function that does the conversion and change the conversion factor to Fraction using the `Fraction.from_float` when the magnitude is a fraction. The relevant code starts here:

https://github.com/hgrecco/pint/blob/master/pint/unit.py#L1109

If it performs well, I will be happy to merge i.

hgrecco:
Thanks a lot! Merged.

PR #324, opened 2016-01-13
UnitRegistry.convert now returns a Fraction if the user submits a Fra…

emilienkofman:
According to what we figured out in the Issue https://github.com/hgrecco/pint/issues/323, in the case the user wants to convert a quantity with a Fraction magnitude, UnitRegistry.convert will return the result with a Fraction magnitude (and not a Float).

### DIFF
--- a/pint/unit.py
+++ b/pint/unit.py
@@ -1111,7 +1111,7 @@ class UnitRegistry(object):
             factor = Decimal(str(factor))
 
         if isinstance(value, Fraction):
-            factor = Fraction(Decimal(str(factor)))
+            factor = Fraction(str(factor))
 
         if inplace:
             value *= factor
--- a/pint/unit.py
+++ b/pint/unit.py
@@ -18,6 +18,7 @@ import functools
 import operator
 import pkg_resources
 from decimal import Decimal
+from fractions import Fraction
 from contextlib import contextmanager, closing
 from io import open, StringIO
 from collections import defaultdict
@@ -1108,6 +1109,9 @@ class UnitRegistry(object):
         # must first convert to Decimal before we can '*' the values
         if isinstance(value, Decimal):
             factor = Decimal(str(factor))
+
+        if isinstance(value, Fraction):
+            factor = Fraction(Decimal(str(factor)))
 
         if inplace:
             value *= factor
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -317,7 +317,7 @@ class _Quantity(SharedRegistryObject):
         >>> (1e-2*ureg('kg m/s^2')).to_compact('N')
         <Quantity(10.0, 'millinewton')>
         """
-        if self.unitless:
+        if self.unitless or self.magnitude==0:
             return self
 
         SI_prefixes = {}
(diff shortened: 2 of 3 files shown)


pint/unit.py
```diff
@@ -1109,6 +1110,9 @@ def _convert(self, value, src, dst, inplace=False):
         if isinstance(value, Decimal):
             factor = Decimal(str(factor))
 
+        if isinstance(value, Fraction):
+            factor = Fraction(str(factor))
```
hgrecco: I think that (Fraction.from_float)[https://docs.python.org/3/library/fractions.html#fractions.Fraction.from_float]) would be a better choice

emilienkofman: I'm not sure since it fails both with python2.6 and python2.7 (but differently):

```python
Python 2.6.9 (unknown, Jan 13 2016, 11:45:16) 
[GCC 4.9.2] on linux3
Type "help", "copyright", "credits" or "license" for more information.
>>> from fractions import Fraction
>>> Fraction(1.1)
  File "/home/ekofman/.pyenv/versions/2.6.9/lib/python2.6/fractions.py", line 99, in __new__
    numerator = operator.index(numerator)
TypeError: 'float' object cannot be interpreted as an index
```

and:

```python
Python 2.7.9 (default, Mar  1 2015, 12:57:24) 
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from fractions import Fraction
>>> Fraction(1.1)
Fraction(2476979795053773, 2251799813685248)
```
Which may also be the reason why the same code with Decimal also transforms the factor to a string and then to a Decimal ?

hgrecco: There is a function called `from_float` which works just fine in Python 2.6
```python
from Python 2.6.9 (unknown, Jul 14 2015, 19:46:31)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from fractions import Fraction
>>> Fraction.from_float(1.1)
Fraction(2476979795053773, 2251799813685248)
```

hgrecco: Doing this is more accurate but fractions will have larger numbers. Regarding the Decimal, there is no equivalent (to my knowledge) in Python 2.6

emilienkofman: Yes but anyway if for instance the factor is 1.1 I want Fraction(factor) to evaluated as 11/10, not 2476979795053773/2251799813685248 (which happens due to the fact that 1.1 has no exact float representation?)

```python
Python 2.7.9 (default, Mar  1 2015, 12:57:24) 
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from fractions import Fraction
>>> Fraction(1.1)
Fraction(2476979795053773, 2251799813685248)
>>> Fraction('1.1')
Fraction(11, 10)
```

hgrecco: You are right! I will merge your PR. Thanks for your work.

emilienkofman:
Hi, I just found another small bug in _Quantity.to_compact() (and submitted a fix):
The bug was this:
```python
from pint import UnitRegistry
ur = UnitRegistry()
ur("0 s").to_compact()
  File "build/bdist.linux-x86_64/egg/pint/quantity.py", line 349, in to_compact
ValueError: math domain error
```

PR #331, opened 2016-02-09
Fix siunitx formatter

### LINKED ISSUE #330: SIUnitx Formatter can't handle fractional powers
Hello,
I am using Pint 0.6, but I also set up a separate function to duplicate the siunitx formatter for myself, since it doesn't exist in 0.6. In working with it, I found that it is unable to print fractional powers (e.g. 0.5) because the power is expected to be an integer: https://github.com/hgrecco/pint/blob/5c409ea6695170b74f63ba39fbc64162c482730b/pint/formatting.py#L206

A simple fix is to test if the power is an integer:
    
    return r'\tothe{{{:d}}}'.format(int(power)) if power.is_integer() else r'\tothe{{{:.1f}}}'.format(power)

Of course, knowing the precision required for the float is the next problem.

bryanwweber:
Fixes #330

### DIFF
--- a/pint/formatting.py
+++ b/pint/formatting.py
@@ -221,12 +221,12 @@ def siunitx_format_unit(units):
                 prefix = p
                 unit = unit.replace(prefix, '', 1)
 
-            if power < 0:
-                l.append(r'\per')
-            if prefix is not None:
-                l.append(r'\{0}'.format(prefix))
-            l.append(r'\{0}'.format(unit))
-            l.append(r'{0}'.format(_tothe(abs(power))))
+        if power < 0:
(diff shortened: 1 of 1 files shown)


hgrecco:
Tests are failing, there is some recursion. Also `int` does not have a `is_integer` method. Maybe you can change it to `isinstance(power, int) or (isinstance(power, float) and power.is_integer())`

bryanwweber:
OK I think I fixed both of them now... (he says before the checks finish)... sorry about that! If you would like me to squash these commits together, I can do that and force push onto this branch.

PR #333, opened 2016-02-11
check inheritance with real base class

mortbauer:
The wraps function now uses quantity._Quantity to check for inheritance since otherwise one had to pass the exact same UnitRegistery instance. But maybe that was intention, for me it works better like that i think.


hgrecco:
That was actually the intention. In general mixing registries is not a good idea. But can you elaborate why this works better for you?

hgrecco:
I am closing this for the time being. Please re-open if you have new comments on this.

PR #334, opened 2016-02-16
to_compact now returns self for NaN and infinite magnitudes

emilienkofman:
I think it makes sense for infinity. For NaN magnitudes this is discutable.

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -317,7 +317,8 @@ class _Quantity(SharedRegistryObject):
         >>> (1e-2*ureg('kg m/s^2')).to_compact('N')
         <Quantity(10.0, 'millinewton')>
         """
-        if self.unitless or self.magnitude==0:
+        if (self.unitless or self.magnitude==0 or 
+            math.isnan(self.magnitude) or math.isinf(self.magnitude)):
             return self
 
         SI_prefixes = {}


hgrecco:
Sounds good to me but this change will fail for arrays
```python
>>> import numpy as np
>>> x = np.asarray([1,2,3])
>>> import math
>>> math.isnan(x)
TypeError: only length-1 arrays can be converted to Python scalars
```

hgrecco:
By the way, the current version in master will also fail:
```python
>>> bool(x==0)
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

emilienkofman:
Ah ok so I'll try to elaborate something better, and eventually add a unit test also. Thanks for those tips.

hgrecco:
Actually, thank you. Your commit made me realize that the current implementation is buggy.

emilienkofman:
Hi, I was just trying to figure out what it should do for arrays, but then I realized that obviously I cannot apply `to_compact` to each element since then they could be associated to different magnitudes. So I guess when the magnitudes is a numpy array it should also return self? Maybe it could also raise a warning to tell why this could not work?

hgrecco:
I think the warning idea and returning self is a good one. Another option is calculating the median (or something like that), find the most appropriate units and the transforming all values to that.

emilienkofman:
Ok I implemented the warning option, which is better in my opinion because as I understand, the magnitude could even be non numerical (I don't know if it makes much sense though). So, it raises a warning if the magnitude is not numerical (and returns self) which includes numpy arrays, lists, and so on. For some reason I deleted my repository and created a new fork, do you know how I can submit my changes here? The commit is here https://github.com/emilienkofman/pint/commit/afff318e62011f5852672fa06234c808e0d543d6

hgrecco:
This was fixed thanks to #340

hgrecco:
Thanks a lot for your work

PR #340, opened 2016-03-04
Quantity.to_compact now returns self and raises a warning when magnit…

emilienkofman:
…ude is not numerical.


hgrecco:
Merged!. Thanks a lot

PR #344, opened 2016-03-14
support changing shapes of numpy arrays

tecki:
In numpy one can simply change the shape of an array by setting
the shape attribute. This patch allows the same for pint Quantities.


hgrecco:
Looks good. The checks are failing because of a problem with a related package. I wil merge it end of the week.

PR #345, opened 2016-03-14
support the flat property of numpy

tecki:
In numpy one can simply iterate over a flattened array by iterating
over ndarray.flat. With this PR that's also possible with pint.


hgrecco:
Looks good. The checks are failing because of a problem with a related package. I wil merge it end of the week.

PR #357, opened 2016-04-08
Implement Unit.__deepcopy__

noahl:
Also add a test case verifying the implementation.

### DIFF
--- a/pint/testsuite/test_unit.py
+++ b/pint/testsuite/test_unit.py
@@ -2,6 +2,7 @@
 
 from __future__ import division, unicode_literals, print_function, absolute_import
 
+import copy
 import math
 
 from pint.unit import (UnitRegistry, LazyRegistry)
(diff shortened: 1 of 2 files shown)


noahl:
@hgrecco this PR and one other I've submitted both show failing tests. But when I look at the details, tests pass for all tox environments except for Python 3.5 with uncertainties on. I suspect the issue is related to the environment, not the PRs. Is there anything I can do to help get this PR merged?

hgrecco:
You are right, it is not your fault!. More and more packages are dropping Py2.6 support and maintaining the tests is becoming annoying. I am going to start a poll about dropping 2.6 support, give me a few days.

hgrecco:
See and vote in  #358

noahl:
@hgrecco Thanks for starting the poll, but I don't see how that connects to Python 3.5. It may be my PRs that are at fault - it just seems suspicious that they would pass tests in every environment except for Python 3.5. Do you have any idea what might be causing that? I can look into it more too, but you are much more familiar with the different environments than I am, and I'd like to pick your brain first :).

hgrecco:
My bad! The doctest are failing but is not your fault, it is very likely a change in a uncertainties (a dependent library). I will fix the test to pin to an older version.


.

noahl:
@hgrecco is there anything else you'd like to see for this to get merged?

PR #359, opened 2016-04-13
Add the resulting exception of the operation

yuvallanger:
Should I maybe add more of the error message or would that be too much?

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -305,7 +305,7 @@ also define the registry as the application registry::
     >>> q2 = UnitRegistry().meter
     >>> # q1 and q2 belong to different registries!
     >>> id(q1._REGISTRY) == id(q2._REGISTRY) # False
-    False
+    ValueError: Cannot operate with Unit and Unit of different registries.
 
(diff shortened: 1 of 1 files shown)


hgrecco:
But comparing the id of the registries will result in False. What will raise an exception is operating between the two (for example adding q1 and q2)

yuvallanger:
Woops! Sorry about that! A lesson for life:
Never send pull requests while being hurried away by a broom wielding barista with a fast approaching coffee shop closing time. You are going to make mistakes.

A question: I've tried:
```
q1, q2 = pint.UnitRegistry().meter, pint.UnitRegistry().meter

print(1 * q1 == 2 * q2)  # False
print(1 * q1 == 1 * q2)  # True
print(q1 == q1)  # True
print(q1 == q2)  # ValueError: Cannot operate with Unit and Unit of different registries.
```

Only the last one raises an exception.

hgrecco:
In the first case you are comparing units, in the second Quantities. Is a bug, should raise a value error for all cases.

hgrecco:
I have fixed the tutorial, the Unit comparison still needs fixing

PR #364, opened 2016-04-19
patch to append None for extra results

### LINKED ISSUE #363: don't truncate returns from wrapped function if less units than retval
Pint: 0.7.2

**Proposed**: if `len(results)` > `len(ret)` then append `None` to `ret` in [`registry-helpers.py`](https://github.com/hgrecco/pint/blob/master/pint/registry_helpers.py#L170)

```patch
--- pint-registry_helpers.py.orig       2016-04-19 10:08:31.506029000 -0700
+++ pint-registry_helpers.py    2016-04-19 10:28:27.714029000 -0700
@@ -167,10 +167,12 @@
             new_values, values_by_name = converter(ureg, values, strict)

             result = func(*new_values, **kw)
+
+            extra_results = (None, ) * (len(result) - len(ret))

             if container:
                 out_units = (_replace_units(r, values_by_name) if is_ref else r
-                             for (r, is_ref) in ret)
+                             for (r, is_ref) in ret + extra_results)
                 return ret.__class__(res if unit is None else ureg.Quantity(res, unit)
                                      for unit, res in zip(out_units, result))
```

**Expected**: if number of return units specified in wrapper is less than number of actual returns from function, then `None` should appended to `out_units` or missing returns should be appended before wrapper returns.

**Observed**: the returns are truncated because wrapper uses `zip()` and `out_units` is less than results.

This example is from [`uncertianty_wrapper`](../../../SunPower/UncertaintyWrapper/blob/master/uncertainty_wrapper/tests/test_uncertainty_wrapper.py#L177) which appends the covariance and jacobian matrices to the return values.

```python
@UREG.wraps(('deg', 'deg', 'dimensionless', 'dimensionless', None, None),
            ('deg', 'deg', 'millibar', 'degC', None, 'second', None), strict=False)
@unc_wrapper_args(0, 1, 2, 3, 5)
def solar_position(lat, lon, press, tamb, timestamps, seconds=0):
    pass
```

I must add `None, None` to my list of return units otherwise the covariance and jacobian matrices are truncated.

**Related**: #196

mikofski:
fixes #363 

* append tuple of `None` for extra results so they're not truncated by `zip()`


rec:
I think this is a job for [itertools.izip_longest](https://docs.python.org/2/library/itertools.html#itertools.izip_longest).

mikofski:
Thanks @rec I made the change to `izip_longest` and added a test.

@hgrecco this only happens if the return units are given as a sequence, that's why the previous test on [line 330 of `test_unit.py`](../../pint/blob/master/pint/testsuite/test_unit.py#L330) passed.

EG:

```python
import pint

ureg = pint.UnitRegistry()

def f(x,y): return x, y

# return units not a sequence
g = ureg.wraps(None, (None, None))(f)
g(1, 2)
# (1, 2)
# PASSES

# return units a list
g = ureg.wraps([None], (None, None))(f)
g(1, 2)
# [1]
# FAILS

# return units a tuple
g = ureg.wraps((None,), (None, None))(f)
g(1, 2)
# (1,)
# FAILS
```

mikofski:
PS @hgrecco I think your TravisCI build test is failing because of `float128` [on line 502 in `test_issues.py`](https://github.com/hgrecco/pint/blob/master/pint/testsuite/test_issues.py#L502)

```python
  File "C:\Users\mmikofski\Documents\Projects\pint\pint\testsuite\test_issues.py", line 502, in test_issue250
    self.assertEqual(np.float128(a/b), 1000.)
AttributeError: 'module' object has no attribute 'float128'
```

hgrecco:
Looks good to me. Please update the docs and I will merge it.

hgrecco:
Sorry for the really long delay. I will merge this week and hopefully release a new version soon. Could you please rebase to the develop branch? I am combining and testing together many patches before merging them to the master branch.

(Also That branch is not failing )

hgrecco:
Merged in develop. Thanks a lot!

Issue #366, opened 2016-04-22
nautical_miles, nanomiles, and inline comments


whoburg:
Inline comment in `default_en.txt` appears to break nautical_miles formatting and "nmi" lookup:

```python
In [1]: from pint import UnitRegistry

In [2]: ureg = UnitRegistry()

In [3]: "{:~}".format(ureg("nautical_mile"))
Out[3]: '1 nmi # exact'

In [4]: ureg("nmi")
Out[4]: <Quantity(1, 'nanomile')>
```

hgrecco:
Maybe we can just support it by patching the SourceIterator https://github.com/hgrecco/pint/blob/b2287c5e0f345bcc691d158324b2abeddbb51962/pint/util.py#L694

whoburg:
I just reinstalled pint with `pip install pint==0.7.2`, and this issue still exists. Should it really be closed?

PR #367, opened 2016-04-22
remove inline comment, patch for #366

whoburg:
This is a simple patch for #366. Not sure if inline comments in unit definition files are intended to work -- if they are, a better fix should be implemented and tested.


hgrecco:
Inline comments are now supported
 Styling with Markdown is supported
Comment Close and comment

PR #376, opened 2016-05-09
Fix ValueError in parse_units method

cmft:
parse_units raises an ValueError when the input is a string with only white spaces. 
e.g. 
' ', '\t', etc.

Fix the error striping the input_string before working with it.

You can reproduce the error with this code:
import pint
ur = pint.UnitRegistry()
ur.parse_units(' ')

PS: In my solution, I considered  that the behaviour of parse_units  for these cases should be return dimensionless unit, but  another solution can be implemented, e.g. raising an UndefinedUnitError exception in these cases. What do you thing?

### DIFF
--- a/pint/unit.py
+++ b/pint/unit.py
@@ -1216,11 +1216,11 @@ class UnitRegistry(object):
         if as_delta and input_string in self._parse_unit_cache:
             return self._parse_unit_cache[input_string]
 
-        # Sanitize input_string with whitespaces.
-        input_string = input_string.strip()
-
         if not input_string:
             return UnitsContainer()
+
+        # Sanitize input_string with whitespaces.
+        input_string = input_string.strip()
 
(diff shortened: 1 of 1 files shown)


cpascual:
Any news on this?
We are waiting for this to enter pint, but if it won't be merged we may do a workaround in our own code.

hgrecco:
I apologize for the delay and thanks a lot for your bug fix. Could you please rebase to the develop branch? I am combining and testing together many patches before merging them to the master branch.

hgrecco:
I will merge this week and hopefully release a new version soon.

cmft:
No problem and many thanks. I could not do a clean rebase without affecting the history, so for two lines I saw more easy to create a new PR for develop  https://github.com/hgrecco/pint/pull/413

If you are agree please close it and use the other.

Best,
Carlos

PR #377, opened 2016-05-10
Use roman type for LaTeX units and add a space before them

endolith:
Italics should only be used for variables, units should be written in roman text.

See http://latex.wikia.com/wiki/Mathrm_%28LaTeX_symbol%29 or http://physics.nist.gov/cuu/pdf/typefaces.pdf for example.

josephwright/siunitx@51bd580 uses `\ ` for the space, if I'm reading it correctly.

**Before:**

> 2016-05-09 21_56_14-spyder python 2 7

**After:**

> 2016-05-09 21_57_02-spyder python 2 7

**Still a problem:**

> 2016-05-09 21_54_41-spyder python 2 7

which could be fixed by replacing `_` with `\_` in the units?


endolith:
Also could rewrite `1e-13\ \mathrm{meter}` as `1 \times 10^{-13}\ \mathrm{meter}`?

hgrecco:
Thanks a lot. This is really nice. Indeed, displaying constants is still something in the todo list (the constants themselves are)!

Could you please rebase to the develop branch? I am combining and testing together many patches before merging them to the master branch.

endolith:
ok squashed and rebased.  I don't understand what the errors about DPI are though

PR #393, opened 2016-06-26
Ascii formatting

dplepage:
Possibly I'm the only one bothered by this, but this formatting seems weird to me:

```python
>>> str(3/ureg.s)
'3 1 / second'
```
This PR checks for the specific case where you would write `<mag> 1 / <units>` and returns `<mag> / <units>` instead, e.g.
```python
>>> str(3/ureg.s)
'3 / second'
```

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -122,7 +122,7 @@ class _Quantity(SharedRegistryObject):
     def __format__(self, spec):
         spec = spec or self.default_format
 
-        allf = '{0} {1}'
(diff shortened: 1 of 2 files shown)


hgrecco:
Thanks a lot. Sounds good to me! Could you please rebase to the develop branch (where I have a rather large change)? I am combining and testing together many patches before merging them to the master branch.

hgrecco:
This is merged, somehow github is not detecting! Thanks a log!

Issue #396, opened 2016-07-06
Add support for median, average, and others


gerritholl:
I've been trying to add support for more ufuncs that keep units, by adding items to `__copy_units`.  Unfortunately, many ufuncs start with calling `asarray` or `asanyarray`, and we probably don't want to mess with those: people calling `asarray` will not expect to get a `Quantity` in return.  Is there another way this could be achieved?

Relevant ufuncs include: `median`, `average`, `percentile`, `expand_dims`, `atleast_1d`, `atleast_2d`, `atleast_3d`, `tile`, `rot90`, and others.

hgrecco:
I tried this in the past. The problem is that numpy calls asanyarray which does not accept anything that is not a subclass of ndarray.

hgrecco:
Revisit after #905

PR #397, opened 2016-07-06
ENH/TST: Make ndarray-quantity swapaxes-proof and add several tests

gerritholl:
Make a numpy array quantity retain its unit when swapaxes is called.
Previously it would turn into a regular ndarray.  Add a test suite to test
unit retainment for this and other ufuncs in __copy_units.

I also tried to add functionality for others, such as median, average,
percentile, expand_dims, atleast_1d, atleast_2d, atleast_3d, tile, rot90.
However, this doesn't work because those ndarrays start with `asarray` or
`asndarray`.  Onother solution is needed.  See also issue #396.

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -984,7 +984,7 @@ class _Quantity(SharedRegistryObject):
     #: original.
     __copy_units = 'compress conj conjugate copy cumsum diagonal flatten ' \
                    'max mean min ptp ravel repeat reshape round ' \
-                   'squeeze std sum take trace transpose ' \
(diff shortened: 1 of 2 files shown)


gerritholl:
Why does it say coverage has decreased, and how do I fix this?  I added some tests, so coverage should have increased?

hgrecco:
Thanks a lot. Sounds good to me! Could you please rebase to the develop branch (where I have a rather large change)? I am combining and testing together many patches before merging them to the master branch.

hgrecco:
And do not worry about the coverage, it is likely not important. I will look into it.

PR #406, opened 2016-07-25
fix bug in infer_base_unit. 

pmaunz:
infer_base_unit failed on quantities that were generated by multiplying the same dimension with different prefixes such as 'meter * millimeter'. This also leads to problems using .to_compact() as the base_unit of a quantity is inferred incorrectly
An example can be found in the added testcase.

### DIFF
--- a/pint/testsuite/test_numpy.py
+++ b/pint/testsuite/test_numpy.py
@@ -5,7 +5,7 @@ from __future__ import division, unicode_literals, print_function, absolute_impo
 import copy
 import operator as op
 
-from pint import DimensionalityError
+from pint import DimensionalityError, set_application_registry
 from pint.compat import np, unittest
 from pint.testsuite import QuantityTestCase, helpers
 from pint.testsuite.test_umath import TestUFuncs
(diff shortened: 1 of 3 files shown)


pint/testsuite/test_numpy.py
```diff
@@ -240,7 +240,7 @@ def test_reversible_op(self):
 
     def test_pickle(self):
         import pickle
-
+        set_application_registry(self.ureg)
```
hgrecco: I think this change belongs to a different PR

pint/testsuite/test_numpy.py
```diff
@@ -5,7 +5,7 @@
 import copy
 import operator as op
 
-from pint import DimensionalityError
+from pint import DimensionalityError, set_application_registry
```
hgrecco: I think this change belongs to a different PR

hgrecco:
Thanks a lot. Sounds good to me! Could you please rebase to the develop branch? I am combining and testing together many patches before merging them to the master branch.

pmaunz:
Thanks for the feedback. I closed this pull request and submitted two pull requests instead. 
`fix-pickle-unittest`  and
`fix-infer_base_unit` (where I have changed the code to use your udict instead of the defaultdict)

hgrecco:
That was fast!

PR #411, opened 2016-08-01
Fix pickle unittest

pmaunz:
In pickle unittest set_application_registry has to be called to prevent the test from occasionally failing.


hgrecco:
Merged in develop!. Thanks

PR #412, opened 2016-08-01
Fix infer base unit

pmaunz:
infer_base_unit failed on quantities that were generated by multiplying the same dimension with different prefixes such as 'meter * millimeter'. This also leads to problems using .to_compact() as the base_unit of a quantity is inferred incorrectly
An example can be found in the added testcase.


hgrecco:
Merged in develop!. Thanks

PR #413, opened 2016-08-01
Fix ValueError in parse_units method

cmft:
Unit register parse_units method raises an ValueError
when the input is a string with only white spaces.

e.g. ' ', '\t', etc.

Fix the error striping the input_string before working with it.


hgrecco:
Thanks! Merged in Develop!

PR #415, opened 2016-08-01
add return to example in docs

### LINKED ISSUE #363: don't truncate returns from wrapped function if less units than retval

mikofski:
* fixes #363
* amends #364
* improves wraps documentation when there are more returns than units in the wrapper, actually shows 3 returns in example, but only 2 units in wrapper


mikofski:
this is a minor change to the docs only:

```diff
diff --git a/docs/wrapping.rst b/docs/wrapping.rst
index f7ac662..19f4d3c 100644
--- a/docs/wrapping.rst
+++ b/docs/wrapping.rst
@@ -146,7 +146,7 @@ airmass::

     @UREG.wraps(('deg', 'deg'), ('deg', 'deg', 'millibar', 'degC')
     def solar_position(lat, lon, press, tamb, timestamp):
-        pass
+        return zenith, azimuth, airmass


 Specifying relations between arguments
```

mikofski:
I will open a new PR on master, sorry.

PR #420, opened 2016-08-09
Prevent crash when context is None

eric-wieser:
Triggered by `linalg`, among other things

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -1223,7 +1223,7 @@ class _Quantity(SharedRegistryObject):
     __array_priority__ = 17
 
     def __array_prepare__(self, obj, context=None):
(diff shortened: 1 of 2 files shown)


eric-wieser:
There also seems to be a problem with `__array_wrap__`  and `np.maximum`

eric-wieser:
Coverage has "decreased" only for numpy-less builds - not really a surprise or a failure

hgrecco:
:+1:

eric-wieser:
Ok, I've combined this with #421

eric-wieser:
@hgrecco, what's the upgrade path from a 👍 to a merge? ;)

hgrecco:
I think this is no longer needed after #905. I am closing, please reopen if needed.

PR #421, opened 2016-08-10
np.maximum causes an error in __array_wrap__

eric-wieser:


eric-wieser:
`np.isnan` causes this problem too

eric-wieser:
Dupe of #420

PR #435, opened 2016-09-16
fix to_compact for negative magnitude 

pmaunz:
Q(-1000, 's').to_compact() did fail because it tried to compute the logarithm of the negative magnitude.


hgrecco:
Fixed by @65a7aa2

PR #443, opened 2016-10-06
Fixed issue with negative values in to_compact() method

nowox:
An exception `MathError` was raised when trying to get the compact version of `-1.0 meter` because the log10 of a negative magnitude `-1.0`  cannot be computed.


hgrecco:
👍

hgrecco:
Thanks

PR #457, opened 2016-11-17
Correctly order units in siunitx formatting

### LINKED ISSUE #441: siunitx formatter does not preserve order of units
The `{:Lx}` siunitx formatter does not preserve the order of units.  In the example below, the order of units has completely changed when using the siunitx formatter:

```
In [57]: q = ureg.Quantity(1, "count * m ** 2 * sr / cm / mW")

In [58]: print("{:~}".format(q))
1 count * m ** 2 * sr / cm / mW

In [59]: print("{:Lx}".format(q))
\SI[]{1}{\per\centi\meter\count\meter\squared\per\milli\watt\steradian}
```

hgrecco:
Close #441

### DIFF
--- a/pint/formatting.py
+++ b/pint/formatting.py
@@ -243,11 +243,14 @@ def siunitx_format_unit(units):
             # limit float powers to 3 decimal places
             return r'\tothe{{{:.3f}}}'.format(power).rstrip('0')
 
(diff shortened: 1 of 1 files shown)


hgrecco:
👍

PR #458, opened 2016-11-17
Rename `huh` to `out_i`, meaning "output index"

eric-wieser:
Presumably `huh` previously meant "I'm not quite sure what this is", and when you worked it out, you never renamed it?


hgrecco:
👍

PR #464, opened 2016-11-22
support dimensionless values in wraps

tecki:
when wrapping a function with UnitRegistry.wraps, we enforce that
every argument that has a variable unit actually has a unit.
This means we cannot use the function in the way it was used before.

As an example:

    @ureg.wraps("=A", ("=A", "=A"))
    def f(x, y):
        return x + y

    f(3 * ureg.m, 4 * ureg.m)  # works fine
    f(3, 4)  # doesn't work, but should
    f(3, 4 * ureg.m)  # should (and does) raise an exception

This PR simply treats every non-pint-quantity as dimensionless.

### DIFF
--- a/pint/registry_helpers.py
+++ b/pint/registry_helpers.py
@@ -13,7 +13,7 @@ import functools
 
 from .compat import string_types, zip_longest
 from .errors import DimensionalityError
-from .util import to_units_container
+from .util import to_units_container, UnitsContainer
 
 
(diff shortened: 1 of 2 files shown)


hgrecco:
👍

hgrecco:
Thanks a lot!

tecki:
That was quick, thanks!

Issue #350, opened 2016-03-21
The Measurements Overhaul for 0.8


hgrecco:
This issue is to summarize the overhaul of Measurements targeted for Pint 0.8. The features/open questions are:

- Support all basic math operations. See #278 
- Support for ndarray See #128 
- Support parsing measurements (No issue created yet)
- Support for numpy functions
- How to deal with mean and std methods See #348 
- A math module that supports Quantities and Measurements See #24

alexfeld:
I figured out a better fix for this the first problem listed if you are interested. The problem comes in the call to the `__new__` constructor. I added the following lines to the `_Quantity.__new__` method before anything else is done.

```python
if isinstance(value, ufloat) and cls is not cls._REGISTRY.Measurement:
            cls = cls._REGISTRY.Measurement
```
The added if statement allows the constructor to use the proper `cls` type, which was the issue before. Otherwise, only the class of the object whose operator was called is passed to the constructor of the result, regardless of the type of `value`.

There was a constructor bug in the `_Measurement` class that I fixed as well

```python
def __new__(cls, value, error, units=MISSING):
        if units is MISSING:
            # new if statement
            # missed option before
            if isinstance(value, ufloat):
                inst = super(_Measurement, cls).__new__(cls, value, error)
                return inst
           
           # the old method 
           try:
                value, units = value.magnitude, value.units
            except AttributeError:
                try:
                    value, error, units = value.nominal_value, value.std_dev, error
                except AttributeError:
                    units = ''
        try:
            error = error.to(units).magnitude
        except AttributeError:
            pass

        inst = super(_Measurement, cls).__new__(cls, ufloat(value, error), units)

        if error < 0:
            raise ValueError('The magnitude of the error cannot be negative'.format(value, error))
        return inst
```

My apologies for not making this suggestion via a pull request. I ended up using my own uncertain math package instead of `uncertainties` which required a few other changes to the `pint` package that aren't necessarily compatible with `uncertainties.ufloat`.

hgrecco:
Thanks for the feedback. Can you elaborate about why are you dropping uncertainties in favor of another option? It might be useful to feedback into the Measurement design.

alexfeld:
Mostly I need a package that can handle non-linear functions. For example, here are the results for a nonlinear function of two random variables that are distributed according to a multi-variate normal. 

```python
x_mean = [1.0, 2.0]
x_covariance = [[1.0, 0.1], [0.1, 1.0]]

def f(x):
[...]

PR #469, opened 2016-12-03
add stochastic math implementation of random variable math

alexfeld:
Different method for implementing uncertain quantities

Also adds fix for #350 regarding functions of Quantities and Measurements not returning the correct objects.

### DIFF
--- a/pint/compat/__init__.py
+++ b/pint/compat/__init__.py
@@ -112,7 +112,8 @@ except ImportError:
         return value
 
 try:
-    from uncertainties import ufloat
+    # from uncertainties import ufloat
+    from .stochastic_math import RandomVariable as ufloat
     HAS_UNCERTAINTIES = True
 except ImportError:
     ufloat = None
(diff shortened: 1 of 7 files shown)


hgrecco:
I am finally able to put some time onto this again. The PR looks interesting indeed, but I think it deserves major discussion as it removes uncertainties and replace it for something completely different. In addition, I do think that we should create a sibling package for measurements.

alexfeld:
Agreed. This is more of a suggestion to start the discussion.

hgrecco:
My wish list:
1. a sibling package with everything that is Measurement related (let's call it PM for now)
2. PM should depend on Pint, but Pint should not depend on PM
3. Pint should be aware of PM and use if installed
4. PM should contain a list of definitions 
5. Pint should be able to integrate such definitions into the registry
6. PM should be able to do basic uncertainty propagation

More:
7. PM should be able to do propagate uncertainties with numpy functions without wrappers.
8. PM should be able to deal with covariance (like uncertainties)
9. PM should be able to propagate general functions (e.g. using the solution proposed here)

hgrecco:
Please start this as a sibling package.

PR #474, opened 2016-12-21
Actually support floor division and modulo

tecki:
Floor division is currently broken in pint. It is implemented like the normal division, just calling the floor division operator on the magnitude.

This is wrong. Floor division with units serves an important use case: imagine the question "I have a 1 m long log, how may pieces of 1 in can I make from that?". This question has a clear answer (it's 39), but pint gives the weird answer "1 meter / inch", which is just wrong.

Python gives a clear definition of what is expected for the divmod function, which calculates floor division and the modulo: "For floating point numbers the result is (q, a % b), where q is usually math.floor(a / b) but may be 1 less than that. In any case q * b + a % b is very close to a, if a % b is non-zero it has the same sign as b, and 0 <= abs(a % b) < abs(b)".

This statement also works well in the united case. The implications are: a // b must be unitless, because the floor of a united value does not make sense. a % b must have the same unit (or at least dimensionality) as a (and b), as otherwise the above sum looses sense.

This pull request implements and tests all of the above.

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -861,6 +861,24 @@ class _Quantity(SharedRegistryObject):
         else:
             raise DimensionalityError(self._units, 'dimensionless')
 
+    def __divmod__(self, other):
+        if not self._check(other):
+            other = self.__class__(other, UnitsContainer({}))
+        q, r = divmod(self._magnitude, other.to(self._units)._magnitude)
+        return (self.__class__(q, UnitsContainer({})),
+                self.__class__(r, self._units))
+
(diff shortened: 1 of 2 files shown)


hgrecco:
Finally I have some time to put into this and hopefully Pint will continue to flow now. I really like the reasoning you are giving even though it departs from Pint "historical" semantics. But that's why we have not reached 1.0 yet, to be able to apply some breaking changes if needs. So I am approving this change, it would be great if you can add a line to the release notes and some docs describing this.

PR #480, opened 2017-02-27
Change `UndefinedUnitError` to inherit from `AttributeError`

jhidding:
Running doctests that use Pint fail because the unit registry raises the wrong error if an attribute is not found, in this case:

```
>>> from pint import UnitRegistry
>>> ur = UnitRegistry()
>>> hasattr(ur, '__wrapped__')
False
```

is the desired behaviour. In current master branch this little test will fail with a `UndefinedUnitError` exception. Changing the `UndefinedUnitError` to inherit from `AttributeError` fixes the issue.

### DIFF
--- a/pint/errors.py
+++ b/pint/errors.py
@@ -19,7 +19,7 @@ class DefinitionSyntaxError(ValueError):
     """
 
     def __init__(self, msg, filename=None, lineno=None):
-        super(ValueError, self).__init__()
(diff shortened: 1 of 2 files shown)


hgrecco:
This is a Breaking Change to be noted in 0.8 release notes

hgrecco:
Thanks for your PR!

Issue #489, opened 2017-03-17
wraps for functions with parameters in arbitrary quantities 


rpuntaie:
Hello,

I've found pint quite nice and wanted to adopt it for our project, but then I failed to wrap my functions.
The following two test cases, added to test_unit.TestRegistry.test_wraps, fail:

```
    f4 = ureg.wraps(10*ureg.uA, [None, ], strict=False)(func)
    self.assertEqual(f4(3.), 30 * ureg.uA)

    f5 = ureg.wraps(10*ureg.uA, [5*ureg.uA, ], strict=False)(func)
    self.assertEqual(f5(10*ureg.uA), 20 * ureg.uA)
```

How can one achieve such wrapping?

If this is not possible I probably can fix it and provide a pull request.

Best, Roland

hgrecco:
Hi! Wraps works by transforming quantities into the required units, feeding the magnitude to the function (which should return a number/array/etc but not a quantity), and then adding the unit to the output. So it is fine that all this cases fail. What are you trying to achieve?

rpuntaie:
Well, the functions I want to wrap return a number whose unit is e.g. 10uA or they expect a number parameter whose unit is 5uA.

hgrecco:
If the units of both are uA, you can just do:
```python
>>> import pint
>>> ureg = pint.UnitRegistry()
>>> def f(x): return x
>>> g = ureg.wraps(ureg.uA, ureg.uA)(f)
>>> g(3 * ureg.uA)
<Quantity(3.0, 'microampere')>
```

rpuntaie:
But the function's unit is e.g. 10uA. A number 1 would mean 10uA. A number 2 would mean 20uA. If I use pint and forward 20uA to the wrapper, the actual value given to the function should be 2. For a return value of 2, the wrapper should give me 20uA. 

I've made a pull request which allows to do that.

Sometimes functions (in C) have such strange units, because it allows to adjust the range of wanted physical current to fit into for example a uint16_t.

hgrecco:
This would collide with other things in Pint. `10uA` is not a valid unit is a quantity with magnitude 10 and unit uA. You could define a unit easily: `ureg.define('Ten_uA = 10 uA')`

rpuntaie:
It may collide with the basic idea behind pint, but I don't see any problem as an extension to `wraps`. If you do, please elaborate.

rpuntaie:
I've tried what you proposed:

        >>> from pint import UnitRegistry
[...]

PR #490, opened 2017-03-17
allow wrapping functions taking and returning units in arbitrary quan…

rpuntaie:
This solves issue #489

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ build/
 dist/
 MANIFEST
 .tox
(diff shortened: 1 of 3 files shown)


hgrecco:
Sorry but I think it is confusing for end users. I think that keeping the parameters just as units (or unit like strings) is simpler to explain, results in a better API, and your requirement can easily be satisfied in other ways.

hgrecco:
Closed #962

Issue #500, opened 2017-04-12
Dimensional reduction


bloer:
I was really excited to see how pint handled a lot of "gotcha" issues, until I ran into something that seemed super simple that pint throws a curve on.

Maybe there's something obvious I'm missing, but it seems like pint should be able to recognize when you have a ratio of two quantities with the same dimensionality and automatically do the conversions to reduce them.  For example: 
    In [1]: import pint
    In [2]: ureg = pint.UnitRegistry()
    In [3]: density = 3*ureg.g / ureg.L
    In [4]: volume = 23*ureg.milliliter
    In [5]: mass = density * volume
    In [6]: print(mass)
    69.0 gram * milliliter / liter
    In [7]: mass.dimensionality
    Out[7]: <UnitsContainer({'[mass]': 1.0})>

So pint even correctly recognizes the dimensionality, and is just carrying around this constant "milliliter/liter" factor.  Users shouldn't have to take an extra step to do something as simple as multiply density*volume to get mass or speed*time to get distance

dalito:
The dimensionality reduction is only done if requested. You can use `mass.to_base_units()` ro `mass.to_root_units()` (see [documentation](https://pint.readthedocs.io/en/0.7.2/systems.html)).

This avoids (i) unnecessary computation and (ii) any decision about type conversion. When following your example in python 2, the dimensionality reduction requires a change from type int to another type (float or decimal?).

bloer:
About the type conversion question, this decisions is already made when you call `to_base_units()`, so I think the same logic should apply there.  

Consider another very unintuitive consequence: 
    x = (10*ureg.feet / 3*ureg.inches)
Now I have a quantity that is both `unitless` and `dimensionless`, but I still have to call `to_base_units()` before I can do anything sensible with it.  Naively I would expect `to_base_units()` to be a no-op on a Quantity that is `unitless`. 

Since there is some potentially unnecessary computation done if dimensions are always reduced by default, what about adding a switch somewhere that users could toggle?  Something like a `Quantity.AUTO_REDUCE_DIMENSIONS` class member?

dalito:
> what about adding a switch somewhere that users could toggle? Something like a Quantity.AUTO_REDUCE_DIMENSIONS class member?

Supporting automatic dimensional reduction via an option sound like a nice idea to me (especially for interactive use). 

[...]

PR #515, opened 2017-05-23
add configurable automatic dimensionality reduction

aglassdarkly:
See #500 

BaseRegistry.auto_reduce_dimensions was added to configure whether to automatically reduce dimensionality. The reduction was implemented using decorators, one for inplace and one for not. The only operations wrapped were multiplication and division. Is this correct?

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -44,6 +44,25 @@ class _Exception(Exception):            # pragma: no cover
         self.internal = internal
 
 
+def reduce_dimensions(f):
+    def wrapped(self, *args, **kwargs):
+        result = f(self, *args, **kwargs)
+        if result._REGISTRY.auto_reduce_dimensions:
+            return result.to_root_units()
+        else:
+            return result
+    return wrapped
+
+
+def ireduce_dimensions(f):
+    def wrapped(self, *args, **kwargs):
+        result = f(self, *args, **kwargs)
+        if result._REGISTRY.auto_reduce_dimensions:
+            result.ito_root_units()
+        return result
+    return wrapped
+
+
 @fix_str_conversions
 class _Quantity(SharedRegistryObject):
     """Implements a class to describe a physical quantity:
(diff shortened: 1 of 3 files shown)


bloer:
This seems to have some undesired side effects, in particular Quantities seem to be converted to base units at construction:

    In [7]: ureg = pint.UnitRegistry(auto_reduce_dimensions=False)

    In [8]: 10*ureg.km
    Out[8]: <Quantity(10, 'kilometer')>

    In [9]: 10*ureg('km')
    Out[9]: <Quantity(10, 'kilometer')>

    In [10]: ureg('10 km')
    Out[10]: <Quantity(10, 'kilometer')>

---

    In [12]: ureg2 = pint.UnitRegistry(auto_reduce_dimensions=True)

    In [13]: 10*ureg2.km
    Out[13]: <Quantity(10000.0, 'meter')>

    In [14]: 10*ureg2('km')
    Out[14]: <Quantity(10000.0, 'meter')>

    In [15]: ureg2('10 km')
    Out[15]: <Quantity(10000.0, 'meter')>

CalebBell:
Hi bloer,

Changing this behavior does not seem to me to be possible to avoid with the current approach using decorators and having it pass on *args. However, doesn't this behavior seems sensible since we've just told pint to automatically reduce units?

I was surprised this was implemented with `to_root_units` not `to_base_units`, though.

bloer:
Hi CalebBell,

In my definition (as a working physicist), there's a difference between "converting" dimensions and "reducing".  It's perfectly sensible to write down speed in 'km/h' and not convert to m/s, but still expect '100 km/h * 30 minutes' to yield 50 km (or even 50,000 m, but not '300 minutes * km / hour').  Of course that's really hard to do, which is why I haven't put in my own PR yet =). 

If the parameter were called 'auto_convert_to_root' this behavior would be expected.

CalebBell:
Hi Bloer,
I realize now/agree about the difference between converting and reducing. It would seem to be a great deal of work to implement this as described, but clarifying the behavior to describe it as converting to base units instead of reducing dimensions wouldn't be hard at all. Looking into this further, the test cases added don't describe what should happen when a single quantity is created. I have no personal preference, but either way there should be a test.

I would wager that your desired behavior for "reducing" would have more applications than converting everything to a base units. From a chemical engineering perspective, the following is undesirable:

```
>>> from pint import *
>>> u = UnitRegistry(auto_reduce_dimensions=True)
>>> print(5*u.Pa*u.s)
5000.0 gram / meter / second

```

Issue #511, opened 2017-05-17
slug missing after version 0.6


whoburg:
I know it's an odd unit, but it looks like `slug = 14.59390 * kilogram` existed in `default_en_0.6.txt` and then went away. Any reason not to include it in pint?

whoburg:
@arthur-brown

awcox21:
I ran into this same thing. Also, even more obscure but still used are slinches. I've been adding them myself with <slug = lbf * s ** 2 / foot> and <slinch = lbf * s ** 2 / inch = blob = slugette>. It would be nice if they were in the default set though.

JerichoKain:
Looks like slug was moved to the `@group` US using Commonwealth1959` under gravitational: https://github.com/hgrecco/pint/commit/6acd64d4e02f3f8ff2bd8cb4d727bc63c02ed81d

and later removed to match the NIST: https://github.com/hgrecco/pint/commit/bf0d2507c066e386f51723eb4dec8987bb3b425a

hgrecco:
This is correct. Just as a comment, units in the groups are still usable just not chosen by default. You can just define it at startup in you script. 
(By the way, yhere was also an idea to support a user configurable unit definition list that is stored in the home directory and loaded automatically upon start. I do not know if there is really a demand for this.

PR #534, opened 2017-07-03
Add slugs to defaults

whoburg:
Fixes issue #511 by adding `slug` (and `slinch`, as suggested by @awcox21).


hgrecco:
Please put it in the corresponding group

hgrecco:
This should be in a group

hgrecco:
I am closing this for the time being. If the needed, please submit this patch but putting the unit within the correct group to avoid polluting the base group.

Issue #506, opened 2017-05-02
Millennial misspellings


oqibidipo:
The correct spellings for the thousand-year units are "millennium" and "millennia" with -nn- in the middle.

hgrecco:
Thanks. Can you submit a PR?

PR #542, opened 2017-08-14
Fix for issue #506

jaredbriskman:
This changes the spelling of **millennium** in the default definition file, raised by issue #506 . It is a breaking change for anything that was defined with the misspelled version, as these versions of the unit were removed. To maintain backwards compatibility, these could be added back as aliases. It also has not been changed in the default_en_0.6 definition file.


crusaderky:
Fixed by #811 a few months ago; please close

PR #549, opened 2017-08-31
Matplotlib support

### LINKED ISSUE #317: plotting with pint quantities
I am interested in a combination between pint and matplotlib.
- The axes would recognize the unit and format the ticks appropriately, including automatically dynamically changing from e.g. meters to kilometers if the range of values is large. It would be analogous to the way datetimes are currently handled.
- The abv. unit would be formatted into the tick labels, or optionally would be added to the axis label (e.g. "height (m)")
- Error bars would automatically be generated based on uncertainty
- Probably more stuff I haven't thought of. Suggestions welcome.

I imagine this would take a lot of work with the matplotlib package, but I thought I'd ask here as there might be more interest here. Has this been proposed or attempted? Does anyone else think this would be really cool?

dopplershift:
This adds a method to `UnitRegistry` that hooks up handlers for `Quantity` instances within matplotlib's unit machinery:
```python
import pint
ureg = pint.UnitRegistry()
ureg.setup_matplotlib()
```
This support can also be disabled with:
```python
ureg.setup_matplotlib(False)
```

Closes #317.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,6 +32,7 @@ Other contributors, listed alphabetically, are:
 * Richard Barnes <rbarnes@umn.edu>
 * Ryan Dwyer <ryanpdwyer@gmail.com>
 * Ryan Kingsbury <RyanSKingsbury@alumni.unc.edu>
+* Ryan May
 * Sundar Raman <cybertoast@gmail.com>
 * Tiago Coutinho <coutinho@esrf.fr>
 * Thomas Kluyver <takowl@gmail.com>
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,8 @@ Pint Changelog
 0.9 (unreleased)
 ----------------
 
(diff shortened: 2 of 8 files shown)


hgrecco:
Awesome!

One short thing. Can we add an image to the docs showing the output. I think it is worth doing it!

dopplershift:
@hgrecco So I used matplotlib's sphinx extension to have it render the plots when you build the docs. I can add a static image if you prefer, but it seemed better to have the image generate "live" when building docs.

hgrecco:
I think building it with the docs is a good idea as it help testing. But for this to work we will need to tell readthedocs that matplotlib, numpy and the sphinx extensions needs to be installed in order to build the docs. Can you create `requirements_docs.txt` file in the project root containing all the requirements? I will then configure readthedocs.

dopplershift:
Done. I think I got what's needed.

Issue #555, opened 2017-09-22
(Bug?) parsing expressions of the kind "600mg/24h"


alvaroabascar:
I've noticed that expressions of the kind `[number1] [units1] / [number2] [units2]` are incorrectly parsed: respecting the order of operations in this case leads to the tree `'(((num1 * unit1) / num2) * unit2)'`, but the actual tree (as intended by the human writing the expression) would be `'((num1 * unit1) / (num2 * unit2))'`.

Example:

```python
from pint import UnitRegistry
reg = UnitRegistry()
q = reg.parse_expression('600mg/24hours')
# out: <Quantity(25.0, 'hour * milligram')>
# output should be:  <Quantity(6.944444444444444e-09, 'kilogram / second')>
```

hgrecco:
This is a tough call because it has to be unambiguous. Maybe we can open this for a wider debate and hear opinions.

dalito:
Pint´s rule is that units are treated exactly like numbers, So pint behaves as specified. However, I agree with @alvaroabascar that the rule contradicts the expectations.  Even after inserting brackets the result stays the same:
```
>>> reg.parse_expression('600mg/(24hours)')
<Quantity(25.0, 'milligram / hour')>
```

@hgrecco do you have any specific ambiguous examples in mind?

hgrecco:
Let me elaborate just to add to the discussion. I think that it might be very confusing because we might have to put rules regarding spacing. We could change the precedence rules and say that units to the right of a number are one element. In this case: what next would mean (together, separated by space(s), by `*`)? Is always clear the intention of the user?

For example:

- `600mg/(24hours)` => `25 mg / hour` 
No issue here.
- `600mg/24hours` => `25 mg / hour`
Intuitively I would say `24hours` is one element.
- `600mg/24 hours`  => ?
Because hours is a unit, you could say `24 hours` is one element and the output should be `25 mg / hour`. However, following the easy to remember rule that units are exactly like numbers, the result is 25 mg * hour.
- `600mg/24*hours`  => ?
Same problem here. What is the intention of the user? 

If the last means 25 mg * hour, does it make sense that `600mg/24hours`, `600mg/24 hours` and `600mg/24*hours` give different results?

The approach in Pint is that we should not guess user intention because it might lead to subtle bugs.

dalito:
[...]

PR #557, opened 2017-10-06
Improve documentation for string parsing rules

dalito:
How pint interprets strings with a mixture of numbers and units has led to confusion, see #555. This adds a note and example to clarify pint´s rule.


jni:
Not a maintainer but 👍!

hgrecco:
thanks!

PR #562, opened 2017-10-19
Drop Support for Python 2.6, 3.0, 3.1, 3.2 and 3.3

### LINKED ISSUE #358: Drop Support for Python 2.6, 3.0, 3.1 and 3.2?
More and more scientific projects are dropping Python 2.6, 3.0, 3.1 and 3.2 support (Some are even dropping Python 3.3 support)

What are your thoughts about this?

hugovk:
Resolves #358 ("Drop Support for Python 2.6, 3.0, 3.1 and 3.2?").

That issue also points out that some scientific projects are also dropping 3.3, so an extra commit here drops 3.3. Let me know if you want to keep 3.3.

Here's the pip installs for Pint from PyPI for the last month, showing these versions have very little use:
```
$ pypinfo --percent --pip pint pyversion
python_version percent download_count
-------------- ------- --------------
2.7              73.5%         22,077
3.6              15.5%          4,667
3.5               8.5%          2,540
3.4               1.9%            563
3.3               0.4%            118
2.6               0.2%             72
3.7               0.0%              3

```

---

I also removed the branch restriction from .travis.yml so I could test my feature branch before submitting a potentially failing PR to you. Let me know if you want it reverting.

I just noticed I edited after `# Do not edit after this line` in .travis-full.yml. Should I revert those edits?


hgrecco:
Thanks for the awesome work. Please revert the edits after to `# Do not edit after this line` And I have one more question ... is there any benefit in dropping 3.3? It seems to me that it does not make a big impact in the code, right?

hugovk:
I'll revert those edits.

There's not such a big gain to be had from dropping 3.3 from a code point of view. One  benefit is an extra Travis slot! Another is not worrying about keeping compatibility, but I don't expect it would be that much extra work.

Here's the new stuff in 3.4: https://docs.python.org/3/whatsnew/3.4.html

dopplershift:
In my experience, there wasn't any code impact for supporting 3.3. Our rationale for dropping 3.3 was that it's not in heavy use and if we support it, we *have* to have an entry in the Travis matrix to test it--shortening Travis build time was more important to us.

hgrecco:
I think we should not drop 3.3 yet. I agree that is not heavily used, but it wouldn't be nice for those still using it (particularly if it not a maintenance burden for us). Python 3.3 has reached its end-of-life less than a [month ago](https://www.python.org/downloads/release/python-337/). I will say we wait another year.

hugovk:
See #567 which is the same as this but keeps 3.3.

hgrecco:
Thanks a lot for your work on this! (and for the reminder in the issues)

hugovk:
@hgrecco Did you mean to merge this one (drops 3.3) or #567 (keeps 3.3)?

PR #578, opened 2017-11-01
Fix to the prefixed condition for building the dimensional equivalent…

em1208:
… cache


em1208:
This pull request is to fix the prefixed boolean condition. I noticed that the boolean was not set correctly for a custom defined unit that started with capital letter E. Looking forward to your feedback and to see it merged soon :).

hgrecco:
I guess we can close this because is in the other one, right?

em1208:
Yes, thanks for merging @hgrecco!

PR #589, opened 2017-12-02
Add Fake Units

Modelmat:
Add support for more units, not much point except for theoretics

based off http://www.mindspring.com/~jimvb/unitsystem.htm

### DIFF
--- a/pint/default_en.txt
+++ b/pint/default_en.txt
@@ -9,26 +9,52 @@
 @end
 
 # decimal prefixes
-yocto- = 1e-24 = y-
-zepto- = 1e-21 = z-
-atto- =  1e-18 = a-
+deka- = 1e+1 = da- = deca-
+hecto- = 1e2 = h-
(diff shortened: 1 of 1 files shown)


onealj:
First time I've seen those units. I would certainly be surprised and confused with unknown and non-standard prefixes that we thought up by some random website, not universally used or taught, not published in textbooks, nor recognized by any international standards bodies.

I can't think of too many scenarios that extended prefixes would be helpful. Using an atomic scale ruler to measure the universe?

Modelmat:
That's a good point - I was needing them for Schwarzschild Radius Calculator.

Maybe add something like
```python
from pint import extended_units
```
Or something - I have no idea how to do that though

Modelmat:
@onealj Hello? Could we have that maybe?

hgrecco:
We can put them on an external file or maybe in a group

PR #590, opened 2017-12-05
Improve IPython pretty printers

tecki:
IPython has nice ways to make objects look nice. Pint already implements the `_repr_html_` and `_repr_latex_` methods. This pull request:

- adds a `_repr_pretty_` method. This does pretty printing on the IPython command line.
- lets all `_repr_` methods obey the `default_format`. If the `default_format` contains a `~` character, most likely the user always wants the short version of the units, in all versions of `_repr_`.
- writes numbers in exponential form as such, so `3.5e15` is formatted as 3.5×10<sup>15</sup>, in both L<sup>A</sup>T<sub>E</sub>X and HTML.

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -238,6 +238,12 @@ class _Quantity(SharedRegistryObject):
         else:
             return "${:L}$".format(self)
 
+    def _repr_pretty_(self, p, cycle):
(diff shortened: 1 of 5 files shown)


hgrecco:
Awesome PR! Thanks

PR #603, opened 2017-12-16
Added greek letter mu ( μ ) as an alias for micro

KOLANICH:


hgrecco:
Thanks again but I cannot merge it because bors choke due to the fact that is an unknown repositor

PR #626, opened 2018-03-23
Added warning for unsupported ufunc

### LINKED ISSUE #620: Need bigger warning for unsupported scipy functions
I was running an analysis that used scipy.special.erf and scipy.special.erfc and got bit by pint today.  It appears that the magnitude of the quantity was silently used by the function without proper dimensionality reduction.

While I understand that this is simply a byproduct of how pint works, I wish there had been a warning that this could occur.

For supported numpy ufunc's like exp the dimensionality reduction to dimensionless magnitude is performed.

 I'd rather either a safe option where Quantities are not usable by unsupported numerical libraries, or behavior that always resolves dimensionless magnitudes when cast to a float, or a big monster warning in the pint documentation that when sending quantities into unsupported functions, there's no telling what magnitude you are sending without explicitly doing `erf(x.m_as(''))` for example.

kanhua:
This PR attempts to close #620.

A ```UnitStrippedWarning``` is raised when an unsupported numpy ufunc is performed on the quantity.

This warning can be muted or amplified using ```warnings.simplefilter()```, for example,

```python
import warnings
import numpy as np
from pint import UnitRegistry, UnitStrippedWarning
warnings.simplefilter('ignore', category=UnitStrippedWarning)

ureg=UnitRegistry()
dist=24*ureg.meter

print(np.power(dis,2))

```

### DIFF
--- a/pint/__init__.py
+++ b/pint/__init__.py
@@ -18,7 +18,7 @@ import pkg_resources
 from .formatting import formatter
 from .registry import (UnitRegistry, LazyRegistry)
 from .errors import (DimensionalityError, OffsetUnitCalculusError,
-                   UndefinedUnitError)
+                   UndefinedUnitError, UnitStrippedWarning)
 from .util import pi_theorem, logger
 
 from .context import Context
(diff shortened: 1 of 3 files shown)


hgrecco:
Thanks

PR #637, opened 2018-04-28
Add Unit.from_ and Unit.m_from

virtuald:
Converts from a number or quantity to the unit (or the magnitude). This is most useful for when you aren't sure if the input is a Quantity.

While you could technically use `ureg.check()` for this, it's syntactically cleaner (and I suspect more performant) to use this:

```
ureg.inch.from_(1*foot)
ureg.inch.from_(1, strict=False)
```

Added tests and docstrings. I'm not quite sure where in the docs this should go, but I'm willing to add a blurb in there if you'll point me to the right place.

### DIFF
--- a/pint/testsuite/test_quantity.py
+++ b/pint/testsuite/test_quantity.py
@@ -247,6 +247,26 @@ class TestQuantity(QuantityTestCase):
             self.assertIsNot(r, q)
             self.assertIsNot(r._magnitude, a)
 
+    def test_convert_from(self):
+        x = self.Q_('2*inch')
+        meter = self.ureg.meter
+
+        # from quantity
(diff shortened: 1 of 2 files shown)


hgrecco:
In principle, I like this API as I think that is useful and does not overlap with an existing API. But I would like to have more opinions @cpascual  @dalito  @dopplershift

dopplershift:
Seems like a reasonable API addition to me. I’m sure there are some places in my code where this would simplify things.

hgrecco:
Please add documentation as well.

virtuald:
Where in the docs do you think documentation for this should go?

hgrecco:
@virtuald  Well ... I think that the docs require a general overhaul. The overall design was conceived for a much smaller pint (no pun intended). How about if you open an issue to discuss what reorganization could be made.

PR #642, opened 2018-05-29
support numpy's __array_ufunc__

tecki:
newer versions of numpy replace `__array_prepare__` and `__array_wrap__`
with the improved `__array_ufunc__`. This avoids some double-calculation
of data, and improves code readability.

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -1399,6 +1399,108 @@ class _Quantity(PrettyIPython, SharedRegistryObject):
 
     __array_priority__ = 17
 
+    def _call_ufunc(self, ufunc, *inputs, **kwargs):
+        # Store the destination units
+        dst_units = None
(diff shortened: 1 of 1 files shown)


tecki:
The coverage has decreased, but this is actually expected: we have now two code paths, one taken by old (pre-1.13) versions of numpy, the other ones taken by the new versions. So some code paths seem untested, but actually they are, by the other numpy version. I don't know how to deal with that.

wisp3rwind:
I think you could use `coverage erase`, `coverage run -p` and `coverage combine` as detailed in the links from this answer https://stackoverflow.com/a/19288889/3451198 to get accurate coverage reporting. Haven't done that myself before, so YMMV.

hgrecco:
Sorry it took so long and thanks for the great patch. Hopefully this will allow us to support more numpy functions better.

PR #652, opened 2018-07-13
Wraps handles keywords (closes #625, closes #651)

### LINKED ISSUE #625: Wraps crashes when using the same unit for different inputvalues
When I have this function:

```
@ureg.wraps(ureg.M, (ureg.degC, ureg.M, ureg.M))
def hco3_eq_from_tillmans_phreeqc(temperature, CO2, NaCl_conc=__NaCl_conc):
    pass
```
And call it as:

```
temp = ureg.Quantity(12, ureg.degC)
print(mdl.hco3_eq_from_tillmans_phreeqc(temp, 1*ureg.mM, NaCl_conc = 0.1 * ureg.mM))
```
Note how mM=mmol/L, a unit I added to my 

It gives me the following error:

```
Traceback (most recent call last):  File "C:\Users\korevma\.vscode\extensions\ms-python.python-2018.2.1\pythonFiles\PythonTools\visualstudio_py_launcher_nodebug.py", line 74, in run
    _vspu.exec_file(file, globals_obj)
  File "C:\Users\korevma\.vscode\extensions\ms-python.python-2018.2.1\pythonFiles\PythonTools\visualstudio_py_util.py", line 119, in exec_file
    exec_code(code, file, global_variables)
  File "C:\Users\korevma\.vscode\extensions\ms-python.python-2018.2.1\pythonFiles\PythonTools\visualstudio_py_util.py", line 95, in exec_code
    exec(code_obj, global_variables)
```

After removing the third argument, the wrapping works fine. Is it because the same unit is used twice? Does it have something to do with the usage of `unit_args_ndx` in the `_parse_wrap_args` function?

Im am using python 3

### LINKED ISSUE #651: ureg.wraps for optional arguments
How does ureg.wraps work for functions with optional arguments?

```python
from math import sqrt
from pint import _DEFAULT_REGISTRY
ureg = _DEFAULT_REGISTRY
Q_ = ureg.Quantity

@ureg.wraps(ureg.second, (ureg.meters, ureg.meters/ureg.second**2))
def calculate_time_to_fall(height, gravity=9.8):
    """Calculate time to fall from a height h.

    By default, the gravity is assumed to be earth gravity,
    but it can be modified.

    d = .5 * g * t**2
    t = sqrt(2 * d / g)
    """
    return sqrt(2 * height / gravity)

lunar_module_height = Q_(10, 'm')
#calculate_time_to_fall(lunar_module_height)
# 1.42 second

moon_gravity = Q_(1.625, 'm/s^2')
calculate_time_to_fall(lunar_module_height, moon_gravity)
# should return 3.51 seconds, but it errors instead
```

jondoesntgit:
This passed the test that I wrote, and didn't break anything else when I tested it on my Python 3 install.

I import `inspect`, which is only available in Python3. I'm not sure how to cleanly support Python2 at the same time...

### DIFF
--- a/docs/wrapping.rst
+++ b/docs/wrapping.rst
@@ -151,7 +151,7 @@ airmass::
 Optional arguments
 ------------------
 
-For a function with named keywords with optional values, use a tuple for all
+In Python 3, for a function with named keywords with optional values, use a tuple for all
 arguments:
 
 .. doctest::
@@ -202,7 +202,7 @@ You can use more than one label:
     ... def some_function(x, y, z):
     ...     pass
 
-With optional arguments:
+With optional arguments in Python 3:
 
 .. doctest::
 
--- a/docs/wrapping.rst
+++ b/docs/wrapping.rst
@@ -151,7 +151,7 @@ airmass::
 Optional arguments
 ------------------
 
-In Python 3, for a function with named keywords with optional values, use a tuple for all
+For a function with named keywords with optional values, use a tuple for all
 arguments:
 
 .. doctest::
@@ -202,7 +202,7 @@ You can use more than one label:
     ... def some_function(x, y, z):
     ...     pass
 
-With optional arguments in Python 3:
+With optional arguments
 
 .. doctest::
 
--- a/docs/wrapping.rst
+++ b/docs/wrapping.rst
@@ -148,6 +148,37 @@ airmass::
     def solar_position(lat, lon, press, tamb, timestamp):
         return zenith, azimuth, airmass
 
+Optional arguments
+------------------
+
+For a function with named keywords with optional values, use a tuple for all
+arguments:
+
+.. doctest::
+
+    >>> @ureg.wraps(ureg.second, (ureg.meters, ureg.meters/ureg.second**2))
+    ... def calculate_time_to_fall(height, gravity=Q_(9.8, 'm/s^2'), verbose=False):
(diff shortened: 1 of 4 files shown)


jondoesntgit:
So it seems like `signature`'s [`BoundArguments.apply_defaults`](https://docs.python.org/3/library/inspect.html#inspect.BoundArguments.apply_defaults) method is new in Python 3.5 according to the [docs](https://docs.python.org/3/library/inspect.html#inspect.BoundArguments.apply_defaults).

jondoesntgit:
AFAIK, this should be good to go on my end. @hgrecco might want to tidy some things up according to his tastes.

This feature is only available in Python 3. There may be a way to port it to 2.7, but I'm not clamoring for that feature in 2.7. If anyone else is, they can ping me.

hgrecco:
It looks really good. 

Can you explore using [funcsigs](https://pypi.org/project/funcsigs/) to keep backwards compatibility with 2.7? I think it should be as simply as adding a conditional requirement in `setup.py` and importing this in the [compat](https://github.com/hgrecco/pint/blob/master/pint/compat/__init__.py).

AFAIK it does not implement `apply_defaults` but backwards compatibility to 2.7 would be nice.

jondoesntgit:
It's now backwards compatible.

hgrecco:
Do you think we can do something like this for `check` as well?

Issue #659, opened 2018-07-20
DeprecationWarning callable is None when running unit tests


labrys:
When running unit tests, `test_nonnumeric_magnitudes` raises a `DeprecationWarning`.

hgrecco:
I think this is solved, right? Closing. Reopen if needed.

PR #661, opened 2018-07-21
Fix `test_nonnumeric_magnitudes`

labrys:
The test for non-numeric magnitudes was using `assertRaises` incorrectly.  While the context-manager form should be preferred, with the given style it should be passed the function to test, not the result of the function.

```python
# Wrong
self.assertRaises(RuntimeError, self.compareQuantity_compact(x,x))

# Better
self.assertRaises(RuntimeError, self.compareQuantity_compact, x, x)

# Best
with self.assertRaises(RuntimeError):
    self.compareQuantity_compact(x,x)
```

While fixing this I found that `RuntimeError` was not being raised, instead a `RuntimeWarning` was being issued.  Since `unittest.TestCase.assertWarns` was not added until Python 3.2 (see [docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns)), I'm testing for the warning as recommended in the [Python 2.7 docs](https://docs.python.org/2/library/warnings.html#testing-warnings).

Fixes hgrecco/pint#659

### DIFF
--- a/pint/__init__.py
+++ b/pint/__init__.py
@@ -26,7 +26,7 @@ from .context import Context
 
 try:                # pragma: no cover
     __version__ = pkg_resources.get_distribution('pint').version
-except:             # pragma: no cover
+except TypeError:  # pragma: no cover
     # we seem to have a local copy not installed without setuptools
     # so the reported version will be unknown
     __version__ = "unknown"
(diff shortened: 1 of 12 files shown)


hgrecco:
I think this PR is no longer necessary due to the recent merge of #910

Issue #374, opened 2016-04-28
Request for more pint-specific exceptions during pint_eval


moltman:
Recently ran into an error which raised a general Exception and bubbled up our application. Is it possible to create for Pint-specific exceptions for this so we don't have to do a blanket catch Exception at such a low level.

We already wrap certain Pint calls with try-except on TokenError, UndefinedUnitError, etc which helps us very specifically know when we have unit errors, but we're adverse to try: except Exception.

File ".../lib/python2.7/site-packages/pint/unit.py", line 1271, in parse_expression
    return build_eval_tree(gen).evaluate(lambda x : self._eval_token(x, case_sensitive=case_sensitive,
  File ".../lib/python2.7/site-packages/pint/pint_eval.py", line 88, in evaluate
    raise Exception('missing unary operator "%s"' % op_text)
Exception: missing unary operator "/"

hgrecco:
That is indeed a nice adition. Maybe you can look at [pint eval](https://github.com/hgrecco/pint/blob/master/pint/pint_eval.py) to propose some specific errors.

moltman:
Will evaluate and send along some thoughts, happy to contribute if you agree with the proposal

hgrecco:
Done

PR #679, opened 2018-08-22
pint_eval uses DefinitionSyntaxError in place of general Exception

IamJeffG:
This allows clients to catch pint-specific errors, rather than a blanket `catch Exception`.

Partial progress towards #374 -- `context.py` and `quantity.py` still each raise a general `Exception`.

Discussion on #374 allowed for introduction of new custom exception types, but my understanding of these error cases seem to be described by the extant `DefinitionSyntaxError`.  Correct me if I'm wrong.

### DIFF
--- a/pint/pint_eval.py
+++ b/pint/pint_eval.py
@@ -15,6 +15,8 @@ import operator
 
 import token as tokenlib
 
+from .errors import DefinitionSyntaxError
+
 # For controlling order of operations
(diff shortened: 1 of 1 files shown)


cpascual:
IMHO the `DefinitionSyntaxError` describes these exceptions well.
Also the change seems innocuous to me, so I am directly merging it.

bors r+

Issue #666, opened 2018-07-25
Use 'µm' as abbreviation of micrometer?


ericpre:
Is there a reason to use `'um'` and not `'µm'`, considering that the last one is the correct abbreviation according to the SI units? 

```python
import pint
ureg = pint.UnitRegistry()

q = ureg('µm')
print('{~}'.format(a))
```
will output '1 um'.

jondoesntgit:
Three things:

The 'mu' symbol may not work in all environments. It should be fine in more modern systems where unicode is ubiquitous, but we probably can't make that guarantee for older systems.

Secondly:

Using 'u' as a substitue for 'micro' is pretty standard among the engineers here at Stanford, so it isn't that unusual...

Thirdly:

Pint coded in such a way as to give the end user a lot of control. If you want to change the 'u' to a 'µ', edit [this line](https://github.com/hgrecco/pint/blob/8edd45f35599071fac19f784619cc1871baa6c7e/pint/default_en.txt#L18) in your `default_en.txt` file.

ericpre:
Thanks for the quick reply and the hint.
This would work on my local install but it wan't when it is used in another library and particularly for unit testing?

francisco-dlp:
In this case wouldn't it suffice to replace
```
micro- = 1e-6  = u- = µ-
```
with
```
micro- = 1e-6  = µ- = u-
``` 
to make µ the default abbreviation? I understand why ``u`` was a better choice before Python 3 was widespread and all OSs had good UTF support, but why keeping this constraint today?

jondoesntgit:
Perhaps legacy compatibility?

AFAIK (from discussions on #577 and #652), @hgrecco and others want to keep backwards compatibility for 2.7, the packages that have pint as a dependency, etc...

ericpre:
I just had a look at the [string formatting section](http://pint.readthedocs.io/en/latest/tutorial.html#string-formatting) of the pint tutorial and tested the following in python 2.7:

```python
from pint import UnitRegistry
ureg = UnitRegistry()
accel = 1.3 * ureg['meter/second**2']

from __future__ import unicode_literals # as documented in the tutorial
print('The pretty representation is {:P}'.format(accel))
```
output:
```
The pretty representation is 1.3 meter/second²
```

And the following works also fine:
```python
a = 1.0 * ureg(u'Å')
print(u'The pretty representation is {:~P}'.format(a))
```
output:
```
The pretty representation is 1.2 Å
```
[...]

PR #680, opened 2018-08-23
Use µ as default abbreviation for micro.

ericpre:
As suggested and discussed in #666, this PR sets `µ` as default abbreviation for micro.

### DIFF
--- a/pint/default_en.txt
+++ b/pint/default_en.txt
@@ -15,7 +15,7 @@ atto- =  1e-18 = a-
 femto- = 1e-15 = f-
 pico- =  1e-12 = p-
 nano- =  1e-9  = n-
-micro- = 1e-6  = u- = µ-
+micro- = 1e-6  = µ- = u-
 milli- = 1e-3  = m-
 centi- = 1e-2  = c-
 deci- =  1e-1  = d-
--- a/pint/definitions.py
+++ b/pint/definitions.py
@@ -11,6 +11,7 @@
 
 from __future__ import (division, unicode_literals, print_function,
                         absolute_import)
+import sys
 
 from .converters import ScaleConverter, OffsetConverter
 from .util import UnitsContainer, _is_dim, ParserHelper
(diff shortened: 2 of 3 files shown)


hgrecco:
Thanks for the PR. I have a more general questions. Should we provide a Unicode and non Unicode version of symbols, abbreviation and unit names?

jondoesntgit:
Something akin to sensing whether babel is installed?

Would this mean we would have two default.txt files, one with Unicode and the other without?

cpitclaudel:
Merging this would be great, though a more general way to change a unit's default abbreviation would be great.

hgrecco:
I am still doubtfull about unicode support. How to distinguish/select?

ericpre:
There are only a few symbols or abbreviations with the unicode issue in python 2.7, one way to deal with it is simply to [ignore these when reading definitions](https://github.com/hgrecco/pint/pull/680/files#diff-d36b35485d40005c8f0bf6dcfd421a39R48).

This fixes the following on python 2.7:
```python
import pint
ureg = pint.UnitRegistry()
a = 1.0E-6*ureg.hbar
print("{:~P}".format(a))
```
which is raising an `UnicodeEncodeError` on master since it can't read the alias (`ħ`); same for `Ω` (ohm), `σ` (Stefan Boltzmann constant), `Å` (ångström).
There is also an `UnicodeEncodeError` with `P/` (printers_pica) on python 2.7 and I don't know what is this abbreviation, so I would prefer not to touch it.

The tests on python 2.7 are failing because the expected `UnicodeEncodeError` is not raised! :( Locally, the error is nicely raised...

ericpre:
@hgrecco, do you think the fix of this PR to deal with unicode character on python 2.7 is sensible? If so I will work out the test failure.

hgrecco:
Sorry for the delay. It looks good to me. Go ahead

ericpre:
Thanks @hgrecco for the feedback! The test failures are fixed now.

hgrecco:
thanks

PR #683, opened 2018-08-26
BUG: Fix using Quantity with numpy.searchsorted

dopplershift:
The method on Quantity was not up to date with the numpy set of arguments, which [as of 1.7](https://docs.scipy.org/doc/numpy/reference/generated/numpy.searchsorted.html?highlight=searchsorted#numpy.searchsorted) includes a `sorter` argument. Adding this allows the numpy function to properly delegate to the Quantity method.


hgrecco:
Thanks for the PR

PR #664, opened 2018-07-22
Code Quality

labrys:
This fixes the following:

- Comparison to None
- Tests for membership
- Tests for object identity
- Bare except
- Old style classes
- Builtins being shadowed
- File encoding
- Deprecated function call
- Trailing comma on bare tuple

Issue #645, opened 2018-06-04
Pandas support using new API


wisp3rwind:
The most recent pandas release (0.23) introduces [extension types](http://pandas.pydata.org/pandas-docs/version/0.23/extending.html#extending-extension-types) which I believe provide a natural way to integrate pint types into pandas dataframes on a per row/column basis. Is this a feature that could be merged here (assuming no breakage when older/no pandas is installed)?
If so, I'd be interested in researching that possibility further and maybe even implementing it myself. As a disclaimer, I've only started using pint quite recently and am not very proficient with pandas either and I'm not sure when I really get the time to do this. In any case, if someone else beats me to coding the feature, that'd be perfect, too :)

Related issues (which I _haven't_ read fully):
- #401 
- https://github.com/pandas-dev/pandas/issues/10349,
- an early user of the new API is [cyberpandas](https://github.com/ContinuumIO/cyberpandas)

hgrecco:
I would be very interested to support this. Maybe you can describe here how you envision this. But I would really like to see this happening.

bilderbuchi:
@hgrecco is this closed as wontfix, works already, or refer to pint-pandas?

Issue #401, opened 2016-07-12
Consider support for pandas


tomchor:
I'd like to suggest as an enhancement that support for [Pandas](http://pandas.pydata.org/) be considered. I see the need for this is several places over the internet ([this one](http://stackoverflow.com/questions/19251766/pandas-interpolate-data-with-units), [this one](https://www.reddit.com/r/Python/comments/1yxchj/pint_a_python_units_library_with_numpy_support/), [this one](http://scicomp.stackexchange.com/questions/10770/pandas-limitations-and-its-alternatives-in-python) and [this one](http://stackoverflow.com/questions/16409050/associating-units-with-pandas-dataframe) for example) and I also have discussed this need in personal communications. 

Some points:
- I know that, with the huge amount of functionality that pandas has, full support is very hard, but maybe the main functions (mean, std, etc.) and multiplication/division functionality
- However, given that Numpy is already supported, an expansion to pandas wouldn't have to start from scratch, so there's that
- Furthermore, since I have a lot of interest in a Pandas-with-units extension for my own personal projects, I could help out with the development once we agree on how it should be done (as I'm sure other people would also like to do)

Some examples of how it could work would be:

```
df = pd.DataFrame(np.random.randn(6,3))
df = df * (ureg.m, ureg.km, ureg.cm)
df.sum().sum()
df.mean().sum()
```

Currently, probably because of the numpy support, `df.sum().sum()` works, but `df.mean().sum()` doesn't. Furthermore, in this way, each element has its own unit:

```
                       0                          1                          2
0   -1.14025689919 meter    2.09836283028 kilometer  0.979569146473 centimeter
1    1.00550215217 meter   0.447878034084 kilometer  0.831879097545 centimeter
2   -1.37479008756 meter  0.0440848611227 kilometer    1.2712599879 centimeter
3   0.250507523208 meter   -1.12502135888 kilometer  0.262470368675 centimeter
4  -0.329479372063 meter  -0.201112807862 kilometer   1.10957183866 centimeter
5      1.049932299 meter   0.109837949121 kilometer  -1.42261917121 centimeter
```

Maybe a better way would be something like

```
<QuantitySet(
          0         1         2
0 -1.140257  2.098363  0.979569
1  1.005502  0.447878  0.831879
2 -1.374790  0.044085  1.271260
3  0.250508 -1.125021  0.262470
4 -0.329479 -0.201113  1.109572
5  1.049932  0.109838 -1.422619
[...]

PR #684, opened 2018-08-28
Add pandas support

znicholls:
This pull request adds pandas support to pint (hence is related to #645, #401 and pandas-dev/pandas#10349).

An example can be seen in `example-notebooks/basic-example.ipynb`.

It's a little bit hacksih, feedback would be greatly appreciated by me and @andrewgsavage. One obvious example is that we have to run all the interface tests with `pytest` to fit with `pandas` test suite, which introduces a dependency for the CI and currently gives us this awkward testing setup (see the alterations we had to make to `testsuite`). This also means that our code coverage tests are fiddly too.

If you'd like us to squash the commits, that can be done.

If pint has a linter, it would be good to run that over this pull request too as we're a little bit all over the place re style.

Things to discuss:

- [x]  general feedback and changes
- [x] test setup, especially need for pytest for pandas tests and hackish way to get around automatic discovery
- [x] squashing/rebasing
- [x] linting/other code style (related to #664 and #628: we're happy with whatever, I've found using an automatic linter e.g. black and/or flake8 has made things much simpler in other projects)
- [x] including notebooks in the repo (if we want to, I'm happy to put them under CI so we can make sure they run)
- [x] setting up the docs correctly

### DIFF
--- a/docs/pandas.rst
+++ b/docs/pandas.rst
@@ -1,0 +1,329 @@
+.. _pandas:
+
+Pandas support
+==============
+
+It is convenient to use the `Pandas package`_ when dealing with numerical data, so Pint provides `PintArray`. A `PintArray` is a `Pandas Extension Array`_, which allows Pandas to recognise the Quantity and store it in Pandas DataFrames and Series.
+
+For this to work, we rely on `Pandas Extension Types`_ which are still experimental. As a result, we currently have to build the latest version of Pandas' master branch from source as documented in the `Pandas README`_.
+
+Basic example
+-------------
+
+This example gives you the basics, but is slightly fiddly as you are not reading from a file. A more normal use case is given in `Reading a csv`_.
+
+To use Pint with Pandas, as stated above, firstly ensure that you have the latest version of Pandas installed. Then import the relevant packages and create an instance of a Pint Quantity:
+
+.. doctest::
+
+   >>> import pandas as pd
+   >>> import numpy as np
+   >>> import pint
+   >>> from pint.pandas_interface import PintArray
+
+   >>> ureg = pint.UnitRegistry()
+   >>> Q_ = ureg.Quantity
+
+.. testsetup:: *
+
+   import pandas as pd
+   import numpy as np
+   import pint
+   from pint.pandas_interface import PintArray
+
+   ureg = pint.UnitRegistry()
+   Q_ = ureg.Quantity
+
+Next, we can create a DataFrame with PintArray's as columns
+
+.. doctest::
+
+   >>> torque = PintArray(Q_([1, 2, 2, 3], "lbf ft"))
+   >>> angular_velocity = PintArray(Q_([1000, 2000, 2000, 3000], "rpm"))
+   >>> df = pd.DataFrame({"torque": torque, "angular_velocity": angular_velocity})
+   >>> print(df)
+    torque angular_velocity
+   0      1             1000
+   1      2             2000
+   2      2             2000
+   3      3             3000
+
+
+Operations with columns are units aware so behave as we would intuitively expect
+
+.. doctest::
+
+   >>> df['power'] = df['torque'] * df['angular_velocity']
+   >>> print(df)
+     torque angular_velocity power
+   0      1             1000  1000
+   1      2             2000  4000
+   2      2             2000  4000
+   3      3             3000  9000
+
+
+Each column can be accessed as a Pandas Series
+
+.. doctest::
+
+   >>> print(df.power)
+   0    1000
+   1    4000
+   2    4000
+   3    9000
+   Name: power, dtype: pint
+
+
+Which contains a PintArray
+
+.. doctest::
+
+   >>> print(df.power.values)
+    PintArray([1000 foot * force_pound * revolutions_per_minute,
+               4000 foot * force_pound * revolutions_per_minute,
+               4000 foot * force_pound * revolutions_per_minute,
+               9000 foot * force_pound * revolutions_per_minute],
+              dtype='pint')
+
+
+Which contains a Quantity
+
+.. doctest::
+
+   >>> print(df.power.values.data)
+   [1000 4000 4000 9000] foot * force_pound * revolutions_per_minute
+
+
+Pandas Series accessors are provided for most Quantity properties and methods, which will convert the result to a Series where possible.
+
+.. doctest::
+
+   >>> print(df.power.pint.dimensionality)
+   [length] ** 2 * [mass] / [time] ** 3
+
+   >>> print(df.power.pint.to("kW"))
+   0    0.14198092353610375
+   1      0.567923694144415
+   2      0.567923694144415
+   3     1.2778283118249338
+   Name: power, dtype: pint
+
+
+Standard pint conversions can still be performed on the underlying quantity, and will still return a quantity.
+
+.. doctest::
+
+   >>> print(df.power.values.data.to("kW"))
+   [0.14198092 0.56792369 0.56792369 1.27782831] kilowatt
+
+Reading a csv
+-------------
+
+Thanks to the DataFrame accessors, reading from files with unit information becomes trivial. The DataFrame accessors make it easy to get to PintArrays.
+
+Setup
+~~~~~
+
+Here we create the DateFrame and save it to file, next we will show you how to load and read it.
+
+We start with an DateFrame with column headers only.
+
+.. doctest::
+
+   >>> speed = [1000, 1100, 1200, 1200]
+   >>> mech_power = [np.nan, np.nan, np.nan, np.nan]
+   >>> torque = [10, 10, 10, 10]
+   >>> rail_pressure = [1000, 1000000000000, 1000, 1000]
+   >>> fuel_flow_rate = [10, 10, 10, 10]
+   >>> fluid_power = [np.nan, np.nan, np.nan, np.nan]
+   >>> df_init = pd.DataFrame({"speed": speed, "mech power": mech_power, "torque": torque, "rail pressure": rail_pressure, "fuel flow rate": fuel_flow_rate, "fluid power": fluid_power,})
+   >>> print(df_init)
+      speed  mech power  torque  rail pressure  fuel flow rate  fluid power
+   0   1000         NaN      10           1000              10          NaN
+   1   1100         NaN      10  1000000000000              10          NaN
+   2   1200         NaN      10           1000              10          NaN
+   3   1200         NaN      10           1000              10          NaN
+
+Then we add a column header which contains units information
+
+.. doctest::
+
+   >>> units = ["rpm", "kW", "N m", "bar", "l/min", "kW"]
+   >>> df_to_save = df_init.copy()
+   >>> df_to_save.columns = pd.MultiIndex.from_arrays([df_init.columns, units])
+   >>> print(df_to_save)
+     speed mech power torque  rail pressure fuel flow rate fluid power
+       rpm         kW    N m            bar          l/min          kW
+   0  1000        NaN     10           1000             10         NaN
+   1  1100        NaN     10  1000000000000             10         NaN
+   2  1200        NaN     10           1000             10         NaN
+   3  1200        NaN     10           1000             10         NaN
+
+Now we save this to disk as a csv to give us our starting point.
+
+.. doctest::
+
+   >>> test_csv_name = "pandas_test.csv"
+   >>> df_to_save.to_csv(test_csv_name, index=False)
+
+Now we are in a position to read the csv we just saved. Let's start by reading the file with units as a level in a multiindex column.
+
+.. doctest::
+
+   >>> df = pd.read_csv(test_csv_name, header=[0,1])
+   >>> print(df)
+     speed mech power torque  rail pressure fuel flow rate fluid power
+       rpm         kW    N m            bar          l/min          kW
+   0  1000        NaN     10           1000             10         NaN
+   1  1100        NaN     10  1000000000000             10         NaN
+   2  1200        NaN     10           1000             10         NaN
+   3  1200        NaN     10           1000             10         NaN
+
+Then use the DataFrame's `pint.quantify` method to convert the columns from `np.ndarray`s to PintArrays, with units from the bottom column level.
+
+.. doctest::
+
+   >>> df_ = df.pint.quantify(ureg, level=-1)
+   >>> print(df_)
+       speed mech power torque    rail pressure fuel flow rate fluid power
+   0  1000.0        nan   10.0           1000.0           10.0         nan
+   1  1100.0        nan   10.0  1000000000000.0           10.0         nan
+   2  1200.0        nan   10.0           1000.0           10.0         nan
+   3  1200.0        nan   10.0           1000.0           10.0         nan
+
+
+As previously, operations between DataFrame columns are unit aware
+
+.. doctest::
+
+   >>> df_['mech power'] = df_.speed*df_.torque
+   >>> df_['fluid power'] = df_['fuel flow rate'] * df_['rail pressure']
+   >>> print(df_)
+       speed mech power torque    rail pressure fuel flow rate       fluid power
+   0  1000.0    10000.0   10.0           1000.0           10.0           10000.0
+   1  1100.0    11000.0   10.0  1000000000000.0           10.0  10000000000000.0
+   2  1200.0    12000.0   10.0           1000.0           10.0           10000.0
+   3  1200.0    12000.0   10.0           1000.0           10.0           10000.0
+
+
+The DataFrame's `pint.dequantify` method then allows us to retrieve the units information as a header row once again
+
+.. doctest::
+
+   >>> print(df_.pint.dequantify())
+                      speed                              mech power  \
+     revolutions_per_minute meter * newton * revolutions_per_minute
+   0                 1000.0                                 10000.0
+   1                 1100.0                                 11000.0
+   2                 1200.0                                 12000.0
+   3                 1200.0                                 12000.0
+
+             torque rail pressure fuel flow rate          fluid power
+     meter * newton           bar liter / minute bar * liter / minute
+   0           10.0  1.000000e+03           10.0         1.000000e+04
+   1           10.0  1.000000e+12           10.0         1.000000e+13
+   2           10.0  1.000000e+03           10.0         1.000000e+04
+   3           10.0  1.000000e+03           10.0         1.000000e+04
+
+
+
+This allows for some rather powerful abilities. For example, to change single column units
+
+.. doctest::
+
+   >>> df_['fluid power'] = df_['fluid power'].pint.to("kW")
+   >>> df_['mech power'] = df_['mech power'].pint.to("kW")
+   >>> print(df_.pint.dequantify())
+
+                      speed mech power         torque rail pressure  \
+     revolutions_per_minute   kilowatt meter * newton           bar
+   0                 1000.0   1.047198           10.0  1.000000e+03
+   1                 1100.0   1.151917           10.0  1.000000e+12
+   2                 1200.0   1.256637           10.0  1.000000e+03
+   3                 1200.0   1.256637           10.0  1.000000e+03
+
+     fuel flow rate   fluid power
+     liter / minute      kilowatt
+   0           10.0  1.666667e+01
+   1           10.0  1.666667e+10
+   2           10.0  1.666667e+01
+   3           10.0  1.666667e+01
+
+
+or the entire table's units
+
+.. doctest::
+
+   >>> print(df_.pint.to_base_units().pint.dequantify())
+
+               speed                          mech power  \
+     radian / second kilogram * meter ** 2 / second ** 3
+   0      104.719755                         1047.197551
+   1      115.191731                         1151.917306
+   2      125.663706                         1256.637061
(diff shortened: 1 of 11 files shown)


znicholls:
I don't understand why this is showing a drop in coverage, https://coveralls.io/builds/18732747 shows an increase...

znicholls:
> Is it because more lines have been added, but only tested in the pandas job, so the non pandas job coverages have fallen?

Seems likely, no idea how to fix it...

znicholls:
> Is it because more lines have been added, but only tested in the pandas job, so the non pandas job coverages have fallen?

Nice, found a solution.

jondoesntgit:
Good work guys!

Yes, coverage is a ratio of (tested lines)/(total lines). That goes down if you write lines of code without adding tests.

znicholls:
@hgrecco any possibility you could let us know whether this is even a feasible merge or whether we should start building a separate repo please?

hgrecco:
This is a long awaited feature. So I am eager to merge it , In general, I like the code and the API. But It is long PR and I need a extra couple of days to play with it.

znicholls:
Great thanks. Take as long as you need! For our research group, if we have a rough timeline then we can work around it (trying to work with no information is always the most difficult)

jondoesntgit:
Will probably want to include some docs. The example ipynb file is good, but folks who read the readthedocs page might not know to look for it on the Github repo.

znicholls:
> Will probably want to include some docs

Yep sounds good! Neither @andrewgsavage nor I could get our heads around how the docs work, do you have a resource you can point to/can you explain where the best place to add them is?

dalito:
To add to the current documentation structure I suggest to create a new file `pint/docs/pandas.rst` and reference it in `pint/docs/index.rst`. You may use `pint/docs/numpy.rst` as guideline for the new `pandas.rst`.

znicholls:
> To add to the current documentation structure I suggest to create a new file pint/docs/pandas.rst and reference it in pint/docs/index.rst. You may use pint/docs/numpy.rst as guideline for the new pandas.rst.

Cheers, have had a crack. Some things aren't that nice but hopefully with some guidance we can turn this into a nice pull request. @andrewgsavage jump in with any additions you want!

andrewgsavage:
Nice, docs! There's a few parts I should probably explain better. Will do so when I have a chance.

Have you tried using df.to_html? That might display the dfs like jupyter does.

Noticed that the dfs are now showing with the unit in the column. I think this is because of a change made to __array__ without changing _format_values. I'm not sure which I prefer - it's nice to see the units but that reduces the space available for columns/is distracting.

jondoesntgit:
Good job with the docs!

I think I’m partial to showing units in the column header. Perhaps on a line below in parentheses (maybe printing using a beeline character). That keeps the data frame horizontally compact.

I think there are pros and cons to wherever you put the units though.

znicholls:
> Have you tried using df.to_html? That might display the dfs like jupyter does.

Yep I thought about that. I didn't know what this 'doctest' stuff was doing though so got nervous and didn't... 

> I think I’m partial to showing units in the column header. Perhaps on a line below in parentheses (maybe printing using a beeline character). That keeps the data frame horizontally compact.

This sounds super nice. If we follow that convention then we force users to only have a single unit for a given column. My only wondering was whether we want that to be the case (I think it's the most sensible, but it's not the only option). We could, for example, allow a column to have different units within the column (e.g. the first value could be in metres, the second in centimetres, the third in millimetres etc.). That may quickly become a nightmare thought...

hgrecco:
I agree that the docs are very nice. A few suggestions regarding the docs, but also the API:

- In my experience, loading from a package resource brings trouble. Either load it from an URL or better, import the example dataframe (this can be done using the python machinery for doing so which is better than hard coding the location), then explain how to save it and then load it.

- Does changing the default formatter affects how it is shown in the dataframe. If so, it would be nice to show it. If not, maybe we should implement it.

andrewgsavage:
> I think I’m partial to showing units in the column header. Perhaps on a line below in parentheses (maybe printing using a beeline character). That keeps the data frame horizontally compact.

Same, but I don't believe this is possible. The .dequantify() adds an extra row to the columns which was the closet I could get to.

Will need to look into the formatting again, as it might be possible to show the units in the first row of data only, eg.
meter * Newton \n
123

> This sounds super nice. If we follow that convention then we force users to only have a single unit for a given column
Data is stored as a pint quantity, which handles the unit converting and such. When you set a value in the PintArray to a different unit to the underlying quantity (in ._data), the value will be converted to the original unit.

znicholls:
> Data is stored as a pint quantity, which handles the unit converting and such. When you set a value in the PintArray to a different unit to the underlying quantity (in ._data), the value will be converted to the original unit.

Awesome

jondoesntgit:
It might require redefining __repr__ or .to_html

andrewgsavage:
> Does changing the default formatter affects how it is shown in the dataframe. If so, it would be nice to show it. If not, maybe we should implement it.

What do you mean by the default formatter? Pandas uses a separate formatter that we pass text/values into, _formatting_values.

andrewgsavage:
> Awesome

Hmm, if you're questioning that it I take it isn't obvious, maybe this needs an example/reference to the pint tutorial?

znicholls:
> Hmm, if you're questioning that it I take it isn't obvious, maybe this needs an example/reference to the pint tutorial?

Reference to the pint tutorial I think. I've never done it (can I admit that now...)

znicholls:
> In my experience, loading from a package resource brings trouble. Either load it from an URL or better, import the example dataframe (this can be done using the python machinery for doing so which is better than hard coding the location), then explain how to save it and then load it.

Latest commit should fix this I believe

znicholls:
> In my experience, loading from a package resource brings trouble. Either load it from an URL or better, import the example dataframe (this can be done using the python machinery for doing so which is better than hard coding the location), then explain how to save it and then load it.

are you ok with having the test file in `pint/testsuite/test-data` or do you want that gone too?

andrewgsavage:
> It might require redefining repr or .to_html

I'm assuming you're referring to pandas' repr/._to_html. These aren't things ExtensionArrays can change so I can't see any way of going about it.

It might be worth submitting an issue to pandas to allow ExtensionArrays to display metadata in column headers.

andrewgsavage:
> Will need to look into the formatting again, as it might be possible to show the units in the first row of data only, eg.
meter * Newton \n
123

Couldn't get that working, options are either as current (with units displayed in each cell), or just the value and no unit shown.

docs/pandas.rst
```diff
@@ -0,0 +1,333 @@
...
+   2  2 foot * force_pound  2000 revolutions_per_minute
+   3  3 foot * force_pound  3000 revolutions_per_minute
+
+                                                 power
+   0  1000 foot * force_pound * revolutions_per_minute
+   1  4000 foot * force_pound * revolutions_per_minute
+   2  4000 foot * force_pound * revolutions_per_minute
+   3  9000 foot * force_pound * revolutions_per_minute
+
+Data accessing is a little bit awkward (fixing this would be helpful), but can be done as shown below
```
jreback: you could register an accessor to do this 
eg

df.power.pint.to()

becomes possible

docs/pandas.rst
```diff
@@ -0,0 +1,333 @@
...
+   1  1000000000000.0 bar  10.0 liter / minute  nan kilowatt
+   2           1000.0 bar  10.0 liter / minute  nan kilowatt
+   3           1000.0 bar  10.0 liter / minute  nan kilowatt
+
+
+
+.. doctest::
+
+   >>> from os.path import join
+   >>> df=pd.read_csv(join("..", "pint", "testsuite", "test-data", "pint_test_data.csv"), header=[0,1])
```
hgrecco: I think here should be now just: pandas_test.csv
So that people load what they have just saved

pint/pandas_interface/pint_array.py
```diff
@@ -0,0 +1,609 @@
...
+              integers or None
+            * ndarray: A 1-d boolean NumPy ndarray the same length as 'self'
+        Returns
+        -------
+        item : scalar or PintArray
+        """
+        if is_integer(item):
+            return self._data[item]
+
+        return type(self)(self._data[item])
```
hgrecco: Why type(self) and not self.__class__ ?

znicholls:
> I was thinking in something along the lines that scikit image uses to have demo data (see here). But I think your current solution is very good.

ahah I see. Well I'm happy to move to something like scikit image (tomorrow) if we think such a `data` module would be useful. Otherwise I'm happy to leave it if we think this is likely to be the only need for it for now.

pint/pandas_interface/pint_array.py
```diff
@@ -0,0 +1,609 @@
...
+                 space).join(u("%s=%s") % (k, v) for k, v in attrs)
+
+        res = u("%s(%s%s)") % (klass, data, prepr)
+
+        return res
+
+    def __array__(self, dtype=None, copy=False):
+    # this is necessary for some pandas operations, eg transpose
+    # note, units will be lost
+        if dtype == None:
```
hgrecco: Shouldn't this be `dtype is None`

pint/pandas_interface/pint_array.py
```diff
@@ -0,0 +1,609 @@
...
+            dtype=object
+        if type(dtype) == str:
```
hgrecco: why (here and below) are you using `dtype ==` instead of isinstance

hgrecco: Also, pint defines `compat.string_types` just in case you need to quicklly check for string and unicode in Py2 and str in Py3

pint/quantity.py
```diff
@@ -64,6 +63,19 @@ def wrapped(self, *args, **kwargs):
...
     return wrapped
 
+def check_implemented(f):
+    def wrapped(self, *args, **kwargs):
+        other=args[0]
+        if other.__class__.__name__ in ["PintArray", "Series"]:
+            return NotImplemented
+        # pandas often gets to arrays of quantities [ Q_(1,"m"), Q_(2,"m")]
+        # and expects Quantity * array[Quantity] should return NotImplemented
+        elif type(other)==list and isinstance(other[0], type(self)):
```
hgrecco: Same question here, why not `isinstance` here?

pint/util.py
```diff
@@ -615,6 +615,10 @@ def _check(self, other):
         same unit registry.
 
         """
+        if type(other).__name__ == 'PintArray':
+            print("notimpl")
```
hgrecco: I guess this should be removed for the final version :-)

hgrecco: And also, I do not understand the purpose of this. Why not directly adding the SharedRegistryObject to all pandas-pint related objects. Is this because everything is stored as strings in a dataframe?

hgrecco:
As I said, I really like it. I have one extra conceptual question due to my ignorance of the Pandas hook method. Is there a method or attribute that can be implemented in Quantity so that is automatically called by pandas when constructing a dataframe?. So that this:

```python
>>> torque = PintArray(Q_([1, 2, 2, 3], "lbf ft"))
>>> angular_velocity = PintArray(Q_([1000, 2000, 2000, 3000], "rpm"))
>>> df = pd.DataFrame({"torque": torque, "angular_velocity": angular_velocity})
>>> print(df)
```

becomes

```python
>>> torque = Q_([1, 2, 2, 3], "lbf ft")
>>> angular_velocity = Q_([1000, 2000, 2000, 3000], "rpm")
>>> df = pd.DataFrame({"torque": torque, "angular_velocity": angular_velocity})
>>> print(df)
```

If this exists, maybe we can implemented conditionally in `Quantity` to conditionally return a `PintArray`

Or maybe there is a way to register a converting method from Quantity to PintArray so that instead of calling PintArray every time we just do:

```python
import pint
pint.register_in_pandas()
```
(name open for discussion)

**Important** This is by no means a blocker. Just a way to make the connection between pandas a pint more friendly

hgrecco:
I do not think that the `data` module is necessary. But I do like that now that we are able to load/save dataframes we might start proving in a different pint project some curated datasets :-D

pint/pandas_interface/pint_array.py
```diff
@@ -0,0 +1,609 @@
...
+            if len(units) > 1:
+                # need to work out a way to test this
+                raise TypeError("The units of all quantities are not the same"
+                                " for input {}".format(values))
+
+            magnitudes = [v.magnitude for v in values]
+
+            return self._dtype.type(magnitudes, values[0].units)
+
+        return NotImplementedError
```
jbrockmendel: not “Raise”?

znicholls:
> I do not think that the data module is necessary. But I do like that now that we are able to load/save dataframes we might start proving in a different pint project some curated datasets :-D

Perfect!

docs/pandas.rst
znicholls: ok sounds sensible (if @andrewgsavage can do it as this is his domain).

My comment was more directed at the fact that you have to do `df.power.values.data` to get your pint array back rather than just `df.power.values` but maybe that can't be helped

Makefile
```diff
@@ -0,0 +1,14 @@
+# Not a real makefile, just the steps
```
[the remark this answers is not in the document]
znicholls: Yep sure. I could also get rid of it completely as all the 'instructions' are in the CI?

docs/pandas.rst
znicholls: oops that was meant to be deleted, good pick up!

docs/pandas.rst
jreback: df.power.values should be a PintArray otherwise your impl is incorrect

pint/pandas_interface/pint_array.py
znicholls: No idea, we have lots of oddities in our style (which could be fixed with a linter, see discussion point above?). Changed now.

pint/pandas_interface/pint_array.py
znicholls: yep, fixed

pint/pandas_interface/pint_array.py
znicholls: fixed

pint/util.py
znicholls: no idea... @andrewgsavage ?

pint/pandas_interface/pint_array.py
znicholls: oops, fixed

znicholls:
> Is there a method or attribute that can be implemented in Quantity so that is automatically called by pandas when constructing a dataframe?

Maybe new issue?

pint/pandas_interface/pint_array.py
```diff
@@ -157,7 +157,7 @@ def __array__(self, dtype=None, copy=False):
             dtype = object
         if isinstance(dtype, string_types):
             dtype = getattr(np, dtype)
-        if dtype == object:
+        if dtype == object:  # it seems impossible to avoid using this...
```
hgrecco: Then it should be `dtype is object`

pint/quantity.py
znicholls: fixed

docs/pandas.rst
znicholls: ok then I guess it has to stay like this

pint/pandas_interface/pint_array.py
znicholls: oddly even that seems to fail...

pint/util.py
andrewgsavage: This can be deleted, it was added when I was trying to get quantity + pintarray (or any operation) to return NotImplemented

andrewgsavage:
>Is there a method or attribute that can be implemented in Quantity so that is automatically called by pandas when constructing a dataframe?

That's what the PintArray and PintType are, they're pandas ExtensionArray and ExtensionType which tell pandas how to handle the data. 

To make Quantity satisfy the ExtensionArray requirements might be possible, but I think it would be difficult to keep the pandas specific things seperate.

 I think at one point (before CI tests) I had a version that allowed:
````
PA_= PintArray
torque = PA_([1, 2, 2, 3], "lbf ft")

df = pd.DataFrame({"torque": torque})
````
The initialization code was very confusing though. Maybe this could work well?
````
PA_= PintArray.from_values_and_unit
torque = PA_([1, 2, 2, 3], "lbf ft")

df = pd.DataFrame({"torque": torque})
````
where .from_values_and_units creates a quantity then creates a PintArray with that quantity

> Or maybe there is a way to register a converting method from Quantity to PintArray

Maybe, that might be a bit confusing intialising a Quantity to get a PintArray.

> Maybe new issue?

Yea I think so. There's a few other similar cases where it would be nice for it work in a certain way, but it doesn't. Could we put non-blocking issues there?

hgrecco:
@andrewgsavage I understand the purpose of PintArray and PintType. I would say that is not a good idea to make Quantity an ExtensionArray and ExtensionType. My feeling is that it will make things much more complicated. We might do it the future (Maybe even having a QuantityType and QuantityArray) but this will require major refactoring

But some libraries provide a different approach to interoperability. IPython is a good example. It does not force you to inherit from anything to provide nice html represenation you just need to provide a method. Another example is Python itself. If you need a boolean of you object just provide `__bool__`. So I was wondering if there is a method that Pandas will look for to have a compatible type. AFAIK this is not possible.

Ps.- Move all the non-blocking comments to a new issue so we can merge this PR before friday.

docs/pandas.rst
andrewgsavage: > df.power.pint.to()

Yea had series accessors in there, but not documented/tested. They're now in the docs and in a PR with tests.

znicholls:
ok @hgrecco I think with the latest series accessors we have covered off most of the issues before you take a final lookover? Looking at the checklist, the things still to discuss are:

- test setup:
    - do you mind that we need pytest for the pandas tests?
    - are you ok with our hackish way to avoid the pandas tests when they're not relevant?
- squashing/rebasing
    - would you like us to squash/rebase this commit before it's merged or will you just do that automatically?
- including notebooks in the repo
    - would you like us to remove the example notebooks from the repo?
    - if not, would you like us to put the example notebooks under CI?

hgrecco:
> test setup: 
I actually like pytest. I am fine to keep it as long as it does not harm the usual testsuite. To do so, please raise a clear exception when it is not installed or when pandas is not installed.

> squashing/rebasing 
Please squash it. My suggestion is that you do it in 4 commits: code, tests, docs, exampla

> including notebooks in the repo 
I would like to keep them. But maybe we should put it in a specific folder.

znicholls:
great will do tomorrow

znicholls:
> To do so, please raise a clear exception when it is not installed or when pandas is not installed.

Any suggestions on how to do this without putting the entire pandas test stuff inside a try/except block?

hgrecco:
The following might work.

1. Define a `HAS_PANDAS` in https://github.com/hgrecco/pint/blob/master/pint/compat/__init__.py in the same way we have defined `HAS_NUMPY`. (you can also calle it `HAS_PROPER_PANDAS` until pandas has stabilized this feature like we did for babel)
2. Define a `requires_pandas` decorator. in your testsuite using [this](https://docs.pytest.org/en/latest/skipping.html#id1). Normally this would go in https://github.com/hgrecco/pint/blob/master/pint/testsuite/helpers.py but this would require to have pytest installed (which I would rather not for the general case)
3. put import pandas in a try except import and print a clear error "This pint feature requires ..." or similar.

Having `HAS_PANDAS` (or similar) is useful in general for further extensions as a unique place to define the compatibility.

What I do not know how to include the pytests in a unittest.TestSuite. If we achive that, then running `python setup.py test` will run all tests

znicholls:
ok I think I've done all the steps we wanted now. The last big problem is that in the process of rebasing, I seem to have removed any mention of @andrewgsavage's contribution. I'm not very comfortable with this as all of the interface code is his (I've basically only done the test code). Do you have a suggestion for how to fix this whilst also doing such a mega rebase?

> What I do not know how to include the pytests in a unittest.TestSuite. If we achieve that, then running python setup.py test will run all tests

I don't think that's possible either...

znicholls:
@hgrecco this is good from our end (tests pending)

hgrecco:
I have asked bors to merge it. Big thanks and congratulations for such wonderful work and thanks everybody else that helped by reviewing and commenting on the code. I think that now we deserve a new release

hgrecco:
Done! I have opened #693 to track the new release

TomAugspurger:
Would you all be OK with adding `Pint`'s extension array to the pandas pandas ecosystem page at http://pandas.pydata.org/pandas-docs/version/0.24/ecosystem.html#extension-data-types?

hgrecco:
Sure. Please add something like beta or prerelease or provisional. I think we need to test it better on our side via a broader use.

znicholls:
I think we have this here https://github.com/pandas-dev/pandas/pull/22582 so maybe just need to re-open and update?

TomAugspurger:
Yes, that'd be perfect if you could.

znicholls:
@hgrecco @andrewgsavage to be clear, Pint no longer has the pandas support and we're moving it into https://github.com/hgrecco/pint-pandas ?

hgrecco:
True. But it would be an extension of pint. As soon as we think that is stable we can add it to pint/setup.py as optional.

znicholls:
Ok great. I'll update pandas-dev/pandas#22582 later today

Issue #685, opened 2018-08-29
Alias hz to Hz


Benjamin-Lee:
Was just surprised that `hz` is not a valid unit. I understand that the standard unit is `Hz`, but `hz` unambiguously means `Hz`.

hgrecco:
From Wikipedia:

As with every International System of Units (SI) unit named for a person, the first letter of its symbol is upper case (Hz). However, when an SI unit is spelled out in English, it is treated as a common noun and should always begin with a lower case letter (hertz)—except in a situation where any word in that position would be capitalized, such as at the beginning of a sentence or in material using title case.

PR #686, opened 2018-08-29
Fix for #685

Benjamin-Lee:


cpascual:
This PR refers to #685 
Considering @hgrecco's reply (https://github.com/hgrecco/pint/issues/685#issuecomment-416900394) I personally think that this PR should not be merged: even if "hz" is unambiguous, other units may be ambiguous if unofficial capitalization is used and pint should try to avoid inconsistencies in its treatment of the various units.

Also, this PR is doing some unrelated redefinition of angstroms.

dalito:
I agree with @cpascual - please reject the PR #685 and close this issue. Case is so important in the context of units, that we should prefer correctness over convenience.

hgrecco:
I also agree. When there is a clear correct way we should do it that way. No guessing has always been an underlying motto of pint.

PR #702, opened 2018-09-21
Fix issue #701: Import breaks with pandas<=0.22.0 installed

### LINKED ISSUE #701: Import breaks with pandas<=0.22.0 installed
When trying to import pint as pulled from here in an enviroment with pandas 0.22.0 installed, I get an `AttributeError`. This seems to be caused by the attribute pandas.core.arrays not present there.

*Expected Behaviour*: pint detects incompatible version of pandas and deactivates pandas extension.
*Actual Behaviour*: pint cannot be imported, import breaks with `AttributeError`.

```
>>> import pandas as pd
>>> pd.__version__
'0.22.0'
>>> hasattr(pd.core, "arrays")
False
>>> import pint
  File "/home/hartelt/repos/pint/pint/__init__.py", line 18, in <module>
    from .formatting import formatter
  File "/home/hartelt/repos/pint/pint/formatting.py", line 16, in <module>
    from .babel_names import _babel_units, _babel_lengths
  File "/home/hartelt/repos/pint/pint/babel_names.py", line 10, in <module>
    from pint.compat import HAS_PROPER_BABEL
  File "/home/hartelt/repos/pint/pint/compat/__init__.py", line 130, in <module>
    HAS_PROPER_PANDAS = hasattr(pd.core.arrays.base, "ExtensionOpsMixin")
AttributeError: module 'pandas.core' has no attribute 'arrays'
```

As far as I see, the appropriate fix would be refining the detection of the `HAS_PROPER_PANDAS` flag in `pint/compat/__init__.py`, for example by changing the breaking line to: 
```
HAS_PROPER_PANDAS = hasattr(pd.core, "arrays") and hasattr(pd.core.arrays.base, "ExtensionOpsMixin")
```
I'll fork and try this soon. Please confirm that this would be the correct usage of the `HAS_PROPER_PANDAS` flag, since I don't have much experience with the pint code.

hartelt:
This fixes #701 as outlined in issue comments. Please review this with respect to the usage of the `HAS_PROPER_PANDAS` flag, which seems to indicate if pandas is present in a compatible version, which is exactly the problem here.

### DIFF
--- a/pint/compat/__init__.py
+++ b/pint/compat/__init__.py
@@ -127,7 +127,8 @@ if not HAS_PROPER_BABEL:
 try:
     import pandas as pd
     HAS_PANDAS = True
-    HAS_PROPER_PANDAS = hasattr(pd.core, "arrays") and hasattr(pd.core.arrays.base, "ExtensionOpsMixin")
+    HAS_PROPER_PANDAS = hasattr(pd.core, "arrays") and \
+                        hasattr(pd.core.arrays.base, "ExtensionOpsMixin")
 except ImportError:
     HAS_PROPER_PANDAS = HAS_PANDAS = False
 
--- a/pint/compat/__init__.py
+++ b/pint/compat/__init__.py
@@ -128,9 +128,9 @@ try:
     import pandas as pd
     HAS_PANDAS = True
     HAS_PROPER_PANDAS = (
-            hasattr(pd.core, "arrays")
-            and hasattr(pd.core.arrays.base, "ExtensionOpsMixin")
-        )
+        hasattr(pd.core, "arrays")
+        and hasattr(pd.core.arrays.base, "ExtensionOpsMixin")
+    )
 except ImportError:
     HAS_PROPER_PANDAS = HAS_PANDAS = False
 
--- a/pint/compat/__init__.py
+++ b/pint/compat/__init__.py
@@ -127,8 +127,10 @@ if not HAS_PROPER_BABEL:
 try:
     import pandas as pd
     HAS_PANDAS = True
-    HAS_PROPER_PANDAS = hasattr(pd.core, "arrays") and \
-                        hasattr(pd.core.arrays.base, "ExtensionOpsMixin")
+    HAS_PROPER_PANDAS = (
+            hasattr(pd.core, "arrays")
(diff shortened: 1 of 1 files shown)


pint/compat/__init__.py
```diff
@@ -127,7 +127,8 @@ def _to_magnitude(value, force_ndarray=False):
 try:
     import pandas as pd
     HAS_PANDAS = True
-    HAS_PROPER_PANDAS = hasattr(pd.core.arrays.base, "ExtensionOpsMixin")
+    HAS_PROPER_PANDAS = hasattr(pd.core, "arrays") and \
+                        hasattr(pd.core.arrays.base, "ExtensionOpsMixin")
 except ImportError:
```
znicholls: can you change to 

```python
HAS_PROPER_PANDAS = (
        hasattr(pd.core, "arrays") 
        and hasattr(pd.core.arrays.base, "ExtensionOpsMixin")
    )
```

please to avoid the `\` continuation

hartelt: Sure thing. Done and pushed.

znicholls: my bad, I think the code review failed because of indenting. I suspect it should be 

```python
try:
    import pandas as pd
    HAS_PANDAS = True
    HAS_PROPER_PANDAS = (
        hasattr(pd.core, "arrays")
        and hasattr(pd.core.arrays.base, "ExtensionOpsMixin")
    )
except ImportError:
HAS_PROPER_PANDAS = HAS_PANDAS = False
```

hartelt: I'm not sure... the indentation shouldn't make a difference with the brackets right? The build actually passed. Only the coverage review complains about decreased coverage... What's this about?

znicholls: The code still works but bors failed it for code reviews which I think is because my review has still not been switched. I'll fix that now, but that indentation looks awkward so if you have 3 minutes to fix that would be great (but as you say not essential and probably better done with an automated linter in future).

The coverage thing is weird and we don't have a solution yet, see https://github.com/lemurheavy/coveralls-public/issues/1190

hartelt: Alright. Done.

Issue #39, opened 2013-07-30
numpy: matrix multiplication by left give wrong results


canesin:
Hi, this is a issue related to #37 , multiplication by left don't work with a numpy matrix, as matrices are not commutative for multiplication. It don't fail, it give wrong results.

https://gist.github.com/canesin/6115159

hgrecco:
Review once that  #905 is implemented

jthielen:
Based on https://docs.scipy.org/doc/numpy-1.17.0/reference/generated/numpy.matrix.html, NumPy matrices should no longer be used, instead, regular arrays should be used. So, can this be closed as wontfix due to upstream deprecation?

hgrecco:
Agreed!

Issue #37, opened 2013-07-24
Operating elementwise with numpy masked arrays


canesin:
It is not working as suposed with masked arrays, it creats objects by elements.

https://gist.github.com/canesin/6072390

hgrecco:
Review once that  #905 is implemented

PR #704, opened 2018-09-23
[WIP] fix #86c, #37, #39, added support to many numpy functions

impact27:
- added a `asQuantity` function to transform an array into a `Quantity` (a bit like numpy's asarray)
- Allowed transformation of array of `Quantity` to `Quantity`
- A dimensionless 0 should be compared and added (Remove the distinction between 0 and dimensionless zero)
- A `Quantity` is not an `ndarray`! is has an `ndarray`. Therefore it should not respond to `['__array_interface__', '__array_struct__' , '__array__', '_data']`
- added `__array_ufunc__` to handle numpy function


hgrecco:
This has been superseed by #905 and related

Issue #711, opened 2018-10-06
Proposal: Allow ureg.wraps to convert string arguments


lewisamarshall:
Hi! I really like pint, and I'm designing a package that uses it for scientific computations. 
I'd like to propose an extension that I would find useful. This extension would automatically convert strings to Quantities on wrapping functions. Example code is below. 

```
@ur.wraps(ur.second, ur.meter, strict=False)
def test(l):
    return l*2

test('1 mm')
# Actual result:    <Quantity(1 mm1 mm, 'second')>
# Desired result: <Quantity(0.002, 'second')>
```

This behavior could be placed behind another keyword argument, for example convert_strings=True. 

The use case here is as follows:
- My module is designed to be part of an scientific stack workflow, where users import objects to perform specific calculations. 
- I want all calculations to use the same unitregistry. To do this, I would need users to import my_module.ureg to make Quantities to pass into my functions. It would also require that the end user understand pint. 
- Allowing strings would mean that users could directly write their units into the calculation inputs without an intermediate layer. The example from the "Wrapping and Checking" documentation would be pendulum_period("10 cm"). This behavior seems very intuitive to me. 

If you think this is a good idea, I'd be happy to submit a PR for it.

jondoesntgit:
As far as using the same registry, I usually

   from pint import _DEFAULT_REGISTRY as u

jondoesntgit:
Can you describe your workflow a little more? If users already understand Pint, can they use Quantity objects? Are you parsing text files?

lewisamarshall:
Sure. The module I'm talking about is `ebb`, which is a pipe flow calculation tool. 

The tutorial is here. 
https://nbviewer.jupyter.org/github/lewisamarshall/ebb/blob/tutorial_update/tutorial.ipynb

The intended audience is fluidic engineers with novice or higher python skill. Anyone interested in performing a fluid flow calculation should be able to import the module and start modeling in a jupyter notebook.

This module is a work in progress. It has the functionality that I'm interested in, but through a different mechanism. At the time I started working on it, I wasn't aware of ureg.wraps.

Typical code that would use the proposed functionality is below. 

```
from ebb.pipes import CircularPipe; from ebb.fluids import Water
pipe = CircularPipe(radius='1 mm', length='10 cm')
pipe.fluid(Water)
print(pipe.maximum_velocity(pressure='0.1 psi'))
```

This code would also work if the strings were replaced by Quantities. The output of functions are all Quantities.

[...]

PR #723, opened 2018-10-21
Wrap converts strings

lewisamarshall:
When strict=True, ureg.wraps is capable of converting strings to Quantities on function input. 

Also includes updates to tests, docs, and error messages. 

This PR is responsive to issue #711.


hgrecco:
Closed by #962

PR #732, opened 2018-11-04
Add missing closing parenthesis

jmartens:
Add missing closing parenthesis to the solar_position example in Multiple arguments or return values section


hgrecco:
Thanks!

Issue #725, opened 2018-10-23
Different-registries error after pickle dump-load


jbriales:
I have saved a pint Quantity into a file with `pickle.dump(...)` then loaded it in a different program execution with `pickle.load(...)`.
In both program executions I create the UnitRegistry in exactly the same way.
However, when comparing or mixing a Quantity that is obtained from the .pkl file via `pickle.load(...)` and a Quantity that is created in the current program execution via the UnitRegistry, I get the error:
> ValueError: Cannot operate with Unit and Unit of different registries.

Following the instructions in the [tutorial](https://pint.readthedocs.io/en/latest/serialization.html), 
my understanding is that in this circumstances everything should work fine. What may be the problem or my misunderstanding?

iankharris:
After some digging, it looks like this is very straightforward:
>>> pint.set_application_registry(ureg)

hgrecco:
This should be fixed in master.

Issue #716, opened 2018-10-14
Define get_application_registry


Silmathoron:
I have trouble understanding the point of the ``set_application_registry`` given that there is no ``get_application_registry``: what good is it for the user if he has to know he should use a "hidden" variable ``_APP_REGISTRY`` or ``_DEFAULT_REGISTRY`` afterwards?
I am also confused about the existence of these two variables, especially since only ``_DEFAULT_REGISTRY`` is used in the package, while the user can solely change ``_APP_REGISTRY``...

It feels like I am missing something: could you clarify the purpose of all three?

hgrecco:
The problems we are trying to solve are the following:
1. Unpickling requires an instantiated registry
2. Working with multiple packages requires a shared registry.

The way it works now:
- `_DEFAULT_REGISTY` is a Lazy Registry (i.e. it is only instantiated when first used). 
- `_APP_REGISTRY` is the registry used for unpickling operations. It defaults to `_DEFAULT_REGISTRY`
- `set_application_registry` is the way you change the default registry.

Part of this solution came when we were understading the proper use of Pint as it is unique in the way it define units. In most packages, units are part of singleton  (i.e. they are just variables in a module). 
The reason that there is no `get_application_registry` is that the user knows the hierarchy of their package and therefore knows where to get pint registry from.

After using Pint in multiple applications, I think the right way would be:
1. Raise an exception if the user tries to unpickle without setting the application registry (i.e. do not guess!)
2. keep `set_application_registry`
3. remove `_DEFAULT_REGISTY`
4. Maybe, add `get_application_registry`

Silmathoron:
Hi @hgrecco thank you for your answer, I completely agree with your 4 items.
Do you want me to implement it and make a PR?

Silmathoron:
Ow, I thought point 1 was something you meant for people using it, not something you wanted to implement. I can implement a warning, though... (I think it is better than a breaking change)

Silmathoron:
I am currently implementing these changes, however, I have a small question: why is ``logger`` assigned a ``NullHandler``? I thought I would use it to print the warning... should I use the ``warnings`` module?

Also, it is easy to put a warning in ``__reduce__`` to warn about pickling, but warning on unpickling is more tricky.
To the best of my knowledge, I will have to either wrap ``pickle.load`` and ``pickle.loads`` or implement ``__setstate__`` and ``__getstate__``; do you have a preferred option?

Silmathoron:
[...]

PR #738, opened 2018-11-23
Add get_application_registry (plus prettify)

Silmathoron:
This PR adresses #716 and helps solve #725.

I also moved the ``__repr__`` to ``PrettyIPython`` so as not to shadow the pretty representation.

Because the doc uses the tuples with pickle, I also added a warning when making tuples, though I don't understand why you do not advocate for direct pickling...
Checks are made in ``UnitsContainer`` to avoid issues with multiple inheritance hierarchies in quantities and units

### DIFF
--- a/pint/__init__.py
+++ b/pint/__init__.py
@@ -75,6 +75,14 @@ def _build_unit(units):
     return _APP_REGISTRY.Unit(units)
 
 
+def get_application_registry():
+    """
+    Get the application registry which is used for unpickling operations.
+    """
+    assert isinstance(registry, UnitRegistry)
+    return _APP_REGISTRY
+
+
 def set_application_registry(registry):
     """Set the application registry which is used for unpickling operations.
 
--- a/pint/__init__.py
+++ b/pint/__init__.py
@@ -79,7 +79,6 @@ def get_application_registry():
     """
     Get the application registry which is used for unpickling operations.
     """
-    assert isinstance(registry, UnitRegistry)
     return _APP_REGISTRY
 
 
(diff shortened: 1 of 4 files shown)


pint/__init__.py
```diff
@@ -75,6 +75,14 @@ def _build_unit(units):
     return _APP_REGISTRY.Unit(units)
 
 
+def get_application_registry():
+    """
+    Get the application registry which is used for unpickling operations.
+    """
+    assert isinstance(registry, UnitRegistry)
```
cpascual: maybe I am missing something, but... where is the `registry` variable defined?

Silmathoron: Obviously you're not and I was so focused on the pickle part that I did not even check the get_app_registry because it looked "so simple to do", and forgot to delete this line...

Silmathoron:
I'm having trouble understanding the check that fails with travis... could you help me out?

iankharris:
While we're at it, why don't we update the docs to mention the `pint.set_application_registry` function?  It's by far the easiest way pint to correctly unpickle pint units (in my opinion).  It seems like such a trivial change that it's almost a shame to make a pull-request for this, but I really wish the documentation had alluded to this function before I trying to make the `to_tuple` method described in the docs to work with dask (it doesn't).

Silmathoron:
Ok, I'll try to update the doc at some point.
It's going to take me some time to correct the pandas errors because of [this](https://github.com/hgrecco/pint/blob/master/pint/compat/__init__.py#L131)... which I think should be changed to ``>= '0.24'`` anyway.

However, I still don't know why this call to data causes the pytest error on [L967](https://travis-ci.org/hgrecco/pint/jobs/458794765#L967)... and I don't think this has anything to do with my PR.

Silmathoron:
Hey, there!
Now that pandas 0.24 is out, I thought I would check again what's going on, but it seems the notebook is not working with 0.24... I get some nasty accessors and delegated objects... can anyone tell me what's the status of the pandas support?

hgrecco:
Closing this due to #880

PR #741, opened 2018-12-01
WIP: Cleanup pandas tests

znicholls:
An attempt to clean the pandas tests up a bit. @hgrecco I'm still a bit unsure whether to use `HAS_PANDAS` or set `pd = None` if there's no Pandas available. It's not that clear which convention to use where and it seems like `pd = None` implies there is no pandas available (like `np = None` implies no numpy available).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,8 +55,7 @@ install:
 
 script:
   # if we haven't installed pandas, don't use pytest to run tests
-  - if [[ $PANDAS == '0' ]]; then python -bb -m coverage run -p --source=pint --omit="*test*","*compat*" setup.py test; fi
-  # - if [[ $PANDAS == '0' ]]; then python -bb -m coverage run -p --source=pint --omit="*test*","*compat*","*pandas*" setup.py test; fi
+  - if [[ $PANDAS == '0' ]]; then python -bb -m coverage run -p --source=pint --omit="*test*","*compat*","*pandas*" setup.py test; fi
   # if we're doing the pandas tests and hence have pytest available, we can
   # simply use pytest to run all the tests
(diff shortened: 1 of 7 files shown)


znicholls:
Yep. The easy question: how are e.g. `HAS_NUMPY` and `pd.compat.np` related?

The harder question,  is it possible to clean up the pandas related tests. I wanted to get rid of this massive `if` clause that we currently have but can't work out how. I've arrived at trying to use `SkipTest`, but that's failing for Python2.7 and Python3.3 and I have no idea why/how to fix it. Any ideas are appreciated but it may just be overkill.

hgrecco:
The idea is that for a feature `X` there is an `HAS_X` boolean that tell us that the feature is available. Additionally (and certainly redundant) we provide all the classes / modules equal to `None` just to allow any rogue import to work. But the ideas is that everybody test agains `HAS_X`

A way out of this problen (annoying but that would work) is to migrate to UnitTest. The conditional testing there works flawlesly and you can use the current NumPy test in pint as  example.

znicholls:
> A way out of this problen (annoying but that would work) is to migrate to UnitTest

Sadly I can't do that as all the underlying pandas tests are written with pytest. 

What do you think about moving all the pandas stuff out into a separate, wrapper repo in the style of cyberpandas https://github.com/ContinuumIO/cyberpandas? It would mean that Pint stays clean and dependency free and all the nasty dependency handling can go elsewhere?

hgrecco:
Sounda good to me. Maybe we can create pint associates packages like
pint-pandas

Hernán

andrewgsavage:
Give us a shout when you do that as I've made a branch that's passing pandas tests again.

hgrecco:
I created a stub repo in https://github.com/hgrecco/pint-pandas

hgrecco:
I think we can close this, right?

znicholls:
Yes if you're happy to go down the route of a new repository

PR #754, opened 2019-01-17
add pint-pandas notebook docs

andrewgsavage:
This seemed too easy!

I've copied the notebook from pint-pandas. It isn't rerun here, it renders using the outputs saved in the .ipynb file.


andrewgsavage:
This needs some installation instructions

PR #755, opened 2019-01-18
Moved the metadata into setup.cfg

KOLANICH:


crusaderky:
There have been major changes to setup.py when we discarded python < 3.6; please look them up in master and reflect them in this PR.

Also, there's a minor conflict with #930, which lets setup.py dynamically generate the version parameter.

hgrecco:
Appart from the conflict with #930, it is still unclear to me if `setup.cfg` is suggested. While I do prefer declarative syntax for these type of files, PyPA tutorials still use the old fashion setup.py https://packaging.python.org/tutorials/packaging-projects/

KOLANICH:
>While I do prefer declarative syntax for these type of files, PyPA tutorials still use the old fashion setup.py 

forget pypa tutorials, they are obsolete. the main packaging tool is setuptools and it supports setup.cfg since december of 2016.

hgrecco:
Don't get me wrong. I am not against this PR. As I said, I do like declarative. But being supported is not the same as being recommended or in the roadmap. pypa is the steering comittee of the python packaging ecosystem.

KOLANICH:
PyPA is promoting multiple systems using `pyproject.toml` (there have been no unified metadata format for `pyproject.toml` standardized), such as `flit`, `poetry` and others in order to encourage competition between build systems in a hope it will cause their improvement. IMHO: it was a big mistake and the result of it is that one build system is not enough and in order to build packages using `poetry` a user has to use venvs because it is a political decision of `poetry` project that it works exclusively through venvs. The tools like [`dephell`](https://github.com/dephell/dephell) have appeared that convert between build systems but ... they have too much dependencies.

`setuptools` currently has no support of moving the metadata into `pyproject.toml` even if it has been requested.

setup.py
```diff
@@ -1,71 +1,18 @@
...
-        "Programming Language :: Python :: 3.8",
-    ],
-    python_requires=">=3.6",
-    install_requires=["setuptools"],
-    extras_require={
-        "numpy": ["numpy >= 1.14"],
-        "uncertainties": ["uncertainties >= 3.0"],
-    },
-)
+    setup(use_scm_version=True)
```
crusaderky: can't use_scm_version go into setup.cfg too?

setup.py
```diff
@@ -1,71 +1,18 @@
...
-        version += ".dev0"
+        reload(sys).setdefaultencoding("UTF-8")
+    except:
+        pass
 
+    try:
+        from setuptools import setup
+    except ImportError:
+        print('Please install or upgrade setuptools or pip to continue')
+        sys.exit(1)
```
crusaderky: Looks like the merge got messed up

setup.py
KOLANICH: no, `setuptools_scm` is a plugin. `setup.cfg` has a hardcoded schema. Supporting plugins in `setup.cfg` has been requested but  noone has implemented this.

setup.py
```diff
@@ -1,51 +1,6 @@
 #!/usr/bin/env python
-
+# -*- coding: utf-8 -*-
```
crusaderky: Unnecessary; utf-8 is the default in Python 3 - please remove

crusaderky:
I tested it; I can confirm these flags work:

- name
- version
- python_requires
- install_requires
- extras_require
- package_data
- test_suite
- include_package_data
- [bdist_wheel] universal = 1

I didn't test most of the metadata section as I couldn't find a way that doesn't require uploading to pypi (please suggest).

It doesn't make me too comfortable that, if there's a spelling error in setup.cfg, pip will say absolutely nothing...

dopplershift:
`twine check` will check the metadata in built wheels.

I can say that MetPy went to putting as much as possible in `setup.cfg` and has a minimal `setup.py` now. For me it was philosophical: I'm not a fan of packages executing arbitrary Python code when I pip install them. (I can't change other packages, but I can change my own.)

hgrecco:
I am ready to merge as soon as the conflicts are gone.

crusaderky:
@KOLANICH  if you merge this https://github.com/KOLANICH/pint/pull/1 we're good to go

orsinium:
> The tools like dephell have appeared that convert between build systems but ... they have too much dependencies.

I don't think it's a big deal. Most of dependencies of dephell are packages from dephell's own ecosystem. Installing DepHell on TravisCI without cache takes 17 seconds. The recommended way to install dephell is `curl -L dephell.org/install | python3` that installs DepHell into its own isolation that keeps the global environment clean. So, it's more about decoupling than a fat dependency tree.

hgrecco:
Thanks a lot!

Issue #761, opened 2019-02-03
missing list of quantities constructor


deniz195:
Using ping frequently, it appears to me that handling of lists of quantities can be easily improved.

Currently the following code
```
import pint
ureg=pint.UnitRegistry()

a_of_mag = [10, 20, 30]
quants = ureg.Quantity(a_of_mag, units='mm')
print(f'{quants:~}')
```
yields:
```
[10 20 30] mm
```

In analogy, one would expect (propose) the same outcome for
```
a_of_quants = [ureg('1cm'), ureg('2cm'), ureg('3cm')]
quants = ureg.Quantity(a_of_quants, units='mm')
print(f'{quants:~}')
```
But unfortunately this yields:
```
ValueError: setting an array element with a sequence.
```

A quantity constructor that yields the output of the first example for the input of the second one, would be logically consistent and resolve the following issues:

https://github.com/hgrecco/pint/issues/492
https://github.com/hgrecco/pint/issues/467
Also relates to: https://github.com/hgrecco/pint/issues/758

hgrecco:
There are a few things to resolve here.
1.- Quantity arrays are homogeneous in units and dimenstionality. What shall we do when this is not the case in the input array? One option would be to convert to the units parameter raising dimensionality error if not posible DimenstionalityError.
2.- What happens if units is not specified in the constructor? One option would be to use the first unit
3.- Do we accept all iterables? 
4.- How do we implemente it?

My suggestion would be: create a classmethod constructor named `from_iterable` (or similar) that handles everything. Once that we are satisfied we see if we need to add it to the default constructor.

andrewgsavage:
I hit the same error message `
ValueError: setting an array element with a sequence.
` in pint-pandas but worked around it. I think there may be places in pandas where that error may come up, but that's for the other discussion.... https://github.com/hgrecco/pint/issues/753


There's a very similiar function in pint-pandas, _from_sequence. https://github.com/hgrecco/pint-pandas/blob/master/pintpandas/pint_array.py#L415
I can see there's corner cases were when there's an empty sequence, nan as the first item/only item. 

Don't think much code can be shared since there's checks for PintType's units in there.

hgrecco:
@deniz195 Your proposal looks good to me. Please submit a PR. I do think that we might share some code with pint-pandas in the near future, we need to think what we do with #753

hgrecco:
Shouldn't this be closed?

deniz195:
Seems fine with me!

PR #762, opened 2019-02-05
Added custom constructors from_sequence and from_list

deniz195:
As previously discussed in issue #761. Tests for these constructors included in test_quantity.py.

(Similar to https://github.com/hgrecco/pint-pandas/blob/master/pintpandas/pint_array.py#L415)

### DIFF
--- a/pint/compat/__init__.py
+++ b/pint/compat/__init__.py
@@ -66,6 +66,14 @@ try:
 except ImportError:
     from itertools import izip_longest as zip_longest
 
+
+
+def _raise_invalid_magnitude(value):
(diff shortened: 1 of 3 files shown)


hgrecco:
Please add the decorator `@helpers.requires_numpy()` to the test so it is ignored when testing without numpy.

deniz195:
Hi, please note, that I had had wrongly commited some changes (935b8d2) that werent ment for this branch. They are undone in (81facee), which now really should be only a 1-line change vs. aee6245. 

(You can verify this here: https://github.com/deniz195/pint/compare/aee6245...81facee)

Please, excuse the confusion.

Issue #751, opened 2019-01-12
iter for single value....


jklymak:
Perhaps related to #55, but I can't see where that was resolved...

I have the input to a function that can either be a single value or an array.  I just want the first value of the array or the value.  Note that this has to work with or without pint.  

If I do 

```python
import pint

ureg = pint.UnitRegistry()
ureg.setup_matplotlib(True)

width = 0.25 * ureg.cm
print(np.iterable(width))
print(width[0])
```

I get:
 
```
True
Traceback (most recent call last):
  File "/Users/jklymak/pint/pint/quantity.py", line 1400, in __getitem__
    value = self._magnitude[key]
TypeError: 'float' object is not subscriptable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "testBar.py", line 13, in <module>
    print(width[0])
  File "/Users/jklymak/pint/pint/quantity.py", line 1404, in __getitem__
    "supports indexing".format(self._magnitude))
TypeError: Neither Quantity object nor its magnitude (0.25)supports indexing
```

Note that `iter(width)` also works fine...

So maybe I'm just missing something, but how do I check if `width` is a singleton?

jklymak:
Ok, but if I do

```
b = array(3)
iter(b)
<ipython-input-10-4e62f6ba0650> in <module>
----> 1 iter(b)

TypeError: iteration over a 0-d array
```

If I do the same with a pint object...

```
In [11]: a = 3 * ureg.cm

In [12]: iter(a)
Out[12]: <iterator at 0x116c39390>
```

I'd argue pint should do the same thing as numpy...

Agreed that `a.m` doesn't iterate, but we can't assume `a` has a method `m`.  We assume that `a` is something like a list, or numpy array, or a singleton, and that generally works for pint except for the singleton.

hgrecco:
True. According to the iter docs

> object must be a collection object which supports the
    iteration protocol (the __iter__() method), or it must support the sequence
    protocol (the __getitem__() method with integer arguments starting at 0).
    If it does not support either of those protocols, TypeError is raised.

Therefore, I think that could be accomplished by doing for `__getitem__` the same trick we do for `__iter__` here: https://github.com/hgrecco/pint/blob/master/pint/quantity.py#L136

crusaderky:
The fix seems straightforward - just need to change ``__iter__`` to return an iterator, as opposed to being an iterator function itself.

```python
class Wrapper:
    def __init__(self, inner):
[...]

Issue #760, opened 2019-02-03
matplotlib example from the docs failure (Neither Quantity object nor its magnitude supports indexing)


slayoo:
The example from: https://pint.readthedocs.io/en/0.9/plotting.html
gives:
```---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/usr/lib/python3/dist-packages/pint/quantity.py in __getitem__(self, key)
   1399         try:
-> 1400             value = self._magnitude[key]
   1401             return self.__class__(value, self._units)

TypeError: 'int' object is not subscriptable

During handling of the above exception, another exception occurred:

TypeError                                 Traceback (most recent call last)
<ipython-input-22-932eb629e0e4> in <module>()
      4 fig, ax = plt.subplots()
      5 ax.plot(x, y, 'tab:blue')
----> 6 ax.axhline(26400 * ureg.feet, color='tab:red')
      7 ax.axvline(120 * ureg.minutes, color='tab:green')

/usr/lib/python3/dist-packages/matplotlib/axes/_axes.py in axhline(self, y, xmin, xmax, **kwargs)
    803         # non-unitized bounds
    804         self._process_unit_info(ydata=y, kwargs=kwargs)
--> 805         yy = self.convert_yunits(y)
    806         scaley = (yy < ymin) or (yy > ymax)
    807 

/usr/lib/python3/dist-packages/matplotlib/artist.py in convert_yunits(self, y)
    193         if ax is None or ax.yaxis is None:
    194             return y
--> 195         return ax.yaxis.convert_units(y)
    196 
    197     @property

/usr/lib/python3/dist-packages/matplotlib/axis.py in convert_units(self, x)
   1519     def convert_units(self, x):
   1520         # If x is already a number, doesn't need converting
-> 1521         if munits.ConversionInterface.is_numlike(x):
   1522             return x
   1523 

/usr/lib/python3/dist-packages/matplotlib/units.py in is_numlike(x)
    124         """
    125         if iterable(x):
--> 126             for thisx in x:
    127                 return isinstance(thisx, Number)
    128         else:

/usr/lib/python3/dist-packages/pint/quantity.py in __getitem__(self, key)
   1402         except TypeError:
   1403             raise TypeError("Neither Quantity object nor its magnitude ({})"
-> 1404                             "supports indexing".format(self._magnitude))
   1405 
   1406     def __setitem__(self, key, value):

TypeError: Neither Quantity object nor its magnitude (26400)supports indexing
```

[...]

Issue #753, opened 2019-01-13
Split in two Quantity Classes for Scalar and Array objects


hgrecco:
When Pint was in its infancy there were several options ahead of us to support numpy arrays:
1. Just one type deriving from a np.array 
Good: Works with numpy. 
Bad: requires numpy for everything.
2. One type deriving from np.array and another which does not
Good: works with numpy without requiring it for non numpy operations
Bad: makes things more complicated.
3. Make an array of Scalar Quantity objects
Good: easy to implement
Bad: not very efficient
4. Create our own dtype derived form a type which is not numpy related
Good: works with numpy without requiring it for non numpy operations
Bad: a custom dtype difficult to get right and distribute. Fragile.
5. (Current solution). One type which stores in the magnitude a numeric object (whatever it is)
Good: works with numpy without requiring it for non numpy operations. Easy to implement
Bad: Some numpy operations are not supported

The current solutions has served us well (very well, I would say) and we were able to enforce correctness for operations with units without requiring users to import a particular math package and without monkey patching. So there is a strong reason to keep things like they are. But there are a few reasons to move forward

1. The landscape of numpy related packages is growing and failures might occurring within them in pint related operations are difficult to track and fix.
2. Pandas is becoming the defacto standar for many data related operations. We currently have a provisional support but we would like to make it easier to work with it.

To recap the reason why some numpy operations are not supported is because within numpy the fuction `asanyarray` is called multiple times and this copies the content of the array but looses the type (Quantity) because Quantity is not a `np.array` subclass. We even had a prototype of solution (2) which you can see a very outdated version in the `_npsubclass`

Please join the discussion.

andrewgsavage:
Thanks for opening this!

I think the reasoning behind Pandas' ExtensionTypes to to get around some difficulties with numpy. One example being int32 not being able to store nan, so we get this behaviour in the `_npsubclass `branch:

```
Q_(np.array([1, 2]),"m").magnitude.dtype
dtype('int32')

Q_(np.array([1, np.nan]),"m").magnitude.dtype
dtype('float64')
```
 
Which makes 1 and 2 undesirable.

From a pint-pandas perspective, an issue I found was having a single class for both Scalars and sequences. It would be good to have QuantityScalar and QuantitySequence to inherit from Quantity.

---

The array_function looks promising. I followed the example implementation and was able to get concatenate to return a quantity:
```python3
@implements(concatenate)
def _(arrays, axis=0, out=None):
    out_units = arrays[0].units
[...]

Issue #484, opened 2017-03-02
numpy arange unsupported


willprice:
Versions:

* `python`: v3.6.0
* `numpy`: v1.12.0
* `pint`: 0.8.dev0 (commit de8fdff)

```python
import numpy as np
import pint
ureg = pint.UnitRegistry()
x = 1 * ureg.meter

y = np.arange(0, x, 0.1)
```

Results in:
```
Traceback (most recent call last):
  File "pint_test.py", line 7, in <module>
    y = np.arange(0, x, 0.1)
  File "/home/will/src/pint/pint/quantity.py", line 431, in __float__
    raise DimensionalityError(self._units, 'dimensionless')
pint.errors.DimensionalityError: Cannot convert from 'meter' to 'dimensionless'
```

Since one can do `np.linspace(0, x, 10)` without problems I think this is a bug, although I may be misunderstanding and this may be an implication of the stated [numpy support](https://pint.readthedocs.io/en/0.6/numpy.html#support)

hgrecco:
The correct way would be:
```python
y = np.arange(0, 1 * ureg.meter, 0.1 * ureg.meter)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: setting an array element with a sequence.
```
However, this fails (with a different error). The reason is that `linspace` is implemented internally by numpy as a python function with additions and multiplications. `arange` is implemented in C and it does not "properly" dispatch to Pint Quantity methods.

masasin:
Something like `np.array([1 * ureg.meter, 0.1 * ureg.meter])` also fails with that error. Is there any way to mitigate that?

hgrecco:
The problem is that there is no way to tell numpy that we will handle it. The suggested way is 

```python
[1, 0.1] * ureg.meter
```

onealj:
And `np.arange(0, 1, 0.1)*ureg.meter`

On Jun 14, 2017 7:04 PM, "Hernan Grecco" <notifications@github.com> wrote:

The problem is that there is no way to tell numpy that we will handle it.
The suggested way is

[1, 0.1] * ureg.meter

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/hgrecco/pint/issues/484#issuecomment-308607690>, or mute
the thread

impact27:
I had the same problem and I had an idea: 

[...]

PR #764, opened 2019-02-12
Split Quantity into scalar and sequence classes

andrewgsavage:


andrewgsavage:
So that's with BaseQuantity having the methods/attributes shared across sequence and scalar, with a mixin (is that the right name?) having the sequence specific ones.

Quantity inherits from BaseQuantity and returns either a QuantityScalar or QuantitySequence when initiated, depending on the value.

QuantitySequence and QuantityScalar both inherit from Quantity, with QuantitySequence also inheriting the QuantitySequenceMixin

hgrecco:
Great start. I do have a lot of questions (I put them within the code).

pint/quantity.py
```diff
@@ -78,7 +78,7 @@ def wrapped(self, *args, **kwargs):
 
 
 @fix_str_conversions
-class _Quantity(PrettyIPython, SharedRegistryObject):
+class BaseQuantity(PrettyIPython, SharedRegistryObject):
```
hgrecco: Why is this classe "public" (i.e. no underscore)? Is this the class users should check against when using `isinstance`?

pint/quantity.py
```diff
@@ -1671,14 +1626,84 @@ def _ok_for_muldiv(self, no_offset_units=None):
...
+        # # Allow exception to propagate in case of non-iterable magnitude
+        it_mag = iter(self.magnitude)
+        return iter((self.__class__(mag, self._units) for mag in it_mag))
 
-    class Quantity(_Quantity):
-        pass
 
+def build_quantity_class(registry, force_ndarray=False):
+    
+    class Quantity(BaseQuantity):
```
hgrecco: Or this is class to be used when calling `isinstance`

hgrecco:
1. Shall we split format between scalar and array code [here](https://github.com/hgrecco/pint/blob/3ad6e39185e38c3617861bad93b8b616f8e6b4b8/pint/quantity.py#L169)
2. Shall we change all the inplace code to be really in place for the array as well (e.g. [here](https://github.com/hgrecco/pint/blob/3ad6e39185e38c3617861bad93b8b616f8e6b4b8/pint/quantity.py#L337)
3. Is all these needed for scalar [here](https://github.com/hgrecco/pint/blob/3ad6e39185e38c3617861bad93b8b616f8e6b4b8/pint/quantity.py#L1201)? I guess yes because you can still do numpy.cos(34). Is there another, cleaner, way?
4. Can we move `fill`, `clip`, etc to `QuantitySequence`?
5. Will this be reached at all? [code](https://github.com/hgrecco/pint/blob/3ad6e39185e38c3617861bad93b8b616f8e6b4b8/pint/quantity.py#L1635)
6. Is `__iter__` the only thing to test for? [code](https://github.com/hgrecco/pint/blob/3ad6e39185e38c3617861bad93b8b616f8e6b4b8/pint/quantity.py#L1692)

I guess that all the questions point to the same thing: How much can we simplify the code and making it more maintainable with this change?

pint/quantity.py
andrewgsavage: Hadn't thought of this when going through it! 

It should be used when the user does know the UnitRegistry. In the current version users should be checking with _Quantity (as done in pint-pandas) 
Using isinstance with a Quantity from one UnitRegistry and another UnitRegistry yields false:
'''
ureg=pint.UnitRegistry()

ureg2 = pint.UnitRegistry()

isinstance(my_array,ureg2.Quantity)
False
'''
Should this be public?

andrewgsavage:
So this was very much a 'get it behaving as you expect' before spending time on those more specific things - I was just moving lines around before spending time working out what paths change/need changing.

1, 2, If the behavior for ndarrays is what's expected from a sequence then yes. I guess this applies everywhere ndarrays are treated differently.

3. Haven't looked at this much. I think __array_function__ may serve us better for some functions, eg add, but won't be all that different when units aren't relevant

4.Yea, missed them - are there any others?

5. Haha no.

6. The main thing that slips through that is strings, which are handled previously.

There should be little difference with this change, from here onwards I can only see moving things around / removing checks like 2.

hgrecco:
And how about the connection to pandas ... will this make pint-pandas easier to implement?

andrewgsavage:
Came across this when looking for numpy functions
https://github.com/hgrecco/pint/issues/484
```python
y = np.arange(0, 1 * ureg.meter, 0.1 * ureg.meter)
ValueError: setting an array element with a sequence.
```
The main reason I'd like this was because of that error message - I hit it many times in pint-pandas and spent a while working around it, so it'll make pint-pandas easier, along with checking for scalar/sequence.
Although I note at present to do means comparing to QuantitySequenceMixin, which I guess needs renaming if it's intended for that use.. 

Back to that error message - I think it's because somewhere in numpy there's a check for iter or getitem, which makes it think it's a sequence and take a wrong path. Splitting into QuantityScalar fails that check so takes a different path:
```python
y = np.arange(0, 1 * ureg.meter, 0.1 * ureg.meter)
y
array([0, <Quantity(0.1, 'meter')>, <Quantity(0.2, 'meter')>,
       <Quantity(0.30000000000000004, 'meter')>, <Quantity(0.4, 'meter')>,
       <Quantity(0.5, 'meter')>, <Quantity(0.6, 'meter')>,
       <Quantity(0.7, 'meter')>, <Quantity(0.7999999999999999, 'meter')>,
       <Quantity(0.8999999999999999, 'meter')>], dtype=object)

```
I'm surprised that works with 0 and not 0*ureg.m

andrewgsavage:
So that's the tests passing again, except the uncertainties ones. What's the plan there - are they being split into another module?

I'm in two minds as to how to go forward with `__array_function`. Functions like arrange
numpy.arange([start, ]stop, [step, ]dtype=None)
look tricky to implement quantity conversions/checks accounting for the various optional args/ optional first arg.

Alternatively we could convert only the args that are quantities to consistent/root units. This would make it easy to implement many functions https://github.com/hgrecco/pint/pull/764/files#diff-d9924213798d0fc092b8cff13928d747R115 but prevents function specific checks. 


The other thing I've noticed is the double underscored attributes, eg __used have become _BaseQuantity__used with the inheritance. Not sure if this is an issue.

hgrecco:
This is looking nice. My suggestionm is that we invite more reviewers. Not only to review the code but the concept itself. It think is such a big change the requires more eyes.  I am inviting some, please feel free to invite

@jondoesntgit @znicholls @cpascual

andrewgsavage:
something odd happening with compress, the test gives an error:
```python3
a = np.ones((3, 2))
q = self.Q_(a, "km")
self.assertEqual(q.u, np.compress(q==q[0,0], q).u)

<ipython-input-30-8122631a52b3> in <module>
      1 a = np.ones((3, 2))
      2 q = self.Q_(a, "km")
----> 3 self.assertEqual(q.u, np.compress(q==q[0,0], q).u)

c:\users\a\repos\numpy\numpy\core\overrides.py in public_api(*args, **kwargs)
    149             relevant_args = dispatcher(*args, **kwargs)
    150             return implement_array_function(
--> 151                 implementation, public_api, relevant_args, args, kwargs)
    152 
    153         if module is not None:

TypeError: no implementation found for 'numpy.compress' on types that implement __array_function__: [<class 'numpy.ndarray'>, <class 'pint.quantity.build_quantity_class.<locals>.QuantitySequence'>]

```

However this works fine:
```python3
self.assertEqual(q.u, np.compress([True]*6, q).u)
```

andrewgsavage:
Has anyone had a chance to look at this?

jthielen:
Just a downstream user here, so I unfortunately don't have the background yet to comment much on all the implementation details, but I'm excited for the possibilities that having an `__array_function__` implementation opens up.

I did a quick look through this, and overall the concept of splitting into scalar and sequence classes seems good as far as I can tell. However, there are a few issues that came up as I was testing it out:

- Many numpy functions that previously worked before the implementation of `__array_function__` by seemingly casting to a ndarray now raise a TypeError for not having an implementation. Should Pint explicitly handle these, or should there be some kind of fallback? While there are probably many others, the list I came up with from MetPy is
  - `amax`
  - `amin`
  - `append`
  - `argsort`
  - `clip`
  - `meshgrid`
  - `nanmax`
  - `nanmin`
  - `ones_like`
  - `shape`
  - `zeros_like`
- Comparisons between `QuantityScalar` and `QuantitySequence` are not working (perhaps https://github.com/hgrecco/pint/blob/4dcbe78766269f985d1898601d8f5384a1511f2c/pint/quantity.py#L1369 needs to change to
    ```python
    if not isinstance(other, Quantity):
    ```
    ?)
- Non-multiplicative units don't seem to be working properly with `numpy.diff` and `numpy.ediff1d`...it's giving the base unit, not the delta unit
- There is quite a bit of extra output to stdout...are the print statements there for debugging? See https://github.com/hgrecco/pint/blob/4dcbe78766269f985d1898601d8f5384a1511f2c/pint/quantity.py#L154 https://github.com/hgrecco/pint/blob/4dcbe78766269f985d1898601d8f5384a1511f2c/pint/quantity.py#L159 https://github.com/hgrecco/pint/blob/4dcbe78766269f985d1898601d8f5384a1511f2c/pint/quantity.py#L185 https://github.com/hgrecco/pint/blob/4dcbe78766269f985d1898601d8f5384a1511f2c/pint/quantity.py#L196

One other thing that isn't directly from the changes in this PR, but is a related and perhaps more-general issue: I noticed when testing this with https://github.com/pydata/xarray/pull/3117 that multiplying `xarray.DataArray` * `Quantity` works as expected, but `Quantity` * `xarray.DataArray` does not (it tries to wrap the DataArray with the Quantity, rather than operating on the Quantity inside the DataArray). Should there be some sort of blacklist or heuristic for objects that, when a Quantity operation encounters them, should return `NotImplemented` (see https://github.com/pydata/xarray/issues/525#issuecomment-483802699)?

andrewgsavage:
1. Yea, there's many numpy functions that aren't implemented yet. The ones that are implemented are here. https://github.com/hgrecco/pint/pull/764/files#diff-d9924213798d0fc092b8cff13928d747R198 
They're the ones that were tested in pint previously, or from the first few pages of numpy functions. 

I'd prefer raising TypeError over a returning something with the potentially units incorrect. It's not a big deal to implement functions - The process for implementing a function is to add to the relevant list depending on how the units should behave, and adding a test.
2. Yea probably needs changing to BaseQuantity

3. I expect diff and ediff1d should be in 'div' list, like gradient. 

4. Yes... That's me trying to work out why this line https://github.com/hgrecco/pint/pull/764/files#diff-d9924213798d0fc092b8cff13928d747R160 and the suffixed "_" variables work, but removing that line + "_" suffixes makes it fail...
5. Yes, try adding DataArray to https://github.com/hgrecco/pint/blob/4dcbe78766269f985d1898601d8f5384a1511f2c/pint/quantity.py#L70

jthielen:
@andrewgsavage Thank you for the replies! I would be glad to help implement and test more of the numpy functions and verify xarray works with the `check_implemented` decorator once this gets merged.

With regards to the TypeError, that sounds good about raising a TypeError instead of returning with potentially incorrect units; I was just mostly concerned about the significant API break. Do you think that a strong warning about this API change being added to the changelog is the best approach, or should there be some deprecation period where unknown functions get handed off to the magnitude of the quantity with a `UnitStrippedWarning` raised? Perhaps like [`__getattr__`](https://github.com/hgrecco/pint/blob/4dcbe78766269f985d1898601d8f5384a1511f2c/pint/quantity.py#L1524-L1547) but with a deprecation warning?

hgrecco:
Downstream feedback is what we need! I am ready to merge this as soon we are satisfied that we are at least as good as we were before and we can fix upcoming problems. More testing is what we need

jthielen:
Also, this appears to fix #751 and #760, which is a nice bonus.

andrewgsavage:
You can add implementations to https://github.com/andrewgsavage/pint/tree/split_quantity and I'll merge.

I'm not sure what to do about depreciation warnings. Many functions will change behavior too, returning units when they did not before. Should these have warnings too?

Would it be better to release a pint version with the unitstrippedwarnings changed to a futurewarning or similiar to alert users, or is a note in the changelog sufficient?

hgrecco:
I think we should alert users but in the alert message describe a very simple way to disable it.

andrewgsavage:
There's a warning - I get the following when I import pint:

```python3
/home/andrew/repos/pint/pint/compat/__init__.py:111: BehaviorChangeWarning: The way pint handles numpy operations has changed. Unimplemented numpy operations will now fail instead of making assumptions about units. Some functions, eg concat, will now return Quanties with units, where they returned ndarrays previously. See https://github.com/hgrecco/pint/pull/764 . To hide this warning use the following code to import pint:

import warnings
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    import pint
---    

  warnings.warn(_msg, BehaviorChangeWarning)

```

jthielen:
A question that came up while I was working on implementing numpy functions and writing tests (I'm sorry for not catching it earlier):

What should be the behavior for mixed argument types (like base ndarrays, or possibly lower-level `__array_function__` implementers like sparse, dask, and cupy)? Right now in this PR, pint returns `NotImplemented` whenever any argument is not a pint `BaseQuantity` (see below).

https://github.com/hgrecco/pint/blob/5a528adde3da518b7ca234ed4caa3015a012233e/pint/quantity.py#L258-L259

However, I could imagine something like the following being desired functionality:

```python
k = np.array([0, 0, 1])
np.dot(k, [1, 2, 3] * ureg.m)
```

Plus, this is probably a question to make sure is cleared up in general. 

My initial idea is to have a whitelist of array types that are known to "fit inside" a `BaseQuantity`, and if any of these are encountered by `BaseQuantity`'s `__array_function__`, they should be wrapped in a dimensionless `Quantity` before being passed to pint's implementation. Any thoughts on that, or alternatives?

andrewgsavage:
Ya that makes sense. Right now its using ```hasattr(value, "__iter__")```
to check for sequences when deciding whether to create a qtysequence or
qtyscalar. Not sure what's the better approach.
Maybe it's worth having an is_list_like function, like pandas does, for
consistency.

jthielen:
I added more numpy function implementations and tests to https://github.com/andrewgsavage/pint/pull/6, along with a few additional changes that can be reviewed. Please do let me know if it is what you were looking for!

There are some additional functions I had hoped to implement, but I did not do so yet:

- `prod`, `product`, and `nanprod`, since determining how many multiplications occur in order to derive the output unit is non-trivial (unless there is some reduction helper function I'm missing)
- `where` and `histogram`, since they seem to make the most sense when mixed with non-unit ndarrays (so waiting on https://github.com/hgrecco/pint/pull/764#issuecomment-523272038 / https://github.com/hgrecco/pint/issues/845)

andrewgsavage:
Before merging jthielen's additions, do we want to make pint require (or not work with) numpy versions without ```__array_function__```? 

I think it'll be confusing supporting np in two different ways, and part of the current implementation + tests are removed with this PR

Other than that it looks good! Are there any other functions that would be necessary for MetPy?

jthielen:
I ran my changes past MetPy's test suite, and I found a few functions I missed. So, I pushed one more commit with those additional implementations (also, the BaseQuantity comparison change mentioned earlier).

I would lean towards maintaining support for numpy without `__array_function__` active for a couple reasons:
- Just having a sufficient version of numpy (numpy>=1.16) does not mean `__array_function__` will work, since it is dependent on the user environment variable `NUMPY_EXPERIMENTAL_ARRAY_FUNCTION` (see https://www.numpy.org/neps/nep-0018-array-function-protocol.html#implementation), so supporting numpy>=1.16 with `NUMPY_EXPERIMENTAL_ARRAY_FUNCTION = 0` should be the similar to supporting numpy<1.16
- Most other libraries I've seen implement `__array_function__` compatibility have maintained apparent support for numpy<1.16 (if I'm interpreting things correctly: explicitly in the case of xarray, dask, and cupy; implicitly for sparse) and `NUMPY_EXPERIMENTAL_ARRAY_FUNCTION = 0`
- I'm not sure if this is the case with other downstream libraries or not, but MetPy tries to support minor versions of dependencies released within the last two years (so, numpy>=1.13 right now). If pint>=0.10 doesn't support numpy<1.16, MetPy would either have to change its release support policy, request back-ported pint 0.9.x releases, or extensively monkey-patch pint in order to get needed bug fixes (such as #751 / #760 right now)

hgrecco:
This is looking really good. Maybe we can try to define a roadmap and put a date for a release.

dopplershift:
If this starts to actually solve the problem of units + xarray and other things, I'm willing to adjust MetPy's dependencies--especially given that this could really help us settle some things for oru 1.0 release.

I really appreciate the hard work that's been/being put in to bridge this gap.

jthielen:
@hgrecco Sounds great about a roadmap!

With that in mind, @keewis, would it be better for your work in https://github.com/pydata/xarray/pull/3238 that this gets merged and released sooner rather than later, or would you rather get the tests written to make sure that pint has what xarray needs before this gets released?

keewis:
I think you can take your time: it is easy enough for me to switch to another environment that has pint with the changes of this PR installed instead of the released version.

jthielen:
Also, if the consensus is to only support `np` functions when `__array_function__` is available from this point forward, just let me know and I can change my PR to this PR (https://github.com/andrewgsavage/pint/pull/6) accordingly.

jthielen:
Based on tests in https://github.com/pydata/xarray/pull/3238, it looks like there is inconsistent behavior in pint between numpy functions that go through `__array_function__` and those that go through `__array_ufunc__`, especially with regards to non-explicitly-handled functions. In that case, the `__array_function__` implementation returns `NotImplemented`, whereas the `__array_ufunc__` implementation returns a result but strips units without a warning.

Should `__array_ufunc__` be brought into consistency with how `__array_function__` is implemented here by only operating on explicitly-handled ufuncs?

andrewgsavage:
Are there any libraries that see such a change between supporting/not supporting `__array_function__`, where users will have to change their code?

Assuming people are happy to only support numpy >=1.16 in the future, I'd suggest the following roadmap:

0.10 - Split Quantity into scalar and sequence classes (would split this PR into one splitting classes and another implementing  `__array_function__`. )
0.11 - Support `__array_function__`, extas_require numpy>=1.16 
We could backport changes to 0.10.x if necessary to maintain support for numpy <1.16

andrewgsavage:
> Based on tests in [pydata/xarray#3238](https://github.com/pydata/xarray/pull/3238), it looks like there is inconsistent behavior in pint between numpy functions that go through `__array_function__` and those that go through `__array_ufunc__`, especially with regards to non-explicitly-handled functions. In that case, the `__array_function__` implementation returns `NotImplemented`, whereas the `__array_ufunc__` implementation returns a result but strips units without a warning.
> 
> Should `__array_ufunc__` be brought into consistency with how `__array_function__` is implemented here by only operating on explicitly-handled ufuncs?

Maybe. This will break many functions that work but aren't explicitly handled. Are people likely to submit PRs to handle these?

jthielen:
> Are there any libraries that see such a change between supporting/not supporting `__array_function__`, where users will have to change their code?

I think it depends on how much you consider "such a change." With any numpy function that internally calls `np.asarray()` or `np.asanyarray()`, an overriding `__array_function__` implementation that preserves non-subclass type will be a breaking change in any library supporting `__array_function__`. Where pint may be different from other non-ndarray-subclass implementers I've seen (dask, sparse, cupy, etc.) is that casting to a plain ndarray and reconstructing afterwards is not an expensive operation, so downstream users are more likely to try working around numpy functions that cast to ndarray by assuming the casting occurs and correcting afterwards, rather than implementing a custom function (this is what MetPy assumed and is now trying to fix, xref https://github.com/Unidata/MetPy/pull/1144).

> Assuming people are happy to only support numpy >=1.16 in the future, I'd suggest the following roadmap:
>
> 0.10 - Split Quantity into scalar and sequence classes (would split this PR into one splitting classes and another implementing `__array_function__`. )
> 0.11 - Support `__array_function__`, extas_require numpy>=1.16
> We could backport changes to 0.10.x if necessary to maintain support for numpy <1.16

I like the idea of splitting this PR into one splitting classes and another implementing `__array_function__`. Right now all the complications relating to `__array_function__` do seem to be derailing and holding back the original titled focus of the PR :grin:. Let me know if you'd like me to submit the new `__array_function__` PR based on my changes in https://github.com/andrewgsavage/pint/pull/6 (which I can close) after you remove the `__array_function__` implementation here.

My one concern with putting off `__array_function__` until v0.11 is the delay in the `__array_function__` implementation making a release. Based on engagement with https://github.com/pydata/xarray/issues/525, https://github.com/hgrecco/pint/issues/849, and related issues, it seems people are anxious to see this in action (I know I'm definitely one of them). For example, it sounds like this would really help MetPy for its upcoming v1.0 release, which is scheduled to be out by the end of 2019. So, if v0.11 is going to be the target release for `__array_function__` support, would it be okay to commit to it being released on a relatively short timeline (within the next couple months)? I'd be glad to do what I can to help make it happen.

Otherwise, this seems like a good roadmap to me!

> Maybe. This will break many functions that work but aren't explicitly handled. Are people likely to submit PRs to handle these?

I believe this will break non-explicitly-handled ufuncs just like the `__array_function__` implementation breaks non-explicitly-handled functions. I think it's crucial that the behavior be consistent for the sake of downstream libraries/users (xref https://github.com/pydata/xarray/pull/3238). What that consistent behavior is (fail if not implemented or strip units and raise `UnitStrippedWarning`), I could be convinced either way. I don't think numpy ufunc vs. function will make a difference in people submitting PRs to handle them.

keewis:
what is the status on this? Now that both pydata/xarray#3238 and pydata/xarray#3447 are either merged or close to being merged, the support for `__array_function__` in pint is one of the major blocks to getting pint to work with xarray.

I would also be willing to help with getting this moving again or starting a new PR that only works on `__array_function__` support.

hgrecco:
If all the interested parts are sattisfied with the currenbt solution, I think we should merge it. (after fixing the pending conflict)

jthielen:
There are a couple reasons I think this shouldn't be merged yet:
- there is some strong opposition to the Quantity class split in its current state in #753 / #875 by @shoyer and @crusaderky
- there are some needed changes in https://github.com/andrewgsavage/pint/pull/6 that haven't been included here yet

Instead, I'd be in favor of taking the `__array_function__` implementation here and in https://github.com/andrewgsavage/pint/pull/6 and putting in a new PR separate from the Quantity class split. However, to avoid merge conflicts, I've been waiting on #880 to be merged into master before finalizing my implementation of it and submitting that PR.

keewis:
if I remember correctly there was some discussion about whether the split into scalars / sequences was a good idea, so I think we should probably remove all code relating to that first (see #753 and #875).

shoyer:
To be clear, I am happy to advise based off my cursory understanding of
pint, but don't take my advice too seriously -- I'm not the one actually
implementing this! You can certainly make __array_function__ overrides work
with either a single class or separate classes for scalars and arrays.
Ultimately this is a decision for pint's maintainers and whoever is
implementing this.

hgrecco:
I just made my comments on #880 Short version, I am happy to merge it after a minor change.

jthielen:
In case it is useful to anyone at this point, a branch I have with a rough implementation of `__array_function__` based on #880 and what was done here is up at https://github.com/jthielen/pint/tree/nep18-1.

Current To-Do List:

- Definitely address now
  - [x] Merge/modify previous changes
  - [ ] Address `np.power`
  - [ ] Address `np.searchsorted` (see https://github.com/pydata/xarray/pull/3238#issuecomment-532325585)
  - [ ] Verify ndarray-returning functions don't mistakenly have units attached (see https://github.com/andrewgsavage/pint/pull/6#issuecomment-532385098)
- Maybe address now, maybe follow-up PR
  - [ ] Mixed argument types (https://github.com/hgrecco/pint/pull/764#issuecomment-523272038)
    - `np.compress` failing test
    - `np.where`, `np.histogram` (https://github.com/hgrecco/pint/pull/764#issuecomment-523749008)
  - [ ] Ensure consistent behavior between `__array_function__` and `__array_ufunc__` (https://github.com/hgrecco/pint/pull/764#issuecomment-527232914)
  - [ ] Wrapping non-ndarray NEP18-compatible arrays (#845 / #878)

keewis:
given that #880 has been merged, how do we want to continue with this? Would it be better to wait on a decision on #753 and #875 or should we go ahead and implement support for `__array_function__` and worry about compatibility with `QuantityScalar` / `QuantitySequence` afterwards should that become necessary? If implementing directly is the consent: @jthielen, do you have the time to convert your branch to a PR?

jthielen:
@keewis My apologies for the delays this past week! I'd like to submit the branch I have as a PR, I just haven't had the time to complete the remaining checklist items yet. I should though have the time in the next few days (just probably not today).

keewis:
I don't think you have to have completed those items when you create the PR (but good to hear you are working on this)

jthielen:
A status update: unfortunately, the current implementation of this PR (with `implement_func`) has not been amendable to handling mixed argument types, so I am still working on re-implementing alternatives that can be more signature-aware. So, it is probably a good thing I didn't submit the PR with what I had, since much of it is still in flux. I've also been finding small errors in function behavior that I've been slowly correcting along the way as well.

Also, a couple side questions:

- While I've been making do without it by writing many unique `@implements` implementations, it would be nice to try using `inspect.signature` to reduce repeated code that only differs by the signature. Would this be enough justification to drop Python 2.7 now, given that it still hasn't been dropped in master yet?
- Some of the NumPy functions of interest have signatures that vary across versions, so before I get too deep into ensuring compatibility for those, what should pint target as its minimum supported NumPy version? I would think following [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) would be best (so, v1.14 until January 7, 2020), but I wanted to ask before jumping to any conclusions.

keewis:
@jthielen, I tried to use your branch to run the tests from xarray and ran into some issues (among others something about having `"dtype"` as dtype or unexpected kwargs). What do you think about opening a PR using the current state so we can discuss these and other issues and maybe also provide feedback on the design? I'd like to help with investigating and squashing bugs, but right now I'm not sure where to suggest changes (should I create PRs against your branch?).

Considering official python 2.7 support expires in [a little more than a month](https://pythonclock.org), I'd say let's drop it, at least on master (but that's only my personal opinion). @hgrecco, what do you think?

jthielen:
@keewis I'm sorry for the continued delays! I've opened #905 as a draft PR with my current status and some notes about what I see as remaining high-level tasks for it. I really need to be making progress on it (if there is still to be hope of having this in a Pint release by the end of the year), but I'm not sure when I'll get the time needed with all my other end-of-semester tasks...hopefully soon though.

As far as suggesting changes, in absence of a better system, I think PRs to my branch would be best for large changes, whereas suggested changes in a review to #905 would be best for small changes?

hgrecco:
See #906

PR #768, opened 2019-02-24
Default group should only have orphan units

### LINKED ISSUE #766: All units included in international group and mks system
Root group of units are added by default to the default group then to the default system of units.

Like this : 
* root group (all units) -> default group (international) -> default system (mks)

This behavior makes useless function like `<Some_Quantity>.get_compatible_units("mks")`, as it will get all compatible units. Filtering by group or system is useless there.

NB: Units in root group are available no matter what in the registry.

I would expect international group to contain only SI or derived SI units defined by CGPM.

Code involved : 
```python
    def _after_init(self):
        super(SystemRegistry, self)._after_init()

        #: Copy units in root group to the default group
        if 'group' in self._defaults:
            grp = self.get_group(self._defaults['group'], True)
            grp.add_units(*self.get_group('root', False).non_inherited_unit_names)

        #: System name to be used by default.
        self._default_system = self._default_system or self._defaults.get('system', None)
```

### Opportunities : 
Moving forward with check like : 

* Q_(1,"lb").is_SI()
* Q_(1,"mph").in_system("mks")

### Proposed solution : 

Make an international group outside of root unit group which contains all SI or SI derived.
Remove include of root units to default group and system

@hgrecco @jondoesntgit  Any thoughts before I submit a PR ?

jules-ch:
As discussed here #766

Previously, all units from root group was added to default group (international group in default definition file)

This PR change that to : 
*  Only units that are not defined in any other group are added to the default group.

This will prevent unintended group nesting to default group and diferentiate default group from root group.

Closes #766

### DIFF
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -1293,12 +1293,27 @@ class SystemRegistry(BaseRegistry):
         self._default_system = system
 
     def _after_init(self):
+        """After init function
+
+        Create default group.
+        Add all orphan units to it.
+        Set default system.
+        """
         super(SystemRegistry, self)._after_init()
 
-        #: Copy units in root group to the default group
+        #: Copy units not defined in any group to the default group
         if 'group' in self._defaults:
             grp = self.get_group(self._defaults['group'], True)
-            grp.add_units(*self.get_group('root', False).non_inherited_unit_names)
+            group_units = frozenset(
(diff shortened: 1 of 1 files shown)


hgrecco:
Looks good to me. 

bors r+

PR #791, opened 2019-04-16
Fix instantiation of errors.DefinitionSyntaxError in context.py

cpitclaudel:
The original message led to the following being displayed:

    pint.errors.DefinitionSyntaxError: While opening units.txt, in line 3:
    Could not parse Context definition defaults: '%s'

instead of the intended:
    
    pint.errors.DefinitionSyntaxError: While opening units.txt, in line 3: 
    Could not parse Context definition defaults: '(factor=xyz)'


hgrecco:
thanks!

PR #796, opened 2019-04-20
Added t_force, short_ton_force and long_ton_force

janheindejong:
I added the definitions of the different ton force metrics. Added: 

- ``force_metric_ton``, ``metric_ton_force``, ``t_force`` and ``force_t`` in the root system 
- ``force_long_ton`` and ``long_ton_force`` in the Avoir-Dupois group 
- ``force_short_ton`` and ``short_ton_force`` in the Avoir-Dupois group 
- ``UK_ton_force`` and ``US_ton_force`` in AvoirdupoisUK and AvoirdupoisUS

Furthermore, I moved ``force_ton`` and ``ton_force`` to the AvoirdupoisUS. 

This definition is backward compatible (i.e. ``force_ton`` and ``ton_force`` are still the same). 

I'm not entirely sure if I'm working with the systems and groups correctly. I haven't been able to create different implementations of ``ton_force``, depending on what system is used. In my case, the one defined last in the definitions file is used, irrespective of what system is used. 

Please let me know what you think!

### DIFF
--- a/pint/default_en.txt
+++ b/pint/default_en.txt
@@ -132,7 +132,7 @@ force_kilogram = g_0 * kilogram = kgf = kilogram_force = pond
 force_gram = g_0 * gram = gf = gram_force
 force_ounce = g_0 * ounce = ozf = ounce_force
 force_pound = g_0 * lb = lbf = pound_force
-force_ton = 2000 * force_pound = ton_force
+force_metric_ton = g_0 * t = metric_ton_force = force_t = t_force
 poundal = lb * feet / second ** 2 = pdl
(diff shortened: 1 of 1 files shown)


hgrecco:
Looks good. Just two things:
1. I think it will be good to enumerate in the commit message the backwards incompatible changes, if any.
2. It would be good to squash all the commits into a single one to keep the history clean

janheindejong:
Voila!

janheindejong:
Great! Any idea when you'll release a new version, so we can start using this in our organisation?

PR #801, opened 2019-04-26
Preserve caches across context changes (fixes #795)

### LINKED ISSUE #795: Using a context results in a 1500x slowdown for .to()
I ran into performance issues with contexts.  Here is a minimal repro:

```
$ python3 -m timeit -s "import pint; ureg = pint.UnitRegistry(); src = ureg('1 m'); dst = ureg.Unit('km')" "src.to(dst)"
10000 loops, best of 3: 49.3 usec per loop

$ python3 -m timeit -s "import pint; ureg = pint.UnitRegistry(); src = ureg('1 m'); dst = ureg.Unit('km')" "src.to(dst, 'chemistry')"
10 loops, best of 3: 66.7 msec per loop

$ python3 -m timeit -s "import pint; ureg = pint.UnitRegistry(); src = ureg('1 m'); dst = ureg.Unit('km')" "src.to(dst, 'spectroscopy')"
10 loops, best of 3: 67.1 msec per loop
```

In other words, `pint` can do 20'000 conversions per second on my machine with an empty context, but only 15 conversions per second when I give it a context.

cpitclaudel:
Things that need to be double-checked:

* With this changed, switching contexts clears the whole cache (until now, only dimensional_equivalents was cleared)
* I'm assuming that the list of context names is enough to characterize a context chain; if not (that is, if context chains or contexts can be mutated without name changes), then this approach is unsound.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,6 +10,7 @@ Other contributors, listed alphabetically, are:
 * Brend Wanders <b.wanders@utwente.nl>
 * choloepus
 * coutinho <coutinho@esrf.fr>
+* Clément Pit-Claudel <clement.pitclaudel@live.com>
 * Daniel Sokolowski <daniel.sokolowski@danols.com>
 * Dave Brooks <dave@bcs.co.nz>
 * David Linke
--- a/pint/context.py
+++ b/pint/context.py
@@ -192,6 +192,8 @@ class Context(object):
         _key = self.__keytransform__(src, dst)
         return self.funcs[_key](registry, value, **self.defaults)
 
+    def __hash__(self):
+        return hash(self.name)
 
 class ContextChain(ChainMap):
     """A specialized ChainMap for contexts that simplifies finding rules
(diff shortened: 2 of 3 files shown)


hgrecco:
Testing in travis is failing. As far as I can see this is due to some `_dimensional_equivalents` still refering to the one outside the new cache object. It might happent to others parts of the new cache.
This should be rewriten for the new names or maybe a read-only `_dimensional_equivalents` property returning the active cache could be created.

cpitclaudel:
Sorry, this fell off my radar. I just pushed the relevant fix… but it's not enough. It seems that the part were the code didn't reset the full cache (but just parts of it) was intentional, since resetting the full cache breaks tests.  I'm referring to this discussion in https://github.com/hgrecco/pint/issues/795.

>> I could look into it.  Quick question: why are  self._root_units_cache and self._dimensionality_cache not cleared at the  beginning of _build_cache?
>
>> It's a bug. It has not surfaced because it is rare that somebody changes the roots units and/or dimensionality without recreating the registry.

Any chance you could have a look? These are not parts of the code that I understand well, unfortunately.

hgrecco:
I will

cpitclaudel:
Thanks a lot :)

crusaderky:
Hi, any update on this? Thanks in advance

hgrecco:
now that travis is fixed, I will merge it

crusaderky:
Hi,
Any progress on this? I fear it's going to heavily collide with the work I'm starting for  https://github.com/hgrecco/pint/issues/853

crusaderky:
Superseded by #909, this one should be closed

PR #805, opened 2019-05-10
Fix definition of dimensionless units

Jellby:
There is something strange going on with dimensionless units/numbers:

```
>>> ureg.get_base_units('pi')
(3.141592653589793, <Unit('dimensionless')>)
>>> ureg.get_base_units('fine_structure_constant')
(1.0, <Unit('fine_structure_constant')>)
```

It seems dimensionless units defined in a file are interpreted as base units, while `pi` is hard-coded and interpreted correctly as a number. I think this is because for an empty list both `all()` and `not any()` are true. The fix is just changing the order of the tests, such that when a unit has no dimensions the `not any()` test passes first and the unit is marked as not-base. After the fix:

```
>>> ureg.get_base_units('pi')
(3.141592653589793, <Unit('dimensionless')>)
>>> ureg.get_base_units('fine_structure_constant')
(0.0072973525698, <Unit('dimensionless')>)
```

In addition, the hard-coded definition of `pi` is better done *before* loading the file definitions, I believe, so `pi` can be used in defining base units for `@system`, and also to allow redefining `pi` in a file, if desired. This also required removing the special treatment for `pi` in `_eval_token()`, which I don't think is need anyway.

Finally, I applied some corrections to the default definitions: Added `ln10` = ln(10), just to test the above fix with a number not subject to updates. Fixed the `deka-` prefix. Made the `acre_foot` consisent, I don't think it should use the survey foot for area times the international foot. Fixed the dry barrell, which I see everywhere defined as 7056 in^3, not 7065. Fixed spelling of "hundredweight" (if it was intentional, I still find it weird that the long and short ones are misspelled differently).

### DIFF
--- a/pint/constants_en.txt
+++ b/pint/constants_en.txt
@@ -4,6 +4,8 @@
 # Source: http://physics.nist.gov/cuu/Constants/Table/allascii.txt
 # :copyright: 2013 by Pint Authors, see AUTHORS for more details.
 
+ln10 = 2.302585092994046
+
 speed_of_light = 299792458 * meter / second = c
 standard_gravity = 9.806650 * meter / second ** 2 = g_0 = g_n = gravity
 vacuum_permeability = 4 * pi * 1e-7 * newton / ampere ** 2 = mu_0 = magnetic_constant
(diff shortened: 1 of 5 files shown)


hgrecco:
Looks really good to me. Thanks
bors r+

PR #809, opened 2019-05-16
Allow units with aliases but no symbol

### LINKED ISSUE #808: Define unit without symbol
Is it currently possible to define a unit without symbol (but with aliases) in a text file? Currently we have behavior such as:

```
>>> print('{:}'.format(ureg['1 millenium']))
1 millenium
>>> print('{:~}'.format(ureg['1 millenium']))
1 millenia
```

Surely no one intended "millenia" to be the symbol for millennium (the multiple misspelling is another matter). The unit is defined as:

```
millenium = 1000 * year = millenia = milenia = milenium
```

It is possible to set an empty symbol with

```
millenium = 1000 * year = = millenia = milenia = milenium
```

but that still gives a unit with a symbol (that happens to be an empty string). Is there another way? Or should this be the preferred way, and then the code should change an empty symbol to `None`?

Jellby:
This should fix #808. As suggested, I used `_` as a placeholder for no symbol, but it could be changed.

I updated the default definitions accordingly, and noticed that the `delta_` prefix was not being added to the aliases, only to the canonical name of an non-multiplicative unit, so I "fixed" this too.


hgrecco:
Great! Thanks

PR #811, opened 2019-05-23
Update to new SI definition and CODATA 2018

Jellby:
The SI has been redefined, by giving exact values to the Planck, Boltzmann and Avogadro constants, and the elementary charge. There are also new recommended values for other constants (CODATA 2018).

I've rewritten the `constants-en.txt` and `default-en.txt` according to these changes and also, I believe, for a better structure. The most significant changes:

* All units and constants defined from a few "experimental" values.
* For as much as I could find them, all factors are exact, except when noted. Floating point representation can introduce inaccuracies, though.
* US and imperial units may have changed their "reference" (to what I considered is most reasonable or authoritative), in practice values stay the same within floating point errors.
* Added some units and quantities.
* Fixed some units like Nm or au and parsec.
* Set the default Btu to the "ISO" Btu, and the default year to the Julian year (as recommended by the IAU)
* Defined units of the Gaussian system consistently, which makes them have different dimensionality from the corresponding SI units. Added a Gaussian context to allow conversions (with some caveats).
* Removed some units for which I could find no definition or reasonable use: chemical_faraday, physical_faraday, sidereal_hour, sidereal_minute, sidereal_second, work_year, work_month (they can of course be added back, hopefully with a better formal definition).
* Added atomic and Planck systems.
* Rearranged the order of units to "importance" (kind of) or field, rather than alphabetic.

I'm happy to undo some of the changes if they are considered undesirable. Note if/when #809 is accepted, I'd like to apply some minor fixes here.

Had to modify a couple of tests:

* For some reason (floating point errors, I guess), 4 mM -> mole/liter is no longer equal to 4e-3 M, only "almost equal".
* Updated the value of the Boltzmann constant in a test, a difference of 1.45e-5 % seems to be too much.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,6 +19,7 @@ Other contributors, listed alphabetically, are:
 * Felix Hummel <felix@felixhummel.de>
 * Francisco Couzo <franciscouzo@gmail.com>
 * Giel van Schijndel <me@mortis.eu>
+* Ignacio Fdez. Galván <jellby@yahoo.com>
 * James Rowe <jnrowe@gmail.com>
 * Jim Turner <jturner314@gmail.com>
 * Joel B. Mohler <joel@kiwistrawberry.us>
--- a/pint/constants_en.txt
+++ b/pint/constants_en.txt
@@ -57,7 +57,7 @@ atomic_mass_constant = 1.66053906660e-27 kg = m_u                               
 electron_mass = 9.1093837015e-31 kg = m_e = atomic_unit_of_mass = a_u_mass                 # (28)
 proton_mass = 1.67262192369e-27 kg = m_p                                                   # (51)
 neutron_mass = 1.67492749804e-27 kg = m_n                                                  # (95)
-lattice_spacing_of_Si = 1.920155716 m = d_220                                              # (32)
+lattice_spacing_of_Si = 1.920155716e-10 m = d_220                                          # (32)
 K_alpha_Cu_d_220 = 0.80232719                                                              # (22)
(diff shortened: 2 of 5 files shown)


hgrecco:
I have merged #809 This commit still has problems.

Jellby:
You mean the checks? I'm trying to debug it. It fails only with python 2, with:

    molar = mole / (1e-3 * meter ** 3) = M

it passes, but with;

    molar = mole / liter = M

it fails, saying:

```
======================================================================
ERROR: test_issue29 (test_issues.TestIssues)
----------------------------------------------------------------------
  File "/home/ignacio/test/pint/pint/testsuite/test_issues.py", line 46, in test_issue29
    self.assertAlmostEqual(t.to('mole / liter'), 4e-3 * ureg('M'))
  File "/usr/lib/python2.7/unittest/case.py", line 555, in assertAlmostEqual
    if round(abs(second-first), places) == 0:
  File "/home/ignacio/test/pint/pint/quantity.py", line 600, in __float__
    raise DimensionalityError(self._units, 'dimensionless')
DimensionalityError: Cannot convert from 'molar' to 'dimensionless'
```

which makes no sense to me. It looks like there is some bug unrelated to this PR.

hgrecco:
agreed. Weird.

Jellby:
Should I change test_issue29 to: watt = joule / second instead of molar = mole / liter ? It seems to pass that way, but I don't know if that is fully equivalent for the purpose. Any other idea?

hgrecco:
Change it as you have suggested.

crusaderky:
Closes #506 
Closes #542

PR #825, opened 2019-06-29
Allow np arrays of scalar quantities to be plotted

andrewgsavage:
This is to make plotting PintArrays work. PintArrays are converted to np arrays of scalar quantities.
This currently works:

```
import matplotlib.pyplot as plt
import numpy as np
import pint
import pandas as pd

ureg=pint.UnitRegistry()

Q_=ureg.Quantity

ureg.setup_matplotlib(True)

y = np.linspace(0, 30) * ureg.miles
x = np.linspace(0, 5) * ureg.hours

fig, ax = plt.subplots()
ax.plot(x, y, 'tab:blue')
```

This allows the following to work:

```
yy = np.array([item for item in y],dtype = "O")
xx = np.array([item for item in x],dtype = "O")

fig, ax = plt.subplots()
ax.plot(xx, yy, 'tab:blue')
```


hgrecco:
I had to split the travis fix to make it work. Would you mind spliting your commit and just keeping  6c880b

andrewgsavage:
Does that work?

hgrecco:
awesome

PR #837, opened 2019-07-24
Updated documentation for string formatting 

samgoodrich1:
Documented the rules for where the abbreviated units come from in the unit registry. Also briefly added some information regarding expanded formatting such as scientific notation. Also updated the default unit registry to fix the abbreviation for becquerel from 'becquerel' to 'Bq' by adding a duplicate 'Bq' in the first and 3rd columns and moving 'becquerel' to the 4th column.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -292,6 +292,17 @@ Pint's physical quantities can be easily printed:
    >>> print('The magnitude is {0.magnitude} with units {0.units}'.format(accel))
    The magnitude is 1.3 with units meter / second ** 2
 
+Pint supports float formatting for numpy arrays as well:
+
+.. doctest::
+
+   >>> accel = np.array([-1.1, 1e-6, 1.2505, 1.3]) * ureg['meter/second**2']
+   >>> # float formatting numpy arrays
+   >>> print('The array is {:.2f}'.format(accel))
+   The array is [-1.10 0.00 1.25 1.30] meter / second ** 2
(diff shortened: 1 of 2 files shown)


cpascual:
As far as I can tell, pint also supports [f-strings](https://www.python.org/dev/peps/pep-0498/) from python>=3.6 :

```python
>>> accel = 1.3 * ureg['meter/second**2'] 
>>> print(f'The str is {accel}')
The str is 1.3 meter / second ** 2
>>> print(f'The str is {accel:.3e}')
The str is 1.300e+00 meter / second ** 2
>>> print(f'The str is {accel:~}')
The str is 1.3 m / s ** 2
>>> print(f'The str is {accel:~.3e}')
The str is 1.300e+00 m / s ** 2
>>> print(f'The str is {accel:~H}')
The str is 1.3 m/s²  
```


Maybe this could also be mentioned in this docs?

samgoodrich1:
@cpascual yeah that makes sense to me. 

@hgrecco I don't understand how the dif for this PR is so complicated - I forked your master, made my modest changes to a few lines, and committed but I didn't delete lines or add more than about 20 lines.

hgrecco:
@samgoodrich1 I guess that your are not based in my current master. Rebase and it will look nice.

samgoodrich1:
@hgrecco - ok I managed to get something akin to a rebase and it looks much cleaner now. I added the f-string documentation recommended by @cpascual as well.

PR #842, opened 2019-08-13
Missing conversion definition when mutating context transformations.

dcupif:
Thanks for the library, we use it at our company and it's proven
really helpful so far. We are currently moving to Python 3.7 which
made me spot a potential bug in the `BaseRegistry` class (some
of our tests failed when upgrading).

Everything runs fine in Python 3.5, but in 3.7, we are getting a
`DimensionalityError` resulting from the ill-definition of a conversion.
Root cause is that we are mutating context transformations while
iterating over it.

This PR aims to avoid this potential problem. Since we modify the
transformations dictionary structure while iterating over it, we may
fail to iterate over all entries.

Explanation: "Iterating views while adding or deleting entries in the
dictionary may raise a `RuntimeError` or fail to iterate over all
entries"

https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects

This is my first time contributing to Pint, so please let me know if
something needs to be improved / changed.

### DIFF
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -1102,7 +1102,8 @@ class ContextRegistry(BaseRegistry):
         for ctx in ctxs:
             if getattr(ctx, '_checked', False):
                 continue
-            for (src, dst), func in ctx.funcs.items():
+            funcs_copy = dict(ctx.funcs)
+            for (src, dst), func in funcs_copy.items():
                 src_ = self._get_dimensionality(src)
                 dst_ = self._get_dimensionality(dst)
                 if src != src_ or dst != dst_:


hgrecco:
I agree that a change in this direccion is necessary. Why I don't like so much is the need to make this method private. The original idea to expose these methods was to allow users building/changing context objects. To be honest, I do not know how many people are using this but I still think is a good thing to do it. How about adding a safe way to iterate?

dcupif:
@hgrecco Sorry for the delayed answer. I thought the initial intention was to hide these methods which is why I made them private. I will roll back my changes to simply add a safe way to iterate. :+1:

dcupif:
@hgrecco small reminder to make sure you did not miss my update, I made the necessary changes according to your comments.

loriab:
The https://github.com/MolSSI/QCElemental/ project is also hitting the iterations error (below) upon using Python 3.8. We'd love to see this fix incorporated into pint. As an alternative, `for (src, dst), func in ctx.funcs.copy().items():` also works.

```
        # Check if the contexts have been checked first, if not we make sure
        # that dimensions are expressed in terms of base dimensions.
        for ctx in ctxs:
            if getattr(ctx, '_checked', False):
                continue
>           for (src, dst), func in ctx.funcs.items():
E           RuntimeError: dictionary keys changed during iteration
```

Thanks for the great project!

PR #846, opened 2019-08-21
Improve OffsetUnitCalculusError message.

### LINKED ISSUE #839: Improve OffsetUnitCalculusError message
Currently, the OffsetUnitCalculusError looks like this:

`Ambiguous operation with offset unit (degC)`

I have encountered people being puzzled by this, not knowing what to do against it. Would you think it a good idea to add a link to the docs at https://pint.readthedocs.io/en/latest/nonmult.html that explain how to resolve this error?

bilderbuchi:
Link to the docs page explaining the issue.
Closes #839.


bilderbuchi:
@hgrecco I don't think the build failure is my fault. The build script [only vaguely references](https://github.com/hgrecco/pint/blob/master/.travis.yml#L44) some Travis bug, but the current fix fails the build.

PR #868, opened 2019-09-03
New syntax to add aliases to already existing definitions

### LINKED ISSUE #860: Add aliases to already existing definitions
I'd like to write a custom definitions file that adds new aliases to definitions from defaults_en.txt - as opposed to creating entirely new units.

However, I could not find a way to do it without scouting for the original definition line and writing it again whole.

For example, defaults_en.txt contains:
```
calorie = 4.184 * joule = cal = thermochemical_calorie = cal_th
```

If I want to add an alias, I have to write in my custom definitions file:
```
calorie = 4.184 * joule = cal = thermochemical_calorie = cal_th = CUSTOM_CAL
```
whereas I'd like to just write something like:
```
calorie = ... = CUSTOM_CAL
```

crusaderky:
Closes #860 
FYI @jthielen

### DIFF
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -54,7 +54,7 @@ from .util import (logger, pi_theorem, solve_dependencies, ParserHelper,
 
 from .compat import tokenizer, string_types, meta
 from .definitions import (Definition, UnitDefinition, PrefixDefinition,
-                          DimensionDefinition)
+                          DimensionDefinition, AliasDefinition)
 from .converters import ScaleConverter
 from .errors import (DimensionalityError, UndefinedUnitError,
                      DefinitionSyntaxError, RedefinitionError)
@@ -263,6 +263,11 @@ class BaseRegistry(meta.with_metaclass(_Meta)):
         elif isinstance(definition, PrefixDefinition):
             d, di = self._prefixes, None
 
+        elif isinstance(definition, AliasDefinition):
+            d, di = self._units, self._units_casei
+            self._define_alias(definition, d, di)
+            return d[definition.name], d, di
+
         else:
             raise TypeError('{} is not a valid definition.'.format(definition))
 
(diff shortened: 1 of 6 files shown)


pint/registry.py
```diff
@@ -325,6 +330,13 @@ def _define_single_adder(self, key, value, unit_dict, casei_unit_dict):
         if casei_unit_dict is not None:
             casei_unit_dict[key.lower()].add(key)
 
+    def _define_alias(self, definition, unit_dict, casei_unit_dict):
+        unit = unit_dict[definition.name]
+        unit.add_aliases(*definition.aliases)
+        for alias in unit.aliases:
+            unit_dict[alias] = unit
+            casei_unit_dict[alias.lower()].add(alias)
```
crusaderky: I could not figure out how casei_unit_dict works - or find any tests for it. So I just copied what already is done for new units.

crusaderky:
I can see how this same syntax could be potentially exploited in the future for localization; e.g. something like

```
@locale it
    @alias meter = metro
@end
```
```python
>>> ureg = pint.UnitRegistry(locale="it")
>>> ureg("1 meter"), ureg("1 metro")
1 metro, 1 metro
```
However that would require a lot of extra effort which is far beyond the scope of this PR.

crusaderky:
Doesn't work with UnitRegistry.load_definitions() - working on a fix

crusaderky:
Fixed

crusaderky:

pint/registry.py
hgrecco: That is fine

hgrecco:
> I can see how this same syntax could be potentially exploited in the future for localization; e.g. something like
> 
> ```
> @locale it
>     @alias meter = metro
> @end
> ```
> 
> ```python
> >>> ureg = pint.UnitRegistry(locale="it")
> >>> ureg("1 meter"), ureg("1 metro")
> 1 metro, 1 metro
> ```
> 
> However that would require a lot of extra effort which is far beyond the scope of this PR.

It is a good idea, but I agree we need to move the dicussion to another PR. In particular, what worries me is that localization is not just aliasing but renaming

hgrecco:
+1 for the docs in the PR.

crusaderky:
Thanks!

> In particular, what worries me is that localization is not just aliasing but renaming

As somebody who had to endure working with localised Excel, I strongly advocate that localisation should impact exclusively the output of ``__repr__`` and ``__str__``.

Issue #855, opened 2019-08-28
Reciprocal of composite units


crusaderky:
In several contexts, one needs to convert between a composite unit and its reciprocal.

A few use cases:
- historically, car efficiency was measured in km/litre. More recently, it's become more common to express it in litres/100 km.
- volume is typically expressed in kg/m^3 or kg/litre ([mass]/[volume]). However, in oil trading, crude density is normally expressed in bbl/metric_ton ([volume]/[mass]).

This at the moment doesn't work:
```
>>> density = 7.3 * ureg("bbl/metric_ton")
>>> density.to("kg/l")
DimensionalityError: Cannot convert from 'barrel / metric_ton'
([length] ** 3 / [mass]) to 'kilogram / liter' ([mass] / [length] ** 3)
```

crusaderky:
This piece of documentation:


falls over if I actually trying using the context!
```python
>>> ureg("1 light_year").to("years", "ab")
DimensionalityError: Cannot convert from 'speed_of_light / light_year' (1 / [time]) to 'year' ([time])
```

crusaderky:
The same example falls over even sooner if the dimensions are not on the same right side of the fraction as stated by add_transformation:
```python
>>> ureg("1 MHz").to("nm", "ab")
DimensionalityError: Cannot convert from 'megahertz' (1 / [time]) to 'nanometer' ([length])
```

crusaderky:
Related: #854

PR #870, opened 2019-09-05
Anonymous contexts

crusaderky:
Allow creating ``pint.Context()`` without a name.
Also fixes a minor error in the documentation as discussed in #855.

### DIFF
--- a/pint/context.py
+++ b/pint/context.py
@@ -62,7 +62,7 @@ class Context(object):
 
     """
 
-    def __init__(self, name, aliases=(), defaults=None):
+    def __init__(self, name=None, aliases=(), defaults=None):
 
         self.name = name
         self.aliases = aliases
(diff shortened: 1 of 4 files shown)


pint/context.py
```diff
@@ -62,7 +62,7 @@ class Context(object):
 
     """
 
-    def __init__(self, name, aliases=(), defaults=None):
+    def __init__(self, name=None, aliases=(), defaults=None):
 
         self.name = name
```
hgrecco: I am not sure how things will work if somebody create two anonymous contexts. Therefore, I think that it would be better to create a random name.

crusaderky: I added tests for two anonymous contexts

crusaderky: @hgrecco ping - is there anything outstanding?

crusaderky:
Thanks!

PR #880, opened 2019-09-13
Global Quantity, Unit, and Measurement

crusaderky:
This PR addresses several issues I found myself facing when using pint in my own projects.

- I found myself systematically needing to reference UnitRegistry, Quantity, and Unit in sphinx docstrings and typing annotations.
- When prototyping and writing snippet code, I got frequently annoyed by having to typing ``ureg = pint.UnitRegistry()`` every time, even when I was perfectly content with defaults_en.txt (as most novice users will).
-  I found myself needing to access pint._APP_REGISTRY when creating a Quantity in a dask.distributed function.

Changes:
- Define global objects ``pint.Quantity``, ``pint.Unit``, and ``pint.Measurement``. They can be used directly, e.g. ``pint.Quantity(1, "kg")``. They use the global application registry.
- Fix bug where Measurement objects would be accidentally upcast to Quantity upon pickling
- All Exception classes are now accessible from the top-level module (also in intersphinx)
- New function ``pint.get_application_registry``
- Documentation for using ``pint.set_application_registry`` in one's application
- intersphinx users can now reference ``:class:`pint.UnitRegistry` `` instead of the internal implementation module ``:class:`pint.registry.UnitRegistry` ``
- Fixed a wealth of sphinx errors (not all)
- Unit tests for pickling/unpickling custom units through the global registry


crusaderky:

crusaderky:
@hgrecco ping

hgrecco:
I would like to test how this feels for a while.

crusaderky:
@hgrecco ETA to either merge or tweak requests?

crusaderky:
@grecco poke

crusaderky:
@hgrecco poke

crusaderky:
@hgrecco if you don't have time to go through this yourself, could you please nominate deputies? It's been 23 days now...

hgrecco:
I have been playing with this for a while:
1) I like how it feels
2) I like the implementation

but, it seems to me that a changed is needed in the method/property `SharedRegistryObject._Registry`. This method is triggered when the `_Registry` attribute has not been set (i.e. when the Quantity was not created after a `UnitRegistry` is created). If I am reading everything correctly, there is a chance that the registry the Quantity referes first to a Registry and then to other if the application registry is changed by the user in between. This is not a good idea. 

If that is the case, I would suggest that after calling `SharedRegistryObject._Registry` property once, the value is set as an attribute. I know that user can still change the `_Registry` attribute (It is Python after all) but it would require the user to write in an explicit manner.

crusaderky:
@hgrecco good points. My latest commits address them.

hgrecco:
Thanks @crusaderky for all the work and congratulations!

PR #887, opened 2019-09-27
Added support for comparing Quantity to NaN.

berntjf:
Hi,
I think this would be a nice little feature. At least I would like if it was there.

Cheers,
Jonas Fløde


hgrecco:
Compare a quantity to `NaN` returns `False` noew

Issue #881, opened 2019-09-13
Sphinx errors


crusaderky:
Downstream of #880, I'm still getting a few errors when running sphinx-build:

```
pint/docs/plotting.rst:32: WARNING: Exception occurred in plotting plotting-1
 from pint/docs/plotting.rst:
Traceback (most recent call last):
  File "pint/pint/quantity.py", line 1476, in __getitem__
    value = self._magnitude[key]
TypeError: 'int' object is not subscriptable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "site-packages/matplotlib/sphinxext/plot_directive.py", line 499, in run_code
    exec(code, ns)
  File "<string>", line 13, in <module>
  File "site-packages/matplotlib/axes/_axes.py", line 798, in axhline
    yy = self.convert_yunits(y)
  File "site-packages/matplotlib/artist.py", line 195, in convert_yunits
    return ax.yaxis.convert_units(y)
  File "site-packages/matplotlib/axis.py", line 1521, in convert_units
    if munits.ConversionInterface.is_numlike(x):
  File "site-packages/matplotlib/units.py", line 126, in is_numlike
    for thisx in x:
  File "pint/pint/quantity.py", line 1480, in __getitem__
    "supports indexing".format(self._magnitude))
TypeError: Neither Quantity object nor its magnitude (26400)supports indexing
pint/docs/plotting.rst:52: WARNING: Exception occurred in plotting plotting-2
 from pint/docs/plotting.rst:
Traceback (most recent call last):
  File "pint/pint/quantity.py", line 1476, in __getitem__
    value = self._magnitude[key]
TypeError: 'int' object is not subscriptable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "site-packages/matplotlib/sphinxext/plot_directive.py", line 499, in run_code
    exec(code, ns)
  File "<string>", line 16, in <module>
  File "site-packages/matplotlib/axes/_axes.py", line 798, in axhline
    yy = self.convert_yunits(y)
  File "site-packages/matplotlib/artist.py", line 195, in convert_yunits
    return ax.yaxis.convert_units(y)
  File "site-packages/matplotlib/axis.py", line 1521, in convert_units
    if munits.ConversionInterface.is_numlike(x):
  File "site-packages/matplotlib/units.py", line 126, in is_numlike
    for thisx in x:
  File "pint/pint/quantity.py", line 1480, in __getitem__
[...]

PR #891, opened 2019-10-16
Reimplement _Quantity.__iter__ to return an iterator

### LINKED ISSUE #751: iter for single value....

### LINKED ISSUE #760: matplotlib example from the docs failure (Neither Quantity object nor its magnitude supports indexing)

jthielen:
This PR implements the fix suggested by @crusaderky in https://github.com/hgrecco/pint/issues/751#issuecomment-542920002. Also, adds a test using `np.iterable()`.

- Closes #751 
- Closes #760 
- Should resolve the `TypeError`s in #881 (not sure if it should close it though, since it looks like there were also some extra warnings to be resolved)

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -148,21 +148,23 @@ class _Quantity(PrettyIPython, SharedRegistryObject):
 
         inst.__used = False
         inst.__handling = None
-        # Only instances where the magnitude is iterable should have __iter__()
-        if hasattr(inst._magnitude,"__iter__"):
-            inst.__iter__ = cls._iter
+
         return inst
 
-    def _iter(self):
-        """
-        Will be become __iter__() for instances with iterable magnitudes
-        """
-        # # Allow exception to propagate in case of non-iterable magnitude
-        it_mag = iter(self.magnitude)
-        return iter((self.__class__(mag, self._units) for mag in it_mag))
     @property
     def debug_used(self):
         return self.__used
+
+    def __iter__(self):
+        # Make sure that, if self.magnitude is not iterable, we raise TypeError as soon as one
+        # calls iter(self) without waiting for the first element to be drawn from the iterator
+        it_magnitude = iter(self.magnitude)
+
+        def it_outer():
+            for element in it_magnitude:
+                yield self.__class__(element, self._units)
+
+        return it_outer()
 
     def __copy__(self):
         ret = self.__class__(copy.copy(self._magnitude), self._units)
--- a/pint/testsuite/test_quantity.py
+++ b/pint/testsuite/test_quantity.py
@@ -392,6 +392,19 @@ class TestQuantity(QuantityTestCase):
         u_array_5 = self.Q_.from_list(u_seq)
         self.assertTrue(all(u_array_5 == u_array_ref))
 
+    @helpers.requires_numpy()
+    def test_iter(self):
+        # Verify that iteration gives element as Quantity with same units
+        x = self.Q_([0, 1, 2, 3], 'm')
+        self.assertQuantityEqual(next(iter(x)), self.Q_(0, 'm'))
+
+    def test_notiter(self):
+        # Verify that iter() crashes immediately, without needing to draw any
+        # element from it, if the magnitude isn't iterable
+        x = self.Q_(1, 'm')
+        with self.assertRaises(TypeError):
+            iter(x)
+
 
 class TestQuantityToCompact(QuantityTestCase):
 
--- a/pint/matplotlib.py
+++ b/pint/matplotlib.py
@@ -13,6 +13,8 @@ from __future__ import absolute_import
 
 import matplotlib.units
 
+from .util import iterable, sized
+
 
 class PintAxisInfo(matplotlib.units.AxisInfo):
     """Support default axis and tick labeling and default limits."""
@@ -31,7 +33,7 @@ class PintConverter(matplotlib.units.ConversionInterface):
 
     def convert(self, value, unit, axis):
         """Convert :`Quantity` instances for matplotlib to use."""
-        if hasattr(value,"__iter__"):
+        if iterable(value):
             return [self._convert_value(v, unit, axis) for v in value]
         else:
             return self._convert_value(value, unit, axis)
@@ -51,7 +53,7 @@ class PintConverter(matplotlib.units.ConversionInterface):
     @staticmethod
     def default_units(x, axis):
         """Get the default unit to use for the given combination of unit and axis."""
-        if hasattr(x,"__iter__") and len(x) > 0:
+        if iterable(x) and sized(x):
             return getattr(x[0], 'units', None)
         return getattr(x, 'units', None)
 
--- a/pint/testsuite/test_numpy.py
+++ b/pint/testsuite/test_numpy.py
@@ -235,6 +235,10 @@ class TestNumpyMethods(QuantityTestCase):
         for q, v in zip(self.q.flatten(), [1, 2, 3, 4]):
             self.assertEqual(q, v * self.ureg.m)
 
+    def test_iterable(self):
+        self.assertTrue(np.iterable(self.q))
+        self.assertFalse(np.iterable(1 * self.ureg.m))
+
     def test_reversible_op(self):
         """
         """
--- a/pint/testsuite/test_util.py
+++ b/pint/testsuite/test_util.py
@@ -9,7 +9,7 @@ from decimal import Decimal
 from pint.testsuite import BaseTestCase, QuantityTestCase
 from pint.util import (string_preprocessor, find_shortest_path, matrix_to_string,
                        transpose, tokenizer, find_connected_nodes, ParserHelper,
-                       UnitsContainer, to_units_container)
+                       UnitsContainer, to_units_container, iterable, sized)
 
 
 class TestUnitsContainer(QuantityTestCase):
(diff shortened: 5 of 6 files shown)


pint/quantity.py
```diff
@@ -148,22 +148,22 @@ def __new__(cls, value, units=None):
...
     @property
     def debug_used(self):
         return self.__used
 
+    def __iter__(self):
+        it_magnitude = iter(self.magnitude)
+
+        def it_outer():
+            for element in it_magnitude:
+                yield element
```
crusaderky:
```suggestion
                yield self.__class__(element, self._units)
```
plus approproriate unit test

pint/quantity.py
```diff
@@ -148,22 +148,22 @@ def __new__(cls, value, units=None):
...
-        """
-        # # Allow exception to propagate in case of non-iterable magnitude
-        it_mag = iter(self.magnitude)
-        return iter((self.__class__(mag, self._units) for mag in it_mag))
     @property
     def debug_used(self):
         return self.__used
 
+    def __iter__(self):
+        it_magnitude = iter(self.magnitude)
```
crusaderky:
```suggestion
        # Make sure that, if self.magnitude is not iterable, we raise TypeError as soon as one
        # calls iter(self) without waiting for the first element to be drawn from the iterator
        it_magnitude = iter(self.magnitude)
```

pint/quantity.py
jthielen: Yikes, yet another obvious thing I missed.

pint/quantity.py
```diff
@@ -148,22 +148,25 @@ def __new__(cls, value, units=None):
...
         return self.__used
 
+    def __iter__(self):
+        # Make sure that, if self.magnitude is not iterable, we raise TypeError as soon as one
+        # calls iter(self) without waiting for the first element to be drawn from the iterator
+        it_magnitude = iter(self.magnitude)
+
+        def it_outer():
+            for element in it_magnitude:
+                print(element)
```
crusaderky: Spurious print

pint/testsuite/test_quantity.py
```diff
@@ -392,6 +392,12 @@ def test_from_sequence(self):
         u_array_5 = self.Q_.from_list(u_seq)
         self.assertTrue(all(u_array_5 == u_array_ref))
 
+    @helpers.requires_numpy()
+    def test_iter(self):
+        # Verify that iteration gives element as Quantity with same units
+        x = self.Q_([0, 1, 2, 3], 'm')
+        self.assertQuantityEqual(next(iter(x)), self.Q_(0, 'm'))
+
```
crusaderky:
```suggestion

    def test_notiter(self):
        # Verify that iter() crashes immediately, without needing to draw any
        # element from it, if the magnitude isn't iterable
        x = self.Q_(1, 'm')
        with self.assertRaises(TypeError):
            iter(x)
```

pint/quantity.py
jthielen: :man_facepalming: I left that one slip through...I'm really not doing too well on this PR.

crusaderky:
Looks ready to merge to me :)

jthielen:
After running this past MetPy's tests, it looks like this doesn't resolve #760 / #881 entirely. A new error is raised; for example:

<details>
        def test_axvline():
            r"""Ensure that passing a quantity to axvline does not error."""
            fig, ax = plt.subplots()
    >       ax.axvline(0 * units('degC'))

    tests/units/test_units.py:57: 
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
    ../../miniconda3/envs/dev/lib/python3.7/site-packages/matplotlib/axes/_axes.py:873: in axvline
        self._process_unit_info(xdata=x, kwargs=kwargs)
    ../../miniconda3/envs/dev/lib/python3.7/site-packages/matplotlib/axes/_base.py:2135: in _process_unit_info
        kwargs = _process_single_axis(xdata, self.xaxis, 'xunits', kwargs)
    ../../miniconda3/envs/dev/lib/python3.7/site-packages/matplotlib/axes/_base.py:2118: in _process_single_axis
        axis.update_units(data)
    ../../miniconda3/envs/dev/lib/python3.7/site-packages/matplotlib/axis.py:1473: in update_units
        default = self.converter.default_units(data, self)
    ../pint/pint/matplotlib.py:54: in default_units
        if hasattr(x,"__iter__") and len(x) > 0:
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    self = <Quantity(0, 'degree_Celsius')>

        def __len__(self):
    >       return len(self._magnitude)
    E       TypeError: object of type 'int' has no len()

    ../pint/pint/quantity.py:1438: TypeError

</details>


due to:

https://github.com/hgrecco/pint/blob/998417f3f99711009b5e4fe5d3a5c841561dcad8/pint/matplotlib.py#L54

This line is likely also problematic:

https://github.com/hgrecco/pint/blob/998417f3f99711009b5e4fe5d3a5c841561dcad8/pint/matplotlib.py#L34

Unfortunately, there are no tests in the testsuite for `matplotlib.py`, so it's not surprising that this issue wasn't caught until tested downstream. I would think these checks should probably behave more like NumPy's `iterable` using a try/except block:

https://github.com/numpy/numpy/blob/3822ef836c240a7655a79634dba14c65b80e069b/numpy/lib/function_base.py#L258-L286

So, I have a couple questions:
- Would vendoring  `iterable` here from NumPy so that it can be used in `matplotlib.py` be a good fix, or is there a better solution?
- Would it be best to fix this here (to get things working as soon as possible), or in a followup PR that also adds tests for `matplotlib.py` (to avoid changes without tests)?

hgrecco:
> If this PRs breaks pint.matplotlib, this PR should also fix it IMHO.

agreed. Let's fix this and merge

jthielen:
@hgrecco I've added a commit with the updates for pint.matplotlib, so hopefully this should be good for a final review and merge.

Issue #882, opened 2019-09-15
Expected behavior of `np.*_like` functions


jthielen:
Based on discussions in https://github.com/pydata/xarray/pull/3238, it is unclear what the expected behavior of numpy functions like `zeros_like`, `ones_like`, and `empty_like` is with pint quantities (and depending on that behavior, how `full_like` should behave), in reference to their eventual `__array_function__` implementation in pint.

The two possibilities discussed there were:

- Have `zeros_like`, `ones_like`, and `empty_like` return base zeros, ones, or empty ndarrays, respectively, and have `full_like` simply match the unit of the `fill_value` argument
- Have `zeros_like`, `ones_like`, and `empty_like` return quantities of zeros, ones, or empty in the same unit as the `other` argument
  - If this is expected behavior, should `full_like`
    - always have the unit of `other`
    - always have the unit of `fill_value`
    - have the unit of `fill_value` only if it matches dimensionality with `other`, raising an error otherwise
    - be tolerant of missing units on `fill_value` and `other` or not?

---

Relevant discussion quoted from https://github.com/pydata/xarray/pull/3238:

@keewis:

> I tried adding tests for `zeros_like`, `ones_like` and `full_like`, but I'm not sure what the expected result should look like, at least for the last one. I would assume `zeros_like` and `ones_like` return a quantity with the same unit as `other`, but what is the unit for `full_like`? The unit of `other`, or the unit of `fill_value`? Also, what happens if either of them has no unit, or both have units that are incompatible with each other?

@jthielen:

> @keewis In [andrewgsavage/pint#6](https://github.com/andrewgsavage/pint/pull/6), I implemented `zeros_like`, `ones_like` and `empty_like` to return base ndarrays rather than quantities, and `full_like` to have the unit of `fill_value`. This seemed like the most sensible behavior to me, since it avoids the ambiguities you mention in the `full_like` case if they were based on the unit of `other`, and for many of my own use cases, I've often wanted a zeros/ones array with a different unit than the quantity whose shape I am basing it off of.
> 
> Does this behavior seem reasonable to you? Also, would this be something that should be cleared up with an issue on pint's end?

@keewis:

> I like your approach because it avoids all those issues, but stripping the units might be a surprise. Then again, what does a function named `*_like` promise? From the docs, it seems that only shape and dtype are guaranteed, so the question is whether or not the unit is part of the data or part of the type?

@jthielen:

[...]

Issue #878, opened 2019-09-12
NEP18 trouble when pint is being wrapped


crusaderky:
FYI @shoyer @hameerabbasi @keewis
numpy 1.17, xarray/dask/sparse/pint git tip

NEP18 doesn't seem to work correctly in several cases.
I'm still in the process of investigating what causes the issue(s).

Works:
- pint wraps around sparse
- pint wraps around dask.array
- xarray wraps around pint
- xarray wraps around sparse
- dask.array wraps around sparse
- xarray wraps around dask.array which wraps around sparse

Broken:
- **[1]** dask.array wraps around pint, and there are 2+ chunks
- **[2]** xarray wraps around pint which wraps around dask
- **[3]** xarray wraps around pint which wraps around sparse

**[1] dask.array wraps around pint, and there are 2+ chunks** 
At first sight, the legitimacy of this use case is arguable, as it feels much cleaner to always have pint wrapping around dask.array (and it saves a few of headaches when dask.distributed and custom UnitRegistries get involved, too, as you never need to pickle your Quantities).

However, the problems of pint->dask and the benefits of dask->pint become clear when one wraps a pint+dask object in xarray.
There, with pint around dask, one would need to write special case handling for pretty much every piece of xarray logic that today has special case handling for dask - which is, _a lot_, whereas with dask around pint I would expect everything to work out of the box as long as NEP18 compliance is respected by all libraries.
@shoyer I'd like to hear your opinion on this...

```python
>>> import dask.array as da
>>> import pint
>>> ureg = pint.UnitRegistry()
>>> q = ureg.Quantity([1, 2], "kg")
>>> da.from_array(q).compute() # single chunk works
<Quantity([1 2], 'kilogram')>
>> da.from_array(q, chunks=1).compute() . # With 2+ chunks, something's calling Quantity.__array__
array([1, 2])
```

**[2] xarray wraps around pint which wraps around dask** 
Following the reasoning of [1], this should happen only when a user manually builds the data, as opposed to calling ``xarray.Dataset.chunk()`` - which should be rare-ish. I'm tempted to write a single piece of logic in ``xarray.Variable.data.setter`` that detects the special pint->dask case and turns  it around to dask->pint.

```python
>>> import dask.array as da
>>> import pint
>>> ureg = pint.UnitRegistry()
>>> q = ureg.Quantity(da.from_array([1, 2]), "kg")
>>> q
[...]

Issue #845, opened 2019-08-21
Insuring interoperability of pint with other (non-unit) array-like types


jthielen:
When working with the new `__array_function__` implementation in #764, seeing what it takes to get pint `Quantity`'s working inside xarray (https://github.com/pydata/xarray/issues/525), and trying to wrap my head around [NEP 18](https://www.numpy.org/neps/nep-0018-array-function-protocol.html), I realized that (to the best of what I could find) pint's expected behavior with other ndarray-like types is not well-documented or tested for. And so, I hope that this issue could be a discussion on clarifying those expected behaviors and determining how to document and test them.

To start that discussion, I think it would make sense to split other array-like types into two categories, those that should wrap pint, and those that should be wrapped by pint, or in other words, determine where pint should fall in the "[type casting hierarchy](https://www.numpy.org/neps/nep-0013-ufunc-overrides.html#type-casting-hierarchy)". Based on https://github.com/pydata/xarray/issues/525#issuecomment-482651181 and other comments I've seen, a split like the following seems to be desired:

**Wrap pint Quantity**

- xarray DataArray

**Wrapped by pint Quantity**

- dask array
- cupy array
- sparse array
- numpy masked array (and of course ndarray)

(Pandas seems to have special handling through pint-pandas, and hopefully others that I'm missing can be added to either list as appropriate)

Does this make sense as a division, and should a list like this be documented anywhere?

Also, once #764 is in place, it seems like there will be a need for explicit identification of types in each category...types that should wrap pint are used in https://github.com/hgrecco/pint/blob/4dcbe78766269f985d1898601d8f5384a1511f2c/pint/quantity.py#L70, and types that should be wrapped by pint would be used with implementing the functionality mentioned in https://github.com/hgrecco/pint/pull/764/#issuecomment-523272038. However, what should be done about array-like types that are unknown to pint, and how can consistent behavior between different `Quantity`-related operations be assured?

Finally, should pint implement tests for all the types it expects to be able to wrap? I got the impression from https://github.com/pydata/xarray/issues/525 and https://github.com/pydata/xarray/pull/2956 that tests fit best with the type higher in the type casting hierarchy (the one doing the wrapping).

Please do correct me on anything I am mistaken about with the above rambling :smile:.

[...]

Issue #867, opened 2019-09-02
Quantity.prod() result unit is incorrect with axis or where argument


jthielen:
Right now, the `.prod()` method assumes the full size of the input is collapsed in the result. This gives incorrect results when the `axis` or `where` arguments are supplied, as seen below:

```python
import pint

ureg = pint.UnitRegistry()

q = [[1, 2], [3, 4]] * ureg.m

print(q.prod())
print(q.prod(axis=0))
print(q.prod(where=[True, False]))
```
```
24 meter ** 4
[3 8] meter ** 4
3 meter ** 4
```

The unit on the first result where the full array is collapsed to a scalar is correct, but the other two results should have `meter ** 2` as the unit.

I don't have a good fix for this yet, since this is the same problem mentioned in https://github.com/hgrecco/pint/pull/764#issuecomment-523749008. If anyone has any suggestions for a performant way to determine how many unit multiplications occur given both `axis` and `where` arguments, please do let me know! Otherwise, I'll try to figure something out and get a PR in for this or include it alongside a `prod` implementation for `__array_function__`.

keewis:
couldn't we use the result of `numpy` to compute the units? I'm thinking of something like
```python
def prod(quantity, *args, **kwargs):
    raw_array = quantity.magnitude
    raw_result = raw_array.prod(*args, **kwargs)
    
    exponent = np.prod(raw_array.shape) / np.prod(raw_result.shape)
    units = quantity.units ** exponent

    return ureg.Quantity(raw_result, units)

q = np.linspace(1, 2, 5 * 10).reshape(5, 10) * ureg.s
```
And for `where`, this could be `exponent = np.sum(where)`.

I'm not sure how to compute the units if both `axis` and `where` are specified, though.

jthielen:
@keewis Good point that in the either/or case of `axis` and `where`, we can use NumPy functions to compute the result. It's probably better to move forward with that as a "good enough but not perfect solution", and only operate on dimensionless when both the `axis` and `where` arguments are provided.

Cases that I had in mind like the following are well-defined, but couldn't come up with a good way to handle in general:
```python
np.prod(np.arange(6).reshape((2, 3)) * ureg.meter, axis=1, where=[True, False, True])
```
```
[ 0 15] meter ** 2
```

keewis:
I guess we could try to filter the cases where this works. I've been trying something like
```python
a = np.linspace(1, 2, 20).reshape(4, 5) * ureg.m
where = a < 1.3
[...]

PR #905, opened 2019-11-29
NEP-18 Compatibility

### LINKED ISSUE #126: Support for numpy NaN functions: nan{min,max,mean,argmin,argmax,sum}

### LINKED ISSUE #396: Add support for median, average, and others

### LINKED ISSUE #424: Calling the ufunc in __array_wrap__ is probably a bad idea
Since by that point, the ufunc has already run once. Although putting it in `__array_prepare__` may also not solve the problem, since that:

> `__array_prepare__` **should not attempt to access the array data** or resize the array, it is intended for setting the output array type, updating attributes and metadata, and performing any checks based on the input that may be desired before computation begins.

### LINKED ISSUE #547: Raise an error if a value returned by a wrapped function is a Quantity
Ps.- Only for values that are supposed to be wrapped (i.e. not those skipped).

### LINKED ISSUE #553: How to best deal with copyto and pint?
The `copyto` function complains that `argument must be numpy.ndarray, not Quantity`. I guess that this numpy function is one of the few that is not supported by pint currently. My current approach to work around this is:

```python    
# ...
tmp = destination.magnitude
np.copyto(tmp, source.magnitude)
destination= tmp * source.units
```

Are there any cleaner options to use `copyto` with units?

### LINKED ISSUE #617: Add numpy.interp to list of functions that don't work
numpy.interp doesn't work as expected for pint quantities.

### LINKED ISSUE #619: Pint's usage with Numpy's concatenate
```python
import numpy as np

from pint import UnitRegistry

ureg = UnitRegistry()

Q_ = ureg.Quantity

args = []

for k in range(0,17):
   uwnd = Q_(np.ones((1,73,144)),'meter / second')
    args.append(uwnd)
uwndTot = np.vstack(args)
print(uwndTot.shape)
for element in args:
   print(type(element)

```
I am using Pint and when  I  print the shape uwndTot I get
(17,73,144).

When I  print the values of element it turns out to be

<class 'pint.quantity.build_quantity_class.<locals>.Quantity'>

However if I iterate through uwndTot I get

```python
for element in uwndTot:
    print(type(element))
```
I get 

<class 'numpy.ndarray'>

I want to be able to use numpy's concatenate with Pint. Is that not possible ? I need the units of meters /second in the 3d dimensional array uwndTot. How do I do that ?

### LINKED ISSUE #682: __array_wrap__ broken for `insert`, etc.
The following code:
```python
import numpy as np
import pint
ureg = pint.UnitRegistry()
a = np.arange(3) * ureg.meter
np.insert(a, 2, 4 * ureg.meter)
```
produces:
```pytb
~/miniconda3/envs/py36/lib/python3.6/site-packages/numpy/lib/function_base.py in insert(arr, obj, values, axis)
   4439         new[tuple(slobj)] = arr[tuple(slobj2)]
   4440         if wrap:
-> 4441             return wrap(new)
   4442         return new
   4443     elif indices.size == 0 and not isinstance(obj, np.ndarray):

~/miniconda3/envs/py36/lib/python3.6/site-packages/pint/quantity.py in __array_wrap__(self, obj, context)
   1342 
   1343     def __array_wrap__(self, obj, context=None):
-> 1344         uf, objs, i_out = context
   1345 
   1346         # if this ufunc is not handled by Pint, pass it to the magnitude.

TypeError: 'NoneType' object is not iterable
```

I have no idea if `insert` has any hope of working right now, but that exception comes from code that is obviously wrong:

https://github.com/hgrecco/pint/blob/241bf7df1404e26d202df64e12e772d111803c31/pint/quantity.py#L1428-L1430

Here `context` defaults to `None` (as is [documented for numpy](https://docs.scipy.org/doc/numpy/reference/arrays.classes.html#numpy.class.__array_wrap__)), but the first line in `__array_wrap__` tries to unpack it unconditionally. I have no idea what the right behavior is in this case.

### LINKED ISSUE #700: UnitStrippedWarning when calling numpy ufuncs
Commit 4be294cf5621cf7278a05aab478de9936d787c9a introduced `UnitStrippedWarning`, and it is now extremely common for me to see this warning (I assume every time I use a ufunc).

For example, simply calling `np.cos` on a quantity will produce the warning. Is there some other way I'm supposed to use numpy ufuncs with quantities, or is this a bug?

If this is a limitation of the current ufunc implementation, is this something that the new [`__array_ufunc__`](https://docs.scipy.org/doc/numpy/reference/arrays.classes.html#numpy.class.__array_ufunc__) api can help with?

### LINKED ISSUE #790: The __array_function__ protocol for overriding NumPy's API
We have been drafting a new protocol to allow for overriding NumPy functions, e.g., for `np.concatenate`: http://www.numpy.org/neps/nep-0018-array-function-protocol.html

It would be great to experiment with getting this working in pint, since arrays with units are one of the intended use-cases.

### LINKED ISSUE #821: NumPy clip(), minimum(), maximum() broken
Running the code with `np.clip()`

```python
import pint
ureg = pint.UnitRegistry()
a = 3 * ureg.meter
b = 0 * ureg.meter
c = 2 * ureg.meter
import numpy as np
np.clip(a, b, c)
```

produces an error similar to #682.

```python-traceback
/usr/local/lib/python3.7/site-packages/pint/quantity.py:1377: UnitStrippedWarning: The unit of the quantity is stripped.
  warnings.warn("The unit of the quantity is stripped.", UnitStrippedWarning)
AttributeError                            Traceback (most recent call last)
/usr/local/lib/python3.7/site-packages/numpy/core/fromnumeric.py in _wrapfunc(obj, method, *args, **kwds)
     55     try:
---> 56         return getattr(obj, method)(*args, **kwds)
     57

/usr/local/lib/python3.7/site-packages/pint/quantity.py in clip(self, first, second, out, **kwargs)
   1294
-> 1295         return self.__class__(self.magnitude.clip(**kwargs), self._units)
   1296

AttributeError: 'int' object has no attribute 'clip'

During handling of the above exception, another exception occurred:

<ipython-input-7-1a75bcf59855> in <module>
----> 1 np.clip(a, b, c)

/usr/local/lib/python3.7/site-packages/numpy/core/fromnumeric.py in clip(a, a_min, a_max, out)
   1956
   1957     """
-> 1958     return _wrapfunc(a, 'clip', a_min, a_max, out=out)
   1959
   1960

/usr/local/lib/python3.7/site-packages/numpy/core/fromnumeric.py in _wrapfunc(obj, method, *args, **kwds)
     64     # a downstream library like 'pandas'.
     65     except (AttributeError, TypeError):
---> 66         return _wrapit(obj, method, *args, **kwds)
     67
     68

/usr/local/lib/python3.7/site-packages/numpy/core/fromnumeric.py in _wrapit(obj, method, *args, **kwds)
     48         if not isinstance(result, mu.ndarray):
     49             result = asarray(result)
---> 50         result = wrap(result)
     51     return result
     52

/usr/local/lib/python3.7/site-packages/pint/quantity.py in __array_wrap__(self, obj, context)
   1566
   1567     def __array_wrap__(self, obj, context=None):
-> 1568         uf, objs, i_out = context
   1569
   1570         # if this ufunc is not handled by Pint, pass it to the magnitude.

TypeError: cannot unpack non-iterable NoneType object
```

Might be unrelated, but `np.minimum()` and `np.maximum()` produces output without unit.

```pycon
>>> np.minimum(a, b)
0
>>> np.maximum(a, b)
3
```

jthielen:
Building off of the implementation of `__array_function__` in #764, this PR adds compatibility with NEP-18 in Pint (mostly in the sense of Quantity having `__array_function__` and being wrappable as a duck array; for Quantity wrapping other duck arrays, see #845). Many tests are added of NumPy functions being used with Pint Quantities by way of `__array_function__`.

Accompanying changes that were needed as a part of this implementation include:
- a complete refactor of `__array_ufunc__` and ufunc attribute fallbacks to work in parallel with `__array_function__`
- promoting `_eq` in `quantity` to `eq` in `compat`
- preliminary handling of array-like compatibility by defining upcast types and attempting to wrap and defer to all others (a follow-up PR, or set of PRs, will be needed to completely address #845 / #878)

Closes #126 
Closes #396 
Closes #424
Closes #547 
Closes #553
Closes #617
Closes #619
Closes #682 
Closes #700
Closes #764
Closes #790
Closes #821

### DIFF
--- a/docs/numpy.rst
+++ b/docs/numpy.rst
@@ -1,12 +1,13 @@
 .. _numpy:
 
 
-NumPy support
+NumPy Support
 =============
 
-The magnitude of a Pint quantity can be of any numerical type and you are free
-to choose it according to your needs. In numerical applications, it is quite
-convenient to use `NumPy ndarray`_ and therefore they are supported by Pint.
+The magnitude of a Pint quantity can be of any numerical scalar type, and you are free
+to choose it according to your needs. For numerical applications requiring arrays, it is
+quite convenient to use `NumPy ndarray`_ (or `ndarray-like types supporting NEP-18`_),
+and therefore these are the array types supported by Pint.
 
 First, we import the relevant packages:
 
@@ -81,11 +82,11 @@ is applied before the requested calculation:
     >>> print(angles)
     [ 0.64350111  0.92729522] radian
 
-You can convert the result to degrees using the corresponding NumPy function:
+You can convert the result to degrees using usual unit conversion:
 
 .. doctest::
 
-    >>> print(np.rad2deg(angles))
+    >>> print(angles.to('degree'))
     [ 36.86989765  53.13010235] degree
 
 Applying a function that expects angles to a quantity with a different dimensionality
@@ -109,13 +110,16 @@ The following ufuncs_ can be applied to a Quantity object:
 - **Comparison functions**: greater, greater_equal, less, less_equal, not_equal, equal
 - **Floating functions**: isreal,iscomplex, isfinite, isinf, isnan, signbit, copysign, nextafter, modf, ldexp, frexp, fmod, floor, ceil, trunc
 
-And the following `NumPy ndarray methods`_ and functions:
+And the following NumPy functions:
 
-- sum, fill, reshape, transpose, flatten, ravel, squeeze, take, put, repeat, sort, argsort, diagonal, compress,  nonzero, searchsorted, max, argmax, min, argmin, ptp, clip, round, trace, cumsum, mean, var, std, prod, cumprod, conj, conjugate, flatten
+- alen, amax, amin, append, argmax, argmin, argsort, around, atleast_1d, atleast_2d, atleast_3d, average, block, broadcast_to, clip, column_stack, compress, concatenate, copy, copyto, count_nonzero, cross, cumprod, cumproduct, cumsum, diagonal, diff, dot, dstack, ediff1d, einsum, empty_like, expand_dims, fix, flip, full_like, gradient, hstack, insert, interp, isclose, iscomplex, isin, isreal, linspace, mean, median, meshgrid, moveaxis, nan_to_num, nanargmax, nanargmin, nancumprod, nancumsum, nanmax, nanmean, nanmedian, nanmin, nanpercentile, nanstd, nanvar, ndim, nonzero, ones_like, percentile, ptp, ravel, result_type, rollaxis, rot90, round\_, searchsorted, shape, size, sort, squeeze, stack, std, sum, swapaxes, tile, transpose, trapz, trim_zeros, unwrap, var, vstack, where, zeros_like
 
-`Quantity` is not a subclass of `ndarray`. This might change in the future, but for this reason  functions that call `numpy.asanyarray` are currently not supported. These functions are:
+And the following `NumPy ndarray methods`_:
 
-- unwrap, trapz, diff, ediff1d, fix, gradient, cross, ones_like
+- argmax, argmin, argsort, astype, clip, compress, conj, conjugate, cumprod, cumsum, diagonal, dot, fill, flatten, flatten, item, max, mean, min, nonzero, prod, ptp, put, ravel, repeat, reshape, round, searchsorted, sort, squeeze, std, sum, take, trace, transpose, var
+
+Pull requests are welcome for any NumPy function, ufunc, or method that is not currently
+supported.
 
 
 Comments
@@ -133,22 +137,31 @@ will be raised.
 
 In some functions that take 2 or more arguments (e.g. `arctan2`), the second
 argument is converted to the units of the first. Again, a `DimensionalityError`
-exception will be raised if this is not possible.
+exception will be raised if this is not possible. ndarray or ndarray-like arguments
+are generally treated as if they were dimensionless quantities, except for declared
+upcast types to which Pint defers (see
+<https://numpy.org/neps/nep-0013-ufunc-overrides.html>). To date, these "upcast types" are:
+
+- ``PintArray``, as defined by pint-pandas
+- ``Series``, as defined by pandas
+- ``DataArray``, as defined by xarray
+
+To achive these function and ufunc overrides, Pint uses the ``__array_function__`` and
+``__array_ufunc__`` protocols respectively, as recommened by NumPy. This means that
+functions and ufuncs that Pint does not explicitly handle will error, rather than return
+a value with units stripped (in contrast to Pint's behavior prior to v0.10). For more
+information on these protocols, see
+<https://docs.scipy.org/doc/numpy-1.17.0/user/basics.dispatch.html>.
 
 This behaviour introduces some performance penalties and increased memory
 usage. Quantities that must be converted to other units require additional
-memory and CPU cycles. On top of this, all `ufuncs` are implemented in the
-`Quantity` class by overriding `__array_wrap__`, a NumPy hook that is executed
-after the calculation and before returning the value. To our knowledge, there
-is no way to signal back to NumPy that our code will take care of the
-calculation. For this reason the calculation is actually done twice:
-first in the original ndarray and then in the one that has been
-converted to the right units. Therefore, for numerically intensive code, you
+memory and CPU cycles. Therefore, for numerically intensive code, you
 might want to convert the objects first and then use directly the magnitude.
 
 
 
 
 .. _`NumPy ndarray`: http://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.html
+.. _`ndarray-like types supporting NEP-18`: https://numpy.org/neps/nep-0018-array-function-protocol.html
 .. _ufuncs: http://docs.scipy.org/doc/numpy/reference/ufuncs.html
 .. _`NumPy ndarray methods`: http://docs.scipy.org/doc/numpy/reference/arrays.ndarray.html#array-methods
--- a/docs/numpy.rst
+++ b/docs/numpy.rst
@@ -156,7 +156,8 @@ information on these protocols, see
 This behaviour introduces some performance penalties and increased memory
 usage. Quantities that must be converted to other units require additional
 memory and CPU cycles. Therefore, for numerically intensive code, you
-might want to convert the objects first and then use directly the magnitude.
+might want to convert the objects first and then use directly the magnitude,
+such as by using Pint's `wraps` utility (see :ref:`wrapping`).
 
 
 
--- a/pint/numpy_func.py
+++ b/pint/numpy_func.py
@@ -228,7 +228,7 @@ copy_units_output_ufuncs = ['ldexp', 'fmod', 'mod', 'remainder']
 op_units_output_ufuncs = {'var': 'square', 'prod': 'size', 'multiply': 'mul',
                           'true_divide': 'div', 'divide': 'div', 'floor_divide': 'div',
                           'sqrt': 'sqrt', 'square': 'square', 'reciprocal': 'reciprocal',
-                          'std': 'sum', 'sum': 'sum', 'cumsum': 'sum'}
+                          'std': 'sum', 'sum': 'sum', 'cumsum': 'sum', 'matmul': 'mul'}
 
 
 # Perform the standard ufunc implementations based on behavior collections
--- a/pint/numpy_func.py
+++ b/pint/numpy_func.py
@@ -224,7 +224,7 @@ matching_input_copy_units_output_ufuncs = ['compress', 'conj', 'conjugate', 'cop
                                            'squeeze', 'swapaxes', 'take', 'trace',
                                            'transpose', 'ceil', 'floor', 'hypot', 'rint',
                                            'copysign', 'nextafter', 'trunc', 'absolute',
-                                           'negative']
+                                           'negative', 'maximum', 'minimum', 'fabs']
 copy_units_output_ufuncs = ['ldexp', 'fmod', 'mod', 'remainder']
 op_units_output_ufuncs = {'var': 'square', 'prod': 'size', 'multiply': 'mul',
                           'true_divide': 'div', 'divide': 'div', 'floor_divide': 'div',
--- a/pint/numpy_func.py
+++ b/pint/numpy_func.py
@@ -7,7 +7,7 @@
     :license: BSD, see LICENSE for more details.
 """
 
-from .compat import is_upcast_type, np, string_types
+from .compat import NP_NO_VALUE, is_upcast_type, np, string_types, eq
 from .errors import DimensionalityError
 from .util import iterable, sized
 
@@ -100,8 +100,9 @@ def get_op_output_unit(unit_op, first_input_units, all_args=[], size=None):
                 product /= x.units
         result_unit = product
     elif unit_op == "div":
-        product = first_input_units._REGISTRY.parse_units('')
-        for x in all_args:
+        # Start with first arg in numerator, all others in denominator
+        product = getattr(all_args[0], 'units', first_input_units._REGISTRY.parse_units(''))
+        for x in all_args[1:]:
             if hasattr(x, 'units'):
                 product /= x.units
         result_unit = product
@@ -139,6 +140,10 @@ def implements(numpy_func_string, func_type):
 
 def implement_func(func_type, func_str, input_units=None, output_unit=None):
     """TODO"""
+    # If NumPy is not available, do not attempt implement that which does not exist
+    if np is None:
+        return
+
     func = getattr(np, func_str)
 
     @implements(func_str, func_type)
@@ -151,7 +156,7 @@ def implement_func(func_type, func_str, input_units=None, output_unit=None):
                 *args, pre_calc_units=first_input_units, **kwargs)
         else:
             # Match all input args/kwargs to input_units, or if input_units is None, simply
-            # strip units 
+            # strip units
             stripped_args, stripped_kwargs = convert_to_consistent_units(
                 *args, pre_calc_units=input_units, **kwargs)
 
@@ -163,8 +168,11 @@ def implement_func(func_type, func_str, input_units=None, output_unit=None):
             return result_magnitude
         elif output_unit == "match_input":
             result_unit = first_input_units
-        else:
+        elif output_unit in ['sum', 'mul', 'delta', 'delta,div', 'div', 'variance', 'square',
+                             'sqrt', 'reciprocal', 'size']:
             result_unit = get_op_output_unit(output_unit, first_input_units, args_and_kwargs)
+        else:
+            result_unit = output_unit
 
         return first_input_units._REGISTRY.Quantity(result_magnitude, result_unit)
 
@@ -176,6 +184,7 @@ TODO: document as before
 
 
 """
+strip_unit_input_output_ufuncs = ['isnan', 'isinf', 'isfinite', 'signbit']
 matching_input_bare_output_ufuncs = ['equal', 'greater', 'greater_equal', 'less',
                                      'less_equal', 'not_equal']
 matching_input_set_units_output_ufuncs = {'arctan2': 'radian'}
@@ -205,22 +214,28 @@ set_units_ufuncs = {'cumprod': ('', ''),
                     'rad2deg': ('radian', 'degree'),
                     'logaddexp': ('', ''),
                     'logaddexp2': ('', '')}
+# TODO (#905 follow-up): while this matches previous behavior, some of these have optional
+# arguments that should not be Quantities. This should be fixed, and tests using these
+# optional arguments should be added.
 matching_input_copy_units_output_ufuncs = ['compress', 'conj', 'conjugate', 'copy',
                                            'diagonal', 'max', 'mean', 'min',
                                            'ptp', 'ravel', 'repeat', 'reshape', 'round',
                                            'squeeze', 'swapaxes', 'take', 'trace',
                                            'transpose', 'ceil', 'floor', 'hypot', 'rint',
-                                           'add', 'subtract', 'copysign', 'nextafter',
-                                           'trunc', 'frexp', 'absolute', 'negative']  # TODO: review extra args/kwargs
+                                           'copysign', 'nextafter', 'trunc', 'absolute',
+                                           'negative']
 copy_units_output_ufuncs = ['ldexp', 'fmod', 'mod', 'remainder']
 op_units_output_ufuncs = {'var': 'square', 'prod': 'size', 'multiply': 'mul',
                           'true_divide': 'div', 'divide': 'div', 'floor_divide': 'div',
-                          'remainder': 'div', 'sqrt': 'sqrt', 'square': 'square',
-                          'reciprocal': 'reciprocal', 'std': 'sum', 'sum': 'sum',
-                          'cumsum': 'sum'}
+                          'sqrt': 'sqrt', 'square': 'square', 'reciprocal': 'reciprocal',
+                          'std': 'sum', 'sum': 'sum', 'cumsum': 'sum'}
 
 
 # Perform the standard ufunc implementations based on behavior collections
+for ufunc_str in strip_unit_input_output_ufuncs:
+    # Ignore units
+    implement_func('ufunc', ufunc_str, input_units=None, output_unit=None)
+
 for ufunc_str in matching_input_bare_output_ufuncs:
     # Require all inputs to match units, but output base ndarray
     implement_func('ufunc', ufunc_str, input_units='all_consistent', output_unit=None)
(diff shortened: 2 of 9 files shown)


keewis:
thank you for working on this, @jthielen. Don't worry about the slow progress, as long as it is not completely on hold.

Running the xarray tests results in quite a few errors (probably more if we count the xfails). I can't do much right now, but I will try to investigate and report more this weekend.

**re** `inspect`: I already mentioned it in #764, but I strongly support dropping py2 (not my call, though).
Also, I'd say let's move adding tests for wrapping other NEP-18-compatible arrays to another PR so we can keep this as small as possible.

pint/quantity.py
```diff
@@ -79,6 +78,262 @@ def wrapped(self, *args, **kwargs):
...
+
+# Handle single unit argument operations (axis ops, aggregations, etc.)
+for func_str in ['expand_dims', 'squeeze', 'rollaxis', 'moveaxis', 'fix', 'around', 'diagonal',
+                 'mean', 'ptp', 'ravel', 'round_', 'sort', 'median', 'nanmedian', 'transpose',
+                 'flip', 'copy', 'trim_zeros', 'average', 'nanmean']:
+    func = getattr(np, func_str)
+    @implements(func)
+    def implementation(a, *args, **kwargs):
+        (a,), output_wrap = unwrap_and_wrap_consistent_units(a)
+        return output_wrap(func(a, *args, **kwargs))
```
keewis: there's something wrong with this and all similar closures below: at the end of quantity, all functions are implemented by wrapping `atleast_3d` (look at `HANDLED_FUNCTIONS[func].__globals__["func"]`).

I'm no expert on closures so I don't know why this happens or how to solve this (I will look into it, though). Does anyone else have any suggestions?

pint/quantity.py
```diff
@@ -79,6 +78,262 @@ def wrapped(self, *args, **kwargs):
...
+for func_str in ['gradient', ]:
+    implement_func(func_str, input_units=None, output_unit='delta,div')
+
+for func_str in ['var', 'nanvar']:
+    implement_func(func_str, input_units=None, output_unit='variance')
+
+# TODO: broadcast_to (how to handle subok?)
+# TODO: 'amax', 'amin', 'nanmax', 'nanmin' (version dependent signatures)
+# TODO: nan_to_num (expected behavior with input values, sensible defaults?)
+
```
keewis: Since it is being used a few times in xarray's codebase: can we also add a TODO for `result_type`?

pint/quantity.py
keewis: looks like creating `implementation` in a wrapper function (similar to `implement_func`) will solve this. We also should look at `HANDLED_FUNCTIONS[func].__closure__` and `HANDLED_FUNCTIONS[func].__code__.co_freevars` instead.

pint/quantity.py
```diff
@@ -79,6 +78,262 @@ def wrapped(self, *args, **kwargs):
...
+    for arg in args_combo:
+        if isinstance(arg, Quantity):
+            out_units = arg.units
+        elif _is_quantity_sequence(arg):
+            out_units = arg[0].units
+        if out_units is not None:
+            break
+    return out_units
+    
+def convert_to_consistent_units(pre_calc_units=None, *args, **kwargs):
```
keewis: it looks like `pre_calc_units` should be a keyword (-only?) argument, since it's easy to forget to pass `None` if we want to keep the default

pint/quantity.py
```diff
@@ -79,6 +78,262 @@ def wrapped(self, *args, **kwargs):
...
+            fill_value.units)
+    else:
+        return (np.ones_like(a, dtype=dtype, order=order, subok=subok, shape=shape)
+                * fill_value)
+
+@implements(np.interp)
+def _interp(x, xp, fp, left=None, right=None, period=None):
+    # Need to handle x and y units separately
+    (x, xp, period), _ = unwrap_and_wrap_consistent_units(x, xp, period)
+    (fp, right, left), output_wrap = convert_to_consistent_units(fp, left, right)
```
keewis: this looks like you swapped `unwrap_and_wrap_consistent_units` and `convert_to_consistent_units` (and maybe add `None` to the arguments of `convert_to_consistent_units`?)

jthielen:
Thanks for the initial glace over the PR @keewis! In all my tinkering with this, I had forgotten to run the most recent set of changes past the test suite, which would have caught most of these. Whoops!

I've pushed some temporary fixups, which should get all the tests passing except the unimplemented ones. These changes may or may not be for naught after I redo the implementation logic using `inspect.signature`, which I hope to get to yet this weekend or early this coming week.

keewis:
we should also keep in mind that with pint we need to make these changes optional: the tests without numpy installed currently fail because in that case `np` is `None`.

pint/testsuite/test_numpy.py
```diff
@@ -154,34 +521,82 @@ def test_clip(self):
...
+    def test_nancumsum_numpy_func(self):
+        self.assertQuantityEqual(np.nancumsum(self.q_nan, axis=0), [[1, 2], [4, 2]] * self.ureg.m)
+
     def test_mean(self):
         self.assertEqual(self.q.mean(), 2.5 * self.ureg.m)
 
+    @helpers.requires_array_function_protocol()
+    def test_mean_numpy_func(self):
+        self.assertEqual(np.mean(self.q), 2.5 * self.ureg.m)
+        self.assertEqual(np.mean(q_temperature), self.Q_(2.5, self.ureg.m))
```
keewis: I guess that's the last error not related to missing implementations, except from the one related to `np.power`?
```suggestion
        self.assertEqual(np.mean(self.q_temperature), self.Q_(2.5, self.ureg.degC))
```

pint/compat/__init__.py
```diff
@@ -66,6 +67,26 @@ def u(x):
...
+'To hide this warning use the following code to import pint:'
+"""
+import warnings
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    import pint
+To disable the new behavior, see
+https://www.numpy.org/neps/nep-0018-array-function-protocol.html#implementation
+---
+""")
```
crusaderky: Wouldn't it be nicer to have the whole thing wrapped by ``"""``? (plus strip/rstrip/dedent as needed)

pint/compat/__init__.py
```diff
@@ -66,6 +67,26 @@ def u(x):
...
 
+# TODO: remove this warning after v0.10
+class BehaviorChangeWarning(UserWarning):
+    pass
+_msg = ('The way pint handles numpy operations has changed with '
+'the implementation of NEP 18. Unimplemented numpy operations '
+'will now fail instead of making assumptions about units. Some '
+'functions, eg concat, will now return Quanties with units, '
+'where they returned ndarrays previously. See '
+'https://github.com/hgrecco/pint/pull/xxxx. '
```
crusaderky:
```suggestion
'https://github.com/hgrecco/pint/pull/905. '
```

pint/compat/__init__.py
```diff
@@ -85,6 +106,23 @@ def _to_magnitude(value, force_ndarray=False):
...
+
+            np.concatenate([FakeArray()])
+            return True
+        except ValueError:
+            return False
+
+    HAS_NUMPY_ARRAY_FUNCTION = _test_array_function_protocol()
+
+    if HAS_NUMPY_ARRAY_FUNCTION:
+        warnings.warn(_msg, BehaviorChangeWarning)
```
crusaderky: This is going to annoy a lot of people who don't use pint with numpy but just happen to have numpy 1.17 installed in the same environment.
Could we issue the warning upon the first call to ``__array_function__``? Or upon the creation of the first Quantity with numpy data?

hgrecco:
> **re** `inspect`: I already mentioned it in #764, but I strongly support dropping py2 (not my call, though).

See #906 but in principle I agree

pint/compat/__init__.py
jthielen: Sounds good! I agree that would be annoying. Which of the two (warn on first `__array_function__` or warn on first creation) would make the most sense here? 

Also, would the following (for example in the warn-on-first-creation case) be a good replacement for the suggested way to hide the warning:

```python
import numpy as np
import warnings
from pint import Quantity

with warnings.catch_warnings():
    q = Quantity([1, 2, 3, 4], 'm')
```

or would something else be better?

hgrecco:
As you sure have seen I have link a lot of issues to this PR. I am trying to make a list of numpy related issues that we should check again after this PR. Many of them will not be solved, but others might. The goal is to end 2019 with less open issues and PR!

jthielen:
I've finally reached the point where all tests are passing locally with NumPy available, so I think that means @keewis we can move forward with the next phase of testing with xarray!

Also, I've updated the initial comment with the main areas I need to work on before I'd feel ready to remove WIP/draft status.

keewis:
I opened a [new issue](https://github.com/pydata/xarray/issues/3594) where I listed the test failures. Most of these are probably issues with the code of xarray, but there are also a few missing function implementations: `nanpercentile`, `copyto`, `einsum`, `all`, `any`, `median`, `prod`, `sum`, `rank`, `isin`. Also `fabs`, `maximum` (and `minimum`?), but these are only used by the tests.

keewis:
another method that currently falls back to the magnitude and causes lots of unexpected stripped units is `astype`.

jthielen:
For the requested function implementations:

- Have been added:
  - `nanpercentile`
  - `copyto`
  - `isin`
  - `fabs`
  - `maximum`
  - `minimum`
- Should have already been there:
  - `median`
  - `sum`
- Need clarification:
  - `rank`: this is deprecated, so I don't think we should add it to new code?
- Can we leave as xfail for now to not hold up progress here?
  - `einsum`: determining the output unit for an arbitrary combination of sums and products from Einstein summation seems rather nontrivial
  - `all`, `any`: see #866 
  - `prod`: see https://github.com/hgrecco/pint/pull/764#issuecomment-523749008 / #867 

Also, the `astype` and `item` methods have been added.

keewis:
Thanks.

> Should have already been there:

yes, that's my fault, I simply checked the test status for the aggregation tests. `sum` and `median` work properly with `int`, but not `float`.

> `rank`: this is deprecated, so I don't think we should add it to new code?

definitely. The warning never shows for me (or gets drowned in the flood of `UnitStrippedWarning`s), though, so I didn't realize.

keewis:
somehow
```python
a = pint.Quantity([1, 3, 5], 'm')
a.ndim
```
raises a `AttributeError: NumPy method ndim was not callable.`, but `ndim` is a property (there is, however, a function named `numpy.ndim`)

pint/quantity.py
```diff
@@ -1442,48 +1437,32 @@ def __ito_if_needed(self, to_units):
...
-                self._magnitude = _to_magnitude(self._magnitude, force_ndarray=True)
-                return getattr(self._magnitude, item)
-        elif item in self.__handled:
-            if not isinstance(self._magnitude, ndarray):
-                self._magnitude = _to_magnitude(self._magnitude, True)
-            attr = getattr(self._magnitude, item)
+                # TODO (#905 follow-up): Potentially problematic, investigate for duck arrays
+                magnitude_as_array = _to_magnitude(self._magnitude, force_ndarray=True)
+                return getattr(magnitude_as_array, item)
+        elif item in HANDLED_UFUNCS or self._wrapped_numpy_methods:
```
keewis: I believe this should be
```python
        elif item in HANDLED_UFUNCS or item in self._wrapped_numpy_methods:
```
which would fix my problem with `ndim`

jthielen: Yep, that is what I had intended. I had a commit incoming, but I thought I'd try bundling it with any other fixes after testing it with your test branch at https://github.com/pydata/xarray/pull/3600. However, I'm still a bit overwhelmed with trying to figure out where all the errors and `__array_interface__`/`__array_struct__` calls are coming from.

jthielen:
Based on running through the xarray integration tests (as cleaned up in https://github.com/pydata/xarray/pull/3600), it looks like the following have issues that would be best resolved here:

- `maximum`
- `isin`
- `where`
- `sum` (in dataset aggregation test)

Also, there are many `UnitStrippedWarnings` raised in the xarray tests for the [`__array_interface__`](https://docs.scipy.org/doc/numpy/reference/arrays.interface.html#__array_interface__) and [`__array_struct__`](https://docs.scipy.org/doc/numpy/reference/arrays.interface.html#c.__array_struct__) attributes. What should the expected behavior from Pint be for these attributes/methods? The current behavior (call on magnitude and warn about stripped units), just call on magnitude without warning, or not implementing the attributes at all?

keewis:
I looked at these issues and have a few comments / design questions:

the `maximum` issue is a problem with how xarray dispatches ufuncs (passing a data array as a second value works just fine), let's discuss that in pydata/xarray#3594.

how should `isin` work? Should it raise on incompatible units, or should it return `False`? I'd vote for not raising, since the different unit makes it a different value.

For `float` arrays, xarray's implementation of `where` uses `np.nan` to mask arrays. Same with xarray's nan-version of `sum`, which uses `where` to replace `nan` with `0`. Should `np.nan` be converted by pint's implementation of `where`, or should xarray be responsible to convert the default masking values? If xarray should change its code, what is the best way to convert without introducing special-cases for pint? `type(x)(np.nan)` creates a `dimensionless` quantity.

jthielen:
@keewis Thanks for the input on those points! I've reimplemented `isin` based on your suggestion. Also, I think 0 and `np.nan` merit special behavior in Pint's implementation of `where`, for both xarray's purposes and general convenience for end users. I've re-implemented it accordingly, so hopefully that helps.

Also, `einsum` and a few other more functions have been added.

With this, I think all of xarray's needs are either met or intentionally deferred (e.g. `prod`/`any`/`all`)? If so, I think all that is left before removing draft status on this PR is further documenting and unit testing `numpy_func.py`, and resolving the [warning](https://github.com/hgrecco/pint/pull/905#discussion_r353427268) and [`__array_stuct__`/`__array_interface__`](https://github.com/hgrecco/pint/pull/905#issuecomment-562351799) discussions.

keewis:
looks good to me. I didn't try `rot90` and `insert`, but `einsum` and `isin` work really well, with the exception of mixed args for `isin`: `np.isin(quantity, np.arange(5))` returns something different from the result of `np.isin(quantity, np.arange(5) * ureg.dimensionless)`. Is that intentional?

Also, `xarray` chose to use 0D arrays instead of scalars (so it passes `np.array(np.nan)` and `np.array(0)` to where). Should we consider these as scalars, too?

keewis:
I just noticed we implement `__matmul__`, but not `Quantity.dot`, which will fall back to the ndarray method. Can we add that, too?

jthielen:
> I didn't try `rot90` and `insert`, but `einsum` and `isin` work really well, with the exception of mixed args for `isin`: `np.isin(quantity, np.arange(5))` returns something different from the result of `np.isin(quantity, np.arange(5) * ureg.dimensionless)`. Is that intentional?

No, thanks for catching that, as I had forgotten the final branch of the conditional. It should be updated now.

> Also, `xarray` uses 0D arrays instead of scalars (so it passes `np.array(np.nan)` and `np.array(0)` to where). Should we consider these as scalars, too?

This shouldn't be a problem (I just changed the condition to testing for non-iterable non-quantity instead of scalar)

> I just noticed we implement `__matmul__`, but not `Quantity.dot`, which will fall back to the ndarray method. Can we add that, too?

Just added!

hgrecco:
It would be good to summarize in the docs what has changed in relation to numpy support.

jthielen:
Documentation and units tests have been added for `numpy_func.py`, and a summary of the changes in relation to NumPy support have been added to the docs. So, I think this is ready for a full review (bearing in mind the 2 unresolved conversations on warning placement and `__array_struct__`/`__array_interface__`), finally!

hgrecco:
Awesome! I would like to merge this on wednesday. So I will kindly ask everyone who was involved to chime in by then.

hgrecco:
By the way, on tueday it will be 7 years. 7 months and 7 days since I started working on Pint

keewis:
I think we can leave the `__array_struct__` / `__array_interface__` issues for a new PR and until we find out what exactly causes them.

hgrecco:
> I think we can leave the `__array_struct__` / `__array_interface__` issues for a new PR and until we find out what exactly causes them.

I agree.

pint/compat/__init__.py
keewis: did we resolve this, yet? If not, I'd vote for raising on first creation since that would make sure people know that what they just created behaves different from before.

To silence it, I think the sample above works. However, I wouldn't assign a name to it (since that can make linters like `flake8` complain about a name being assigned but not used). We could also shorten it to `Quantity([], 'm')` or use `Quantity(np.array([]), "m")`?

jthielen: Since it seems like we have enough of a consensus, I've just pushed a commit to change the warning to on-first-creation.

keewis:
did we resolve #882 yet?

jthielen:
> did we resolve #882 yet?

No, not yet. I've implemented it here for now as in the first option ("Have `zeros_like`, `ones_like`, and `empty_like` return base zeros, ones, or empty ndarrays, respectively, and have `full_like` simply match the unit of the `fill_value` argument"), but that may be changed by how #882 is resolved (I'd personally still argue for this option, though, since it is more straightforward in interpretation and implementation).

keewis:
okay, sounds good to me. I just wanted to make sure it's not forgotten.

@hgrecco, what's your opinion on this? Should we also close #882 with the merge, or do we need more discussion?

hgrecco:
@keewis I do not have a strong opinion on this. I was inclined, as @jthielen said, the the units is part of the data (like dtype). So `*_like` should copy everything. But In any case, I would argue to  merge. I think we need to see this in action, close all resolved issues to see the pending ladscape and review the decision if needed.

keewis:
I'm not sure if I understand you correctly: are you content with the current state or not?
In any case, +1 for merging and reviewing later.

hgrecco:
In my opinon, the `*_like` should copy everything (including the units). But I have no other rationale than my gut feeling for now. In those cases, my motto is "let the person doing the hard work decide, and test if it feels right afterwards". So I will the implemented suggestion a try. I think what is right or wrong in these cases emerges from usage in real codebases.

docs/numpy.rst
```diff
@@ -133,22 +137,31 @@ will be raised.
...
 usage. Quantities that must be converted to other units require additional
-memory and CPU cycles. On top of this, all `ufuncs` are implemented in the
-`Quantity` class by overriding `__array_wrap__`, a NumPy hook that is executed
-after the calculation and before returning the value. To our knowledge, there
-is no way to signal back to NumPy that our code will take care of the
-calculation. For this reason the calculation is actually done twice:
-first in the original ndarray and then in the one that has been
-converted to the right units. Therefore, for numerically intensive code, you
+memory and CPU cycles. Therefore, for numerically intensive code, you
 might want to convert the objects first and then use directly the magnitude.
```
hgrecco: Make a list to `wraps` here.

jthielen: A link has been added!

hgrecco:
I done an initial review and it looks good so far. I still have to look at the tests in more detail. Please

hgrecco:
I review the rest. Looks good to me. If the pending review is solved and no one raises an issue, I will merge it tomorrow.

hgrecco:
@jthielen Sorry but I just realized I had not "submitted" the review. They are all minor things.

pint/quantity.py
```diff
@@ -1302,62 +1315,56 @@ def __bool__(self):
...
-    __skip_other_args = 'ldexp multiply ' \
-                        'true_divide divide floor_divide fmod mod ' \
-                        'remainder'.split()
-
-    __handled = tuple(__same_units) + \
-                tuple(__require_units.keys()) + \
-                tuple(__prod_units.keys()) + \
-                tuple(__copy_units) + tuple(__skip_other_args)
+    # NumPy function/ufunc support
+    __array_priority__ = 17
```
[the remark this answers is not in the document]
jthielen: While [the NumPy docs](https://docs.scipy.org/doc/numpy/reference/arrays.classes.html#numpy.class.__array_priority__) imply that it is no longer used for ufuncs when `__array_ufunc__` is in use, Dask, Sparse, and CuPy still all seem to declare it, so I thought it best to maintain it here. 

17 is simply the value Pint had previously. This puts it above ndarray (10), Dask (11), Sparse (12), and masked arrays (15), which is all good. However, CuPy is unfortunately at 100, so this may need to be increased if Pint Quantities should be above CuPy arrays. Would you want to increment this value now, or leave it at the current value of 17 and discuss any possible changes in #924?

pint/numpy_func.py
```diff
@@ -0,0 +1,609 @@
...
+            and all(_is_quantity(item) for item in arg))
+
+
+def _get_first_input_units(args, kwargs={}):
+    """Obtain the first valid unit from a collection of args and kwargs."""
+    args_combo = list(args) + list(kwargs.values())
+    out_units=None
+    for arg in args_combo:
+        if _is_quantity(arg):
+            out_units = arg.units
```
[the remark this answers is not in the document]
jthielen: Thank you for catching these! They were mostly carry-over from #764, which I should have reviewed more closely. I'll be pushing a commit soon with the updates.

pint/quantity.py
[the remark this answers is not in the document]
hgrecco: Similar, to my previous comment. I think we should merge this ASAP to increase the base of tester. Regarding the number, I have two (not contradictory) options:
1. Kindly Suggest to CuPy to change the number
2. Get this value from an environmental variable (defaulting to 17). We should do this is in a way that is copyable by other packages (e.g. Naming the variable `PINT_NUMPY_ARRAY_PRIORITY` or similar)

hgrecco:
I have just asked bors to merge this PR. I would like to thank each and everyone who has contributed to this PR, either by code, bugs, ideas, documentation and testing. This is a major step forward in Pint and a great example of collaborative work and open source development. 

Please review many Numpy related issues still open after this PR has been merged, closing those which are fixed and noting what is the new status after this PR.

jthielen:
> Please review many Numpy related issues still open after this PR has been merged, closing those which are fixed and noting what is the new status after this PR.

I've gone through all the linked issues that remained open and either commented on their status or incorporated them into more general issues such as #845 and #925, so hopefully that helps!

PR #911, opened 2019-12-04
Implement registry-based string preprocessing as list of callables

### LINKED ISSUE #429: How to use sign %?
## Define Sign %

I define:
`from pint import UnitRegistry`
`ureg=UnitRegistry()`
`ureg.define('percent = 1 = %')`
`x = 20 * u('1 %')`
`print('x: {0}'.format(x))`
`print('x: {:L}'.format(x))`

and the output is:
`x: 20`
ValueError: Unknown format code 'L' for object of type 'int'

I need:
`x: 20%`
`x: 20\%`

### LINKED ISSUE #851: Possibility of support for CF/UDUNITS-style powers of units
The [CF Conventions](http://cfconventions.org/) are a common standard used in the atmospheric and oceanic sciences, and many NetCDF files follow these conventions. The [unit conventions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#units) rely upon the [UDUNITS package](https://www.unidata.ucar.edu/software/udunits/). While most CF/UDUNITS-compliant unit strings can be parsed properly by pint, I've run into some issues recently with [powers of units in the default symbol notation](https://www.unidata.ucar.edu/software/udunits/udunits-current/doc/udunits/udunits2lib.html#Unit-Specification-Examples), i.e., strings such as "m2 s-2" for meters squared per second squared.

Would support for powers of units in this style be welcome in pint?

If so, this should be fairly simple to implement by adding the following regex substitution pair

```python
(r'(?<=[A-Za-z])(?![A-Za-z])(?<![0-9\-])(?=[0-9\-])', '**')
```

to

https://github.com/hgrecco/pint/blob/2afdc4bf5c5727ed9cef6fdaccb00b88813c4a24/pint/util.py#L565-L575

If not, would it be possible to add something to the registry API in order to add regex substitution pairs for the string preprocessor? Even if pint itself doesn't wish to support this style of powers of units (which would be reasonable...this isn't the clearest syntax), having some officially supported way to add new regex substitutions would make it much easier for downstream libraries like MetPy to implement it if desired for CF/UDUNITS-compatibility.

Either way, I'd be glad to put in a PR to implement the desired functionality.

xref https://github.com/Unidata/MetPy/issues/1134

jthielen:
As suggested in #429, this implements simple registry-based unit/expression string preprocessing through a list of callables.

There are a couple questions I was wondering about:
- Is the added argument to the docstring sufficient documentation, or would additional documentation (such as an example with percent as in #429) be useful?
- Right now, the unit string preprocessing occurs in `parse_units`, rather than `_parse_units`, which avoids the cache. For performance reasons, would it be worth the extra complexity of cache invalidation to have this moved inside `_parse_units` after the cache check?

Closes #429
Closes #851

### DIFF
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -101,6 +101,8 @@ class BaseRegistry(meta.with_metaclass(_Meta)):
                             'warn', 'raise', 'ignore'
     :type on_redefinition: str
     :param auto_reduce_dimensions: If True, reduce dimensionality on appropriate operations.
+    :param preprocessors: list of callables which are iteratively ran on any input expression
+                          or unit string
     """
 
     #: Map context prefix to function
@@ -116,13 +118,15 @@ class BaseRegistry(meta.with_metaclass(_Meta)):
             'parse_unit_name', 'parse_units', 'parse_expression',
             'convert']
 
-    def __init__(self, filename='', force_ndarray=False, on_redefinition='warn', auto_reduce_dimensions=False):
+    def __init__(self, filename='', force_ndarray=False, on_redefinition='warn',
+                 auto_reduce_dimensions=False, preprocessors=None):
 
         self._register_parsers()
         self._init_dynamic_classes()
 
         self._filename = filename
         self.force_ndarray = force_ndarray
+        self.preprocessors = preprocessors or []
 
         #: Action to take in case a unit is redefined. 'warn', 'raise', 'ignore'
         self._on_redefinition = on_redefinition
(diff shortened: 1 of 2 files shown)


pint/registry.py
```diff
@@ -116,13 +118,15 @@ class BaseRegistry(meta.with_metaclass(_Meta)):
             'parse_unit_name', 'parse_units', 'parse_expression',
             'convert']
 
-    def __init__(self, filename='', force_ndarray=False, on_redefinition='warn', auto_reduce_dimensions=False):
+    def __init__(self, filename='', force_ndarray=False, on_redefinition='warn',
+                 auto_reduce_dimensions=False, preprocessors=[]):
```
hgrecco: It is usually not a good idea to use a list as the default value as it is mutable. Better use None and then `self.preprocessors = preprocessors or []`

hgrecco:
i am merging this now, but I think before launching the next version we need to revisit the docs and improve it.
Regarding the cache, let's see if we actually need it

PR #916, opened 2019-12-04
Make errors pickleable

czagoni:
Currently none of the errors raised by pint can be pickled. This is problematic in some cases, e.g. in a multiprocessing environment: if one of the child processes has an error raised by pint, it appears as a pickling error on the main thread and thus hides the true source of the problem.

This fix adds the errors' `__init__` method arguments as arguments to the `__init__` method call of the parent class, therefore making sure that the `Exception` base class's `__reduce__` method works as expected.


czagoni:
Whoops - I've just noticed there is another PR out for this (and more)

crusaderky:
indeed - please retire this PR

PR #909, opened 2019-12-03
Preserve caches on context switch

crusaderky:
Closes #795 
Forked from #801

PR #923, opened 2019-12-09
Preserve caches on context switch v2

crusaderky:
Fix bug in #909 where, every time a context is reactivated, it creates a brand new entry in UnitRegistry._caches - thus not only nullifying the whole point of the PR, but also creating a nasty memory leak for users that invoke ``.to`` with an inline context many times.

Not sure what went wrong with the first PR - if I rerun the included benchmark, it doesn't reproduce the same results anymore and clearly shows that nothing is cached.

Same benchmark with this PR:
```python
import pint
ureg = pint.UnitRegistry()
src = ureg('1 m')
dst = ureg.Unit('km')
%timeit src.to(dst)
%timeit -n 1 -r 1 src.to(dst, 'chemistry')
%timeit src.to(dst, 'chemistry')
ureg.enable_contexts('chemistry')
%timeit src.to(dst)
```
29.9 µs ± 1.4 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
57.3 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
76.9 µs ± 1.69 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
43.7 µs ± 955 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -15,7 +15,7 @@ Pint Changelog
 - TODO #913
 - TODO #911
 - Prevent 1500x slowdown when using .to() with a context
-  (Issue #909, Thanks Guido Imperiale)
+  (Issues #909 and #923, Thanks Guido Imperiale)
 - **BREAKING CHANGE**:
   Drop support for Python < 3.6, numpy < 1.14, and uncertainties < 3.0;
   if you still need them, please install pint 0.9.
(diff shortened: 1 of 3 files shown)


crusaderky:

PR #926, opened 2019-12-12
use packaging.version.parse instead of StrictVersion for version checking

### LINKED ISSUE #872: Pint doesn't declare `setuptools` as a dependency.
Pint imports `pkg_resources` (part of `setuptools` AFAIK), but doesn't declare `setuptools` as a dependency.

This means that bundling `pint` in the normal way (i.e. `pip install --target .`) doesn't work properly, as its dependencies aren't included.

keewis:
fixes #872

`distutils.version.StrictVersion` cannot parse numpy development versions like `1.19.0.dev0+578f4e7`. `LooseVersion` is able to parse the version, but it compares
```python
>>> LooseVersion("1.19.0.dev0+578f4e7") < LooseVersion("1.19")
False
```
Using `pkg_resources.extern.packaging.version.parse` instead makes this work as expected. Since `setup.py` requires `setuptools`, I think it being installed is a reasonable assumption?

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -19,7 +19,7 @@ import operator
 import re
 import warnings
 
-from distutils.version import StrictVersion
+from pkg_resources.extern.packaging.version import LegacyVersion
 
 from .formatting import (remove_custom_flags, siunitx_format_unit, ndarray_to_latex,
                          ndarray_to_latex_parts)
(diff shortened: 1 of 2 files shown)


hgrecco:
See also #872, maybe we can fix both things at the same time.

keewis:
it's now in `install_requires`. Is that the right place to put it?

hgrecco:
> it's now in `install_requires`. Is that the right place to put it?

I think so

PR #929, opened 2019-12-13
black, flake8, and isort (part 1)

crusaderky:
I'd like to introduce black, flake8, and isort to pint - exactly in the same way they're already used (very successfully) in xarray.

I'm breaking the PR in two to make it easier to review it and to merge it into other branches:

1) (this PR): I just run ``black -t py36 .``. No manual changes whatsoever. After it's merged, everybody who has branches open will need to run the same command on their own branch before merging - or face a lot of merge conflicts.

2) (#931, to be reviewed after this one is merged) manual changes: configure flake8 and isort; integrate flake8, isort, and black in CI; fix flake8 failures. Reviewing and merging into other branches should be uncomplicated.


crusaderky:
@hgrecco ready for (optional) discussion and merge

crusaderky:
@hgrecco conflicts with master solved

hgrecco:
I am +1 about introducing an automatic formatting tool (that is the one thing I liked the most about go). And I like black style ... except for using double quotes instead of single quotes for strings. I dislike double quotes (except in docstrings). I dislike them a lot. I am not sure if it is because of visual polution, or because a single quote does not requires to press the shift key!. 

I understand that this is a very personal thing and therefore I am ready to let it go.

crusaderky:
@hgrecco I understand your point - I too was of the same opinion before I started to use black.
After a few months using it, I agree with the arguments of the black creators:

1. apostrophes are very pervasive in English, and they're the same character as single-quote. So if you were to use single-quote as a standard to wrap strings, when a string contains an apostrophe you would be forced to either
a. use double quotes for that string only
b. escape the apostrophe
and both options are very ugly.
Double quotes, on the other hand, are extremely rare inside Python strings. The only case where they may be used is to wrap the value of a variable when you print it in a log or exception - for which you can just use single quote.

2. personal style preference is always an inferior argument to coherency across projects, particularly when there isn't a clear 'right' or 'wrong' choice. If you don't like one or more of the black choices, your should put the argument on either the black board or as an amendment of PEP8, and then everybody should submit to the concerted decision of the wider community.

crusaderky:
> a single quote does not requires to press the shift key!.

I always type a single quote and then let black fix it for me before I commit ;)

hgrecco:
> apostrophes are very pervasive in English, and they're the same character as single-quote. So if you were to use single-quote as a standard to wrap strings, when a string contains an apostrophe you would be forced to either

I agree but I almost never use an apostrophes, except in docstring. Thereby my rule: single quotes for everything except docstrings. Which is similar to single quotes for data, doble quotes for text.

But anyway, as I said before, lets go for it.

PR #930, opened 2019-12-13
Auto-increase dev version

crusaderky:
A static version "0.10.dev0" is troublesome for developers who need to use the git tip of pint in their project. 

Namely, if one uses this conda requirements file:
```
name: test
channels:
  - defaults
dependencies:
  - python =3.7
  - pip
  - pip:
    - git+https://github.com/hgrecco/pint.git
```
it has been observed to cause pip to accidentally install an obsolete version of the package from the local cache. This is not a problem on cloud CI (where the environment is properly wiped clean every time) but it impacts Jenkins and personal dev boxes.

Also, a subsequent ``conda list`` command, as is typical in most CI scripts for the purpose of forensic analysis, will print a non-helpful '0.10.dev0'.

This PR replaces the static 'dev0' with an ever-increasing number fetched from git. Ahead of a release, whereas the maintainer before pushed a commit that changed the version to remove the .dev0 suffix, will now have to set the RELEASE_VERSION flag to true.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,9 @@ Pint Changelog
 0.10 (unreleased)
 -----------------
 
+- Auto-increase package version at every commit when pint is installed from the git tip,
+  e.g. ``pip install git+https://github.com/hgrecco/pint.git``.
+  (Issue #930, Thanks Guido Imperiale)
 - Fix HTML (Jupyter Notebook) and LateX representation of some units
   (Issue #927, Thanks Guido Imperiale)
 - **BREAKING CHANGE**:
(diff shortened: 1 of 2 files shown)


crusaderky:

KOLANICH:
You shouldn't have solved the issue this way. `setuptools_scm` should have been used. A single addition to setup command.

crusaderky:
@KOLANICH you're welcome to push a PR with a better solution...

KOLANICH:
https://github.com/hgrecco/pint/pull/934

PR #934, opened 2019-12-14
Use setuptools_scm to get version from git instead of calling subprocess on own.

KOLANICH:


crusaderky:
```
$ pip install git+https://github.com/KOLANICH/pint@patch-1
[...]
Pint-0.8.2.dev404+gf8e561a
```

Something is not working; the latest git tag is 0.9.

KOLANICH:
Is it the last tag on the current branch? I mean is this commit a successor of the commit with the tag?

crusaderky:
@KOLANICH, you have access to the git repo as much as I do. 
0.9 is the latest tag on the master branch, from which you forked.

setup.py
```diff
@@ -66,7 +45,7 @@ def read(filename):
         'Programming Language :: Python :: 3.8',
     ],
     python_requires='>=3.6',
-    install_requires=['setuptools'],
```
keewis: if I read the [documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html) on `setup_requires` correctly, we need `install_requires` **and** `setup_requires`: `pint` depends on `pkg_resources`.

hgrecco:
I am merging this to move forward. Then we 'black' the quotes. (I should have said we "black" the quotes, sorry :-P)

crusaderky:
Weird, now the issue with setuptools-scm not seeing the 0.9 tag was caused by not being on the master branch for some reason. Now it works.
```
Successfully installed Pint-0.10.dev272+g3936c36
pip install git+https://github.com/hgrecco/pint
```

KOLANICH:
In fact my clone lacked the `0.9` tag because I usually use shallow fetches.

PR #938, opened 2019-12-18
Use contexts to redefine units

### LINKED ISSUE #853: Context.define for currency conversions
I'm trying to use pint to perform currency conversions.
Currency FX rates change by the minute and may be different in different markets/situations - which would be modelled perfectly by a pint.Context.

However, while Context allows defining custom conversion functions between different dimensions, it doesn't seem to allow customizing conversions within the same dimension.

I'm currently using throwaway UnitRegistry objects (very expensive to create):
```python
import pint

ureg_day1 = pint.UnitRegistry()
ureg_day1.define("USD = [currency]")
ureg_day1.define("EUR = 1.11 USD")
ureg_day1.define("GBP = 1.22 USD")

ureg_day2 = pint.UnitRegistry()
ureg_day2.define("USD = [currency]")
ureg_day2.define("EUR = 1.13 USD")
ureg_day2.define("GBP = 1.28 USD")

print((100 * ureg_day1("EUR")).to(ureg_day1("GBP")))
print((100 * ureg_day2("EUR")).to(ureg_day2("GBP")))
```
output:
```
90.98360655737706 GBP
88.28124999999999 GBP
```

I could not find a way to convert 100 EUR to USD on day1 and then convert them back to EUR on day2. This doesn't work:
```python
x = 100 * ureg_day1.EUR
x = x.to(ureg_day1.USD)
x = x.to(ureg_day2.EUR)
x
```
Output:
```
100.0 EUR  # Wrong - it should be 98.23 EUR!
```
Workaround:
```python
x = 100 * ureg_day1.EUR
x = x.to(ureg_day1.USD)
x = x.magnitude * ureg_day2.USD
x = x.to(ureg_day2.EUR)
x
```
Output:
```
98.23008849557525 EUR
```

It would make sense to have instead something like:
```python
import pint

ureg = pint.UnitRegistry()
ureg.define("USD = [currency]")
ureg.define("EUR = <TODO: valid currency, but crash if you attempt conversion>")
ureg.define("GBP = <TODO: valid currency, but crash if you attempt conversion>")

ctx_day1 = pint.Context("")
ctx_day1.define("EUR = 1.11 USD")
ctx_day1.define("GBP = 1.22 USD")

ctx_day2 = pint.Context("")
ctx_day2.define("EUR = 1.13 USD")
ctx_day2.define("GBP = 1.28 USD")

x = 100 * ureg.EUR
print(x.to(ureg.GBP, ctx_day1))
print(x.to(ureg.GBP, ctx_day2))

# x = x.to(ureg.USD)  # crash
x = x.to(ureg.USD, ctx_day1)
x = x.to(ureg.EUR, ctx_day2)
print(x)
```
Output:
```
90.98360655737706 GBP
88.28124999999999 GBP
98.23008849557525 EUR
```

### LINKED ISSUE #932: Disable all throwaway contexts
Anonymous contexts (#870) are enabled on the fly with ``ureg.enable_contexts()``, but are not previously registered with ``ureg.add_context()``. Regardless of having a name or not, it is convenient in some cases to use throwaway contexts.

When there are more throwaway contexts enabled than the number of contexts permanently registered with ``ureg.add_context()``,  then ``ureg.disable_contexts()``  invoked without parameter fails to disable them all.

```python
>>> import pint
>>> len(ureg._active_ctx.maps)
1  # Empty context - this is a separate bug
>>> ureg.enable_contexts(*(pint.Context() for _ in range(len(ureg._contexts) + 3)))
>>> len(ureg._active_ctx.maps)
15
>>> ureg.disable_contexts()
>>> len(ureg._active_ctx.maps)
4
>>> ureg.disable_contexts()
>>> len(ureg._active_ctx.maps)
0
```
The context manager syntax is unaffected:
```python
>>> with ureg.context(*(pint.Context() for _ in range(len(ureg._contexts) + 3))):
...     pass
>>> len(ureg._active_ctx.maps)
0
```

crusaderky:
- [x] Closes #853 - Context.define for currency conversions
- [x] Closes #932 - Disable all throwaway contexts
- [x] Contexts that only implement transforms between dimensionalities (that is, all contexts until now) no longer require their own cache and therefore are enabled in microseconds even on the first time; contexts that implement redefinitions require a *much* smaller cache than before which is very quickly built on demand.
- [x] NaN is now parsed as a float

# Demo
```python
import pint

ureg = pint.UnitRegistry()
ureg.load_definitions("""
USD = [currency]
EUR = NaN USD
GBP = NaN USD

@context currencies_20191218
    EUR = 1.11 USD
    GBP = 1.18 EUR
@end

@context dungeons_and_dragons = dnd
    # Simplified conversion rates for the
    # non-English versions of the D&D manuals
    pound = 0.5 kg
    gallon = 4 l
    foot = 1.5/5 m
@end
""".splitlines())

>>> ureg.GBP.dimensionality
<UnitsContainer({'[currency]': 1.0})>
>>> ureg.Quantity("10 GBP")
10 GBP
>>> ureg.Quantity("10 GBP").to("USD")
nan USD
>>> ureg.Quantity("10 GBP").to("USD", "currencies_20191218")
13.098 USD
>>> ureg.Quantity("10 ft").to("m")
3.0479999999999996 meter
>>> ureg.Quantity("10 ft").to("m", "dnd")
3.0 meter
```

### DIFF
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -754,7 +754,7 @@ class BaseRegistry(metaclass=RegistryMeta):
         return self.get_root_units(input_units, check_nonmult)
 
     def _get_root_units_recurse(self, ref, exp, accumulators):
-        for key in sorted(ref):
+        for key in ref:
             exp2 = exp * ref[key]
             key = self.get_name(key)
             reg = self._units[key]
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -527,11 +527,10 @@ class BaseRegistry(metaclass=RegistryMeta):
                 if "[" in unit_name:
                     continue
                 parsed_names = self.parse_unit_name(unit_name)
-                prefix = None
                 if parsed_names:
-                    prefix, base_name, _suffix = parsed_names[0]
+                    prefix, base_name, _ = parsed_names[0]
                 else:
-                    base_name = unit_name
+                    prefix, base_name = '', unit_name
 
                 try:
                     uc = ParserHelper.from_word(base_name)
@@ -897,7 +896,8 @@ class BaseRegistry(metaclass=RegistryMeta):
         for cp, cu, cs in list(candidates):
             assert isinstance(cp, str)
             assert isinstance(cu, str)
-            assert cs == "", "not empty suffix is not supported"
+            if cs != "":
+                raise NotImplementedError("non-empty suffix")
             if cp:
                 candidates.pop(("", cp + cu, ""), None)
         return tuple(candidates)
@@ -1486,7 +1486,7 @@ class ContextRegistry(BaseRegistry):
         ret = super()._get_compatible_units(input_units, group_or_system)
 
         if self._active_ctx:
-            ret = ret.copy()
+            ret = ret.copy()  # Do not alter self._cache
             nodes = find_connected_nodes(self._active_ctx.graph, src_dim)
             if nodes:
                 for node in nodes:
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -530,7 +530,7 @@ class BaseRegistry(metaclass=RegistryMeta):
                 if parsed_names:
                     prefix, base_name, _ = parsed_names[0]
                 else:
-                    prefix, base_name = '', unit_name
+                    prefix, base_name = "", unit_name
 
                 try:
                     uc = ParserHelper.from_word(base_name)
@@ -1296,9 +1296,7 @@ class ContextRegistry(BaseRegistry):
                 f"candidates: {candidates_no_prefix}"
             )
         if not candidates_no_prefix:
-            raise ValueError(
-                f"Can't redefine a unit with a prefix {definition.name}"
-            )
+            raise ValueError(f"Can't redefine a unit with a prefix {definition.name}")
         _, name, _ = candidates_no_prefix[0]
         try:
             basedef = self._units[name]
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -1303,6 +1303,15 @@ class ContextRegistry(BaseRegistry):
         # Rebuild definition as a variant of the base
         if basedef.is_base:
             raise ValueError("Can't redefine a base unit to a derived one")
+
+        dims_old = self._get_dimensionality(basedef.reference)
+        dims_new = self._get_dimensionality(definition.reference)
+        if dims_old != dims_new:
+            raise ValueError(
+                f"Can't change dimensionality of {basedef.name} "
+                f"from {dims_old} to {dims_new} in a context"
+            )
+
         # Do not modify in place the original definition, as (1) the context may
         # be shared by other registries, and (2) it would alter the cache key
         definition = UnitDefinition(
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -717,13 +717,11 @@ class BaseRegistry(metaclass=RegistryMeta):
         if not input_units:
             return 1.0, UnitsContainer()
 
-        # The cache is only done for check_nonmult=True
         cache = self._cache.root_units
-        if check_nonmult:
-            try:
-                return cache[input_units]
-            except KeyError:
-                pass
+        try:
+            return cache[input_units]
+        except KeyError:
+            pass
 
         accumulators = [1.0, defaultdict(float)]
         self._get_root_units_recurse(input_units, 1.0, accumulators)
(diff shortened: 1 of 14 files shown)


pint/registry.py
```diff
@@ -768,7 +752,7 @@ def get_base_units(self, input_units, check_nonmult=True, system=None):
         return self.get_root_units(input_units, check_nonmult)
 
     def _get_root_units_recurse(self, ref, exp, accumulators):
-        for key in sorted(ref):
+        for key in ref:
```
crusaderky: Couldn't find any reason for sorting?

pint/registry.py
```diff
@@ -887,6 +883,25 @@ def parse_unit_name(self, unit_name, case_sensitive=True):
...
+        candidates = dict.fromkeys(candidates)  # ordered set
+        for cp, cu, cs in list(candidates):
+            assert isinstance(cp, str)
+            assert isinstance(cu, str)
+            if cs != "":
+                raise NotImplementedError("non-empty suffix")
+            if cp:
+                candidates.pop(("", cp + cu, ""), None)
+        return tuple(candidates)
+
```
crusaderky: Worth noting that this function was not doing what its docstring said before; it was deduplicating only identical tuples.

pint/registry.py
```diff
@@ -1400,6 +1485,7 @@ def _get_compatible_units(self, input_units, group_or_system):
         ret = super()._get_compatible_units(input_units, group_or_system)
 
         if self._active_ctx:
+            ret = ret.copy()  # Do not alter self._cache
```
crusaderky: This line missing was the one and only reason why contexts required their own cache before!

crusaderky:
@jetheurer could you play with this and tell me how it goes?

crusaderky:
Changed undef to NaN. The user will have to implement his own ``math.isnan`` tests.
Couldn't find a simple way to tell apart internal transformation performed by pint from the final operations performed by the user.

crusaderky:
@hgrecco ready for review and merge :sunglasses:

hgrecco:
Great PR! It adds functionality and improves the existing one.

Just as a closing note: Now that unit definitions are possible within contexts, using currencies is possible. However, I would like to keep pint within the physical units realm. More involved currency conversion features such as downloading values from the web for certain dates, etc would be welcomed as an external package.

crusaderky:
Downloading currency fx rates from the Internet was never in my plans. Pandas-datareader already does it well enough for any prototype work, and writing a function that generates a pint context out of its dataframes is trivial. For production work, each company will have its own bespoke data source, so any integration has to be bespoke as well.

jetheurer:
I love it!  Thanks @crusaderky

numpde:
This looks quite useful, thank you. I'd like to step it up a notch and have the same currency on different dates (or even timestamps) behave essentially like different units (because of the distinction of 'nominal' and 'real' value). Is there a slicker way than defining a bunch of units like EUR_20200515, EUR_20200516, USD_20200515, ...?

crusaderky:
@numpde no there isn't. May I ask why you want to do that? I'm using pint for a trading platform and I am pretty satisfied with modelling FX with one context per timepoint (day or intraday). And I can use transforms to implement asymmetric curves, FX commissions, etc.

numpde:
@crusaderky In principle, currency denominations across time shouldn't mix for the same reason that different currencies don't: their value is nominal. In the long term this is usually because of inflation. In financial projections, cash flows at different timepoints are discounted to a reference timepoint (e.g. the present) before adding them up to capture the [time value of money](https://www.investopedia.com/terms/t/timevalueofmoney.asp). So, while we're all "[consenting adults](https://pint.readthedocs.io/en/latest/defining.html)", I'm looking for a currency unit framework/library that would encourage this distinction (and would also digest currency conversion at time-variable rates, and interoperate with pandas/numpy).

crusaderky:
@numpde I'm well aware of how discounting and Mark-to-Market pricing works. Still don't get why the current design (one context per timestamp) wouldn't allow you to model it.

It's worth noting that you can pass a Quantity that wraps a numpy vector to a transform; this could allow you to model a time axis and/or deterministic or Monte Carlo shocks on your FX rates. Sadly labelling with xarray is fidgety today - see discussion at https://github.com/hgrecco/pint/issues/1099

PR #940, opened 2019-12-19
Reimplement cumulative product NumPy functions

### LINKED ISSUE #939: array function implementation for np.cumprod / np.nancumprod fails with axis
Passing a `axis` argument to `cumprod` / `nancumprod` will make them fail. I tracked it down to https://github.com/hgrecco/pint/blob/4c3114c72bdae1e3c5c9f839659169eeb06071cd/pint/numpy_func.py#L62 where `pre_calc_units` can be a string (`"dimensionless"`), but a `Unit` object was expected. The string was created when `implement_func` was called for `cumprod` and `nancumprod` (by the way, `cumprod` also gets implemented with `set_units_ufuncs`). We can fix this by passing `None` as `input_units` instead, but if feasible I'd prefer `"dimensionless"` since that is way easier to read and understand. For that, we might need to preprocess the `input_units` parameter in `implement_func`.

jthielen:
This PR resolves two issues from #905 that were identified in #939:

First, `convert_arg` in `numpy_func.py` presumes `pre_calc_units` is a `Unit` or `None`, but there are instances where it was passed a string by way of `inputs_units` in `implement_func`. Now, any `inputs_units` as a string are converted to a `Unit` before being passed to `convert_to_consistent_units`.

Second, the cumulative product functions were using `implement_func` with `input_units` set to something other than `None`, which was incorrect, since these functions have arguments that should not have units (namely `axis`). These functions have been reimplemented accordingly.

- [x] Closes #939 
- ~~Documented in docs/ as appropriate~~
- ~~Added an entry to the CHANGES file~~ (fixup to previous change: #905)

### DIFF
--- a/pint/numpy_func.py
+++ b/pint/numpy_func.py
@@ -682,17 +682,31 @@ def implement_atleast_nd(func_str):
 for func_str in ["atleast_1d", "atleast_2d", "atleast_3d"]:
     implement_atleast_nd(func_str)
 
+
+# Handle cumulative products (which must be dimensionless for consistent units across
+# output array)
+def implement_single_dimensionless_argument_func(func_str):
+    # If NumPy is not available, do not attempt implement that which does not exist
+    if np is None:
+        return
+
+    func = getattr(np, func_str)
+
+    @implements(func_str, "function")
+    def implementation(a, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


hgrecco:
Could you document the possible values for `inputs_units`?

PR #941, opened 2019-12-19
Fix infinite recursion with NumPy array/scalar raised to quantity power

jthielen:
After running the current master branch against MetPy's test suite, I found that #905 introduced an infinite recursion error when the base of the exponent was a NumPy array or scalar by naively deferring to standard exponentiation. There were also issues when both the base and power were Quantities. This PR resolves those issues and adds basic tests for those cases.

- ~~Closes~~ (no associated issue)
- ~~Documented in docs/ as appropriate~~
- ~~Added an entry to the CHANGES file~~ (fixup to previous change: #905)

### DIFF
--- a/pint/numpy_func.py
+++ b/pint/numpy_func.py
@@ -396,7 +396,10 @@ def _frexp(x, *args, **kwargs):
 
 @implements("power", "ufunc")
 def _power(x1, x2):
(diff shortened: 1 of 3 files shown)


hgrecco:
Thanks for catching this

Issue #904, opened 2019-11-26
Add option to set locale on UnitRegistry; auto-Babelise (resp. Babelize!)


OJFord:
It would be good to be able to set, for example, `locale='en_GB'` on instantiating a `UnitRegistry`, and then have it automatically `format_babel` to that locale.

As it is, the following is a little odd:

```
>>> ureg.litre
<Unit('liter')>
>>> str(3 * ureg.litre)
'3 liter'
```

it's good that both spellings are accepted in Python, but it's a bit odd to then have it come back out in AmE having explicitly used BrE.

Having `ureg = UnitRegistry(locale='en_GB')` could mitigate that, and then also allow `'litre'` in the response of `ureg.cc.compatible_units()`.

I could help with this, if it is something you agree with. Cheers.

hgrecco:
I like this. My suggestions (which are of course open for discussion):
- add a method (`set_locale`) to the following classes: Unit/System/Quantity/BaseRegistry
  This method should take:
  - None to have no locale  (i.e. no transformation).
  - a string with the locale name.
  - 'auto' to detect the current locale
- `set_locale` in BaseRegistry should call the ones Unit/System/Quantity
- add the `locale` argument to `BaseRegistry`, and call the `set_locale` from there.

Alternatively:
- Create a new class in `registry.py` named LocalizedRegistry with the `locale` modeled like the other registry classes that extend functionality.

Can you submit a PR?

hgrecco:
In addition, a better error message to be provided when babel is not installed would be nice.

hgrecco:
Closed by #944

PR #944, opened 2019-12-21
Improved babel experience

### LINKED ISSUE #899: calling format_babel without babel installed causes an AttributeError
[babel](https://pypi.org/project/Babel/) is not listed as a dependency for pint, however there are cases where not having it installed can lead to errors at runtime.
```
self = <Quantity(200, 'micrometer')>, spec = '~', kwspec = {'locale': 'en_US'}, obj = <Quantity(200, 'micrometer')>

    def format_babel(self, spec='', **kwspec):
        spec = spec or self.default_format
    
        # standard cases
        if '#' in spec:
            spec = spec.replace('#', '')
            obj = self.to_compact()
        else:
            obj = self
        kwspec = dict(kwspec)
        if 'length' in kwspec:
            kwspec['babel_length'] = kwspec.pop('length')
>       kwspec['locale'] = Loc.parse(kwspec['locale'])
E       AttributeError: 'NoneType' object has no attribute 'parse'
```
The line causing the issue is in quantity
https://github.com/hgrecco/pint/blob/b664af69c40821111d13a126accf7deef7dee90d/pint/quantity.py#L272

Would it be sufficient to simply set `kwspec['locale'] = None` in the event that Loc is None?

hgrecco:
1. UnitRegistry now has an optional keyword argument (`locale`) that can be
      used to define the default value for `Quantity.format_babel` locale argument.
   2. When Babel is not installed, `Quantity.format_babel` display a nicer and
      more informative exception.

- [x] Closes #899, #904

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,10 @@ Pint Changelog
 0.10 (unreleased)
 -----------------
 
+- Added `fmt_locale` argument to registry.
+  (Issue #904)
+- Better error message when Babel is not installed.
+  (Issue #899)
 - It is now possible to redefine units within a context, and use pint for currency
   conversions. Read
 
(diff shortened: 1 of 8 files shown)


hgrecco:
I will merge in a day or two. Please feel free to review.

PR #949, opened 2019-12-22
New Implementation of 'Lx' String Format Type

### LINKED ISSUE #814: Incorrect String Formating for type Lx (LaTeX,SIunitx)
I think I have identified a bug with the way that both `pint` and `uncertainties` implement the Lx LaTeX with SIunitx package.

Symptom 1: uncertainty notation () (parentheses) used incorrectly.

> print('{:Lx}'.format( (10.5 * ureg.m).plus_minus(1) ))
>\SI[separate-uncertainty=true]{10.5(1.0)}{\meter}

This is a super nice feature of `pint`, but () notation takes a string of digits that represent the uncertainty of the *least significant digits* of the value and does not accept a decimal.  I would think an easy fix would be to use ' +/- ' notation alternative of \SI to join value and uncertainty, which would accept uncertainty values like this.  LaTeX throws:

> ! siunitx error: "invalid-token-in-uncertainty"
> ! Invalid uncertainty in numerical input '10.5(1.0)'.

Symptom 2: uncertainty notation ' +/- ' used incorrectly

> print('{:Lx}'.format(unc.ufloat(10.5,1) * ureg.m))
> \SI[]{10.5+/-1.0}{\meter} 

I don't understand why `uncertainties` module implements Lx differently, but it is also producing incorrect code.  Should I report this separately to Eric?  `uncertainties` uses the +/- notation, but without the required white space surrounding it.  LaTeX throws:

> ! siunitx error: "misplaced-sign-token"
> ! Misplaced sign token '10.5+'.

I'd be happy to try to fix and send patches, but I want to know if that would be helpful and, if so, if you have an idea about how exactly you would like the bug fix to be implemented.

PorcelainMouse:
- [X] *might* close #814
- [X] The change is fully covered by automated unit tests
- [X] Documented in docs/ as appropriate

Hello hgrecco,

Here is my suggested patch related to issue #814.  After working on it for a while, I considered several solutions and this seemed the best, immediately available option.
This code avoids a compilation error thrown by LaTeX when SIunitx is given uncertainty value containing a decimal place, i.e. \SI{123.4(1.2)}{\meter}.

I think a better solution would be to work with 'uncertainties' to implement a fully compatible string format type implementation, as the current implementation for 'S' is not compatible and, as best I can tell, incorrect.  (See NIST "concise" form for representing uncertain values in a single figure, which is what SIunitx is expecting.)  If this is fixed, I think 'pint' could go back to the original code and achieve even more compatibility.

An alternative solution would be to check for '.' in the () part of the "concise" uncertainty value and just remove it.  I did not do this because it requires that 'uncertainties' always produces nominal and unc values aligned with the same least significant digit, which I think is the case, but I wasn't confident.

However, '.1Lx' still produces invalid LaTeX code.  Such a format spec may be invalid, though; I'm not sure.  This is a separate issue and not related to #814, in any case.

I will now open a bug with 'uncertainties' regarding the implementation of 'S'.  Perhaps I can fix this, too, and then we an re-align the two projects on this issue.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -79,6 +79,11 @@ Pint Changelog
 - TODO #825
 - TODO #824
 - TODO #822
+- New implementation of 'Lx' String Format Type Option
+  The old implementation treated 'Lx' as 'S' as produced by 'uncertainties'
+  package, but that is not fully compatible with SIunitx. The new code protects
+  SIunitx by fixing what unceratinties produces.
+  (Issue #814)
 - TODO #813
 - TODO #811
 - TODO #809
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -356,7 +356,7 @@ The same is true for latex (`L`) and HTML (`H`) specs.
 The formatting specs (ie 'L', 'H', 'P') can be used with Python string 'formatting 
 syntax'_ for custom float representations. For example, scientific notation:
 
-..doctest::
+.. doctest::
    >>> 'Scientific notation: {:.3e~L}'.format(accel)
    'Scientific notation: 1.300\\times 10^{0}\\ \\frac{\\mathrm{m}}{\\mathrm{s}^{2}}'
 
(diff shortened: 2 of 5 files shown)


PorcelainMouse:
Oops, didn't notice the problems. I submitted a new pull request.

PR #951, opened 2019-12-23
ufunc calls with mixed args and upcast types

keewis:
This fixes an issue with calling ufuncs with mixed args including upcast types: Calling a bivariate ufunc like `np.maximum(da, 0 * ureg.m)` resulted in `pint` trying to work with upcast types because `__array_ufunc__` works with type objects while `is_upcast_type` checked object instances.

### DIFF
--- a/pint/compat.py
+++ b/pint/compat.py
@@ -147,6 +147,10 @@ if not HAS_BABEL:
 
 
 def is_upcast_type(other):
+    """ check if the type object is a upcast type
+
+    :param other: type
+    """
     # Check if class name is in preset list
     return other.__name__ in ("PintArray", "Series", "DataArray")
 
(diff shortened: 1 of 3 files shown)


pint/compat.py
```diff
@@ -148,7 +148,7 @@ def _to_magnitude(value, force_ndarray=False):
 
 def is_upcast_type(other):
     # Check if class name is in preset list
```
jthielen: Would you be able to add documentation emphasizing that `other` is expected to be a type object?

keewis: what do you think of
```
check if the type object is a upcast type

:param other: type
```

jthielen: That sounds good for now!

keewis:
I think the tests should go to a PR for #845 or for the tests initially mentioned in #905 (these might actually be the same).

pint/compat.py
keewis: done

hgrecco:
> I think the tests should go to a PR for #845.

I agree, merging

Issue #883, opened 2019-09-15
Integration with Dask (add tests; implement the Dask collection interface on Quantity)


jthielen:
Based on https://github.com/hgrecco/pint/issues/878 and https://github.com/pydata/xarray/issues/525, it would be helpful for [interoperability between xarray, pint, and dask](https://github.com/hgrecco/pint/issues/845) for pint to implement the [dask collection interface](https://docs.dask.org/en/latest/custom-collections.html) for when a pint Quantity wraps a dask array. This should allow a Quantity-wrapped dask array to still behave in a dask-array-like way (i.e., as a "duck dask array"). There could also be convenience methods like `compute()`, `persist()`, and `chunk()`, following xarray's example.

Implementation of this could likely follow or come along with changes discussed in https://github.com/hgrecco/pint/issues/878 and https://github.com/hgrecco/pint/issues/845. Based on @hgrecco's comment (https://github.com/hgrecco/pint/issues/878#issuecomment-530796790), I would guess that this would also all be following a decision being made about https://github.com/hgrecco/pint/pull/875 and https://github.com/hgrecco/pint/pull/764 to know when this should be implemented.

Also, ping @crusaderky, since you've been working a lot with xarray, pint, and dask together, and I'd want to hear your thoughts on this.

crusaderky:
The only caveat I see is that xarray wraps specifically around a dask.array, not a generic dask collection.

The obvious example of another dask collection that should be treated differently is ``dask.dataframe.DataFrame`` -although from a quick test, it looks like ``xarray.Dataset.from_dataframe`` doesn't work, and the constructor of DataArray casts the dataframe into a dask.array, losing index and columns along the way. @shoyer did I miss anything?

So ``isinstance(obj, dask_array_type)`` scattered around the xarray codebase should be changed with something that duck-tests the object to have the ``__dask_*`` magic methods and to be 'numpy-like'. It doesn't help that numpy-like is not something well defined; e.g. ``dask.dataframe.DataFrame`` has a ``__array__`` method, but we don't want to treat it as numpy-like.

jthielen:
> So `isinstance(obj, dask_array_type)` scattered around the xarray codebase should be changed with something that duck-tests the object to have the `__dask_*` magic methods and to be 'numpy-like'. It doesn't help that numpy-like is not something well defined; e.g. `dask.dataframe.DataFrame` has a `__array__` method, but we don't want to treat it as numpy-like.

[...]

Issue #633, opened 2018-04-17
Inconsistency when multiplying NumPy masked arrays


dopplershift:
There's a weird inconsistency when working with numpy masked arrays:
```python
>>> import numpy as np
>>> import pint
>>> ureg = pint.UnitRegistry()
>>> a = np.arange(5)
>>> type(a * ureg.meter)
pint.quantity.build_quantity_class.<locals>.Quantity

>>>type(a * ureg('meter'))
pint.quantity.build_quantity_class.<locals>.Quantity

>>> ma = np.ma.arange(5)
>>> type(ma * ureg.meter)
pint.quantity.build_quantity_class.<locals>.Quantity

>>> type(ma * ureg('meter'))
numpy.ma.core.MaskedArray
```

That last one is creating quite a bit of heartache for me.

So I see that `unit.meter` is a `Unit` instance, whereas `unit('meter')` is a `Quantity`. Also, `Unit.__rmul__` delegates to `Unit.__mul__`:

https://github.com/hgrecco/pint/blob/19d0547a207365c7b0f09012d89ce41c2d3b68dd/pint/unit.py#L141-L147

So this changes a `Unit` into a `Quantity`, and delegates to `Quantity.__mul__` by multiply the `Quantity` on the *left*--that explains why it works.

Based on some tracing with numpy 1.15, it looks like we end up in numpy.ma.core:
```python
    def __mul__(self, other):
        "Multiply self by other, and return a new masked array."
        if self._delegate_binop(other):
            return NotImplemented
        return multiply(self, other)
```
and inside self._delegate_binop():
```python
    def _delegate_binop(self, other):
        # This emulates the logic in
        #     private/binop_override.h:forward_binop_should_defer
        if isinstance(other, type(self)):
            return False
        array_ufunc = getattr(other, "__array_ufunc__", False)
        if array_ufunc is False:
            other_priority = getattr(other, "__array_priority__", -1000000)
            return self.__array_priority__ < other_priority
        else:
            # If array_ufunc is not None, it will be called inside the ufunc;
            # None explicitly tells us to not call the ufunc, i.e., defer.
            return array_ufunc is None
```

Unfortunately, this ends up returning `False`, which results in using the `multiply` function from `numpy.ma.core`. This is because `Quantity` delegates the `__array_ufunc__` attribute to `numpy.ndarray`.

I wonder if it would be enough to set `__array_ufunc__` to `None` in `Quantity`...

hgrecco:
[...]

PR #963, opened 2019-12-30
Add tests and documentation with improvement of downcast type compatibility (part of #845)

### LINKED ISSUE #37: Operating elementwise with numpy masked arrays

### LINKED ISSUE #845: Insuring interoperability of pint with other (non-unit) array-like types

jthielen:
As a part of #845, this PR adds tests for downcast type compatibility with Sparse's `COO` and NumPy's `MaskedArray`, along with more careful handling of downcast types throughout the library. Also included is new documentation on array type compatibility, including the type casting hierarchy digraph by @shoyer and @crusaderky.

While this PR doesn't fully bring Pint's downcast type compatibility to a completed state, I think this gets it "good enough" for the upcoming release, and the remaining issues are fairly well defined:

- MaskedArray non-commutativity (https://github.com/hgrecco/pint/issues/633 / https://github.com/numpy/numpy/issues/15200)
- Dask compatibility (https://github.com/hgrecco/pint/issues/883)
- Addition of CuPy tests (no issue on issue tracker yet)

Because of that, I think this can close #845, but if @hgrecco you want that kept open until the above items are resolved, let me know.

- [x] Closes #37; Closes #845


hgrecco:
I agree with closing 845 and, if needed opening more specific issues referencing the upstream repos

jthielen:
After testing this downstream with MetPy, the issue in https://github.com/hgrecco/pint/issues/37#issuecomment-21671778 reappeared, making it so that multiplying by a Unit on the right gives an object MaskedArray of Quantities instead of a Quantity with MaskedArray magnitude. This ends up being a pretty bad breaking change that makes the situation worse than before.

However, when I reverted the change to Unit replacing `__array_wrap__`/`__array_prepare__` with `__array_ufunc__`, the Sparse test now fails, since Sparse.COO times Unit requires the multiply ufunc to be implemented on Unit (since Sparse rightly defers).

So, unless I'm missing a fix that works for both Sparse and MaskedArray, there is a dilemma:

- Keep the change from this PR implementing `__array_ufunc__` on Unit, which is more future-proof and compatible with hierarchy-obeying types like Sparse.COO, but exacerbates https://github.com/numpy/numpy/issues/15200 and causes more issues with the commonly-used-but-problematic MaskedArray.
- Revert the change until https://github.com/numpy/numpy/issues/15200 is resolved, which keeps the MaskedArray multiplication issue the same as it was, but prevents proper behavior with types that require `__array_ufunc__` on Unit like Sparse.COO

Or, put another way, given that we can't have both, for which type of array would we rather have `array * ureg.m` work:

- Sparse.COO (and other ufunc-reliant types)?
- MaskedArray (as it was before)?

@hgrecco What are your thoughts on this?

hgrecco:
This PR is good step into the right direction. `__array_ufunc__` (or similar) is a much better protocol than the previous one and something that `pint` was waiting for several years. My suggestion would be to keep this PR, document the problem and alert users what is the problem. Hopefully, this will be fixed in numpy soon.

PR #965, opened 2019-12-30
Raise ValueError on ambiguous boolean conversion and add pending NumPy functions/ufuncs from issue tracker

### LINKED ISSUE #419: Monkeypatch np.linalg.solve
Such that `np.linalg.solve(A, x).units = x.units / A.units`

### LINKED ISSUE #470: Support for numpy matrix operations
A few things regarding matrix operations that aren't working properly that would be useful.

- support for `np.dot` e.g. `A.dot(b)` or `b.dot(A)`. Currently, if `A,b` are arrays containing quantities, this works. However, if `A,b` are `Quantity`'s, then this operation does not return the correct result. I believe this is related to #39 . 
- support for `np.linalg.solve`. Solving a system where `A,b` are instances of `Quantity` the method fails with a `TypeError` in line 1235 of `quantity`. I think this is related to #420. If `A,b` are arrays of `Quantity`s the method fails in `linalg` with a `TypeError: no loop matching the specified signature was found for ufunc solve1`

### LINKED ISSUE #807: Applying numpy.cbrt (cubic root) loses quantity
`pint.__version__`
'0.9'
`a = ur.Quantity(100, "m ** 3")`
`print(a)`
100 meter ** 3
`print(np.sqrt(a))`
10.0 meter ** 1.5
`print(np.cbrt(a))`
4.641588833612779

### LINKED ISSUE #866: Expected truthiness of quantities with offsets
With https://github.com/hgrecco/pint/pull/764 and https://github.com/pydata/xarray/pull/3238, in trying to implement proper `np.any` and `np.all` for Quantities, I got to wondering what the expected "truthiness" of a scalar quantity with offset should be. Right now, there seems to be no explicit behavior defined, so it falls back to the magnitude. This leads to the following type of result:

```python
import pint

ureg = pint.UnitRegistry()

temp = ureg.Quantity(0, ureg.degC)

print(bool(temp))
print(bool(temp.to(ureg.degF)))
print(bool(temp.to(ureg.K)))
```
```
False
True
True
```

Should this be the expected behavior?

At first glance, this kind of changing the truthiness of a quantity by changing the unit it is expressed in is a bit unsettling...making a small change like converting the unit doesn't seem like it should change the truthiness. In this case, I think there would need to be an explicit `__bool__` method that either converts to the non-offset equivalent or raises a `ValueError` due to ambiguity.

However, I would also understand if the current behavior is what is expected, since thinking about it more, I would imagine that in most use cases checking for truthiness of a quantity reduces to checking for nonzero values. Also, I'm not sure what pint's stance on breaking existing but undocumented behavior is.

Either way, however, I think tests should be written for the expected behavior and it should be documented clearly (https://pint.readthedocs.io/en/0.9/nonmult.html). Once the expected behavior is decided upon, I'd be glad to put in a PR for that.

jthielen:
There were a few follow-up issues from #905 that were missing fairly simple implementations, so this PR takes care of them in preparation for the upcoming release.

Part of this was resolving #866 by raising a `ValueError` due to ambiguity when casting Quantities with offset units to boolean, which is technically a breaking change (although I would argue that the previous behavior was incorrect as discussed in #866).

Also, @keewis, this implements `np.any` and `np.all`, which are two of the functions you had mentioned were needed by xarray. I think now the only one still missing is `np.prod` (https://github.com/hgrecco/pint/issues/867) which will likely have to wait since there hasn't been a good solution yet.

- [x] Closes #419; Closes #470; Closes #807; Closes #866

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,10 @@ Pint Changelog
 0.10 (unreleased)
 -----------------
 
+- **BREAKING CHANGE**:
+  Boolean value of Quantities with offsets units is ambiguous, and so, now a ValueError
+  is raised when attempting to cast such a Quantity to boolean.
+  (Issue #965, Thanks Jon Thielen)
 - Documentation on Pint's array type compatibility has been added to the NumPy support
   page, including a graph of the duck array type casting hierarchy as understood by Pint
   for N-dimensional arrays.
--- a/docs/numpy.ipynb
+++ b/docs/numpy.ipynb
@@ -284,7 +284,7 @@
     "\n",
     "The following [ufuncs](http://docs.scipy.org/doc/numpy/reference/ufuncs.html) can be applied to a Quantity object:\n",
     "\n",
-    "- **Math operations**: `add`, `subtract`, `multiply`, `divide`, `logaddexp`, `logaddexp2`, `true_divide`, `floor_divide`, `negative`, `remainder`, `mod`, `fmod`, `absolute`, `rint`, `sign`, `conj`, `exp`, `exp2`, `log`, `log2`, `log10`, `expm1`, `log1p`, `sqrt`, `square`, `reciprocal`\n",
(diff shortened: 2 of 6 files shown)


hgrecco:
Awesome. Great work

jthielen:
@hgrecco is there a reason you closed this instead of `bors r+`?

hgrecco:
> @hgrecco is there a reason you closed this instead of bors r+?

my mistake!

PR #987, opened 2020-01-14
minor typo fix

willstott101:
This is probably a pointlessly small change, but I noticed this typo whilst looking at the definitions.


hgrecco:
thanks

PR #989, opened 2020-01-15
Allow constants in units by using leading underscore

### LINKED ISSUE #507: Allow constants within units (e.g. for "L/100km")
A very common unit in car fuel economy is L/100km. Currently, as far as I can tell, there is no way to express this in pint. There's probably other similar common measurements.

As a side note, it would also be good if pint knew how to compute reciprocals directly. (ie (30 ureg.miles / ureg.imperial_gallon).to('L/100km')

jni:
See #507 for discussion.

This change allows code like the following:

```python
In [1]: from pint import UnitRegistry                                              
In [2]: u = UnitRegistry()                                                         
In [3]: u.define('_100km = 100 * kilometer')                                       
In [4]: u.define('mpg = 1 * mile / gallon')                                        
In [5]: volt_petrol_ec = 5 * u.L / u._100km                                        
In [6]: (1 / volt_petrol_ec).to(u.mpg)                                             
Out[6]: 47.042916666666656 <Unit('mpg')>
```

- [x] Closes #507

### DIFF
--- a/pint/testsuite/test_issues.py
+++ b/pint/testsuite/test_issues.py
@@ -699,12 +699,12 @@ class TestIssues(QuantityTestCase):
 
     def test_issue507(self):
         # leading underscore in unit works with numbers
-        u.define("_100km = 100 * kilometer")
-        battery_ec = 16 * u.kWh / u._100km
+        ureg.define("_100km = 100 * kilometer")
+        battery_ec = 16 * ureg.kWh / ureg._100km
         # ... but not with text
-        u.define("_home = 4700 * kWh / year")
+        ureg.define("_home = 4700 * kWh / year")
         with self.assertRaises(AttributeError):
-            home_elec_power = 1 * u._home
+            home_elec_power = 1 * ureg._home
 
 
 try:
--- a/pint/testsuite/test_issues.py
+++ b/pint/testsuite/test_issues.py
@@ -705,6 +705,10 @@ class TestIssues(QuantityTestCase):
         ureg.define("_home = 4700 * kWh / year")
         with self.assertRaises(AttributeError):
             home_elec_power = 1 * ureg._home
+        # ... or with *only* underscores
+        ureg.define("_ = 45 * km")
+        with self.assertRaises(AttributeError):
+            one_blank = 1 * ureg._
 
 
 try:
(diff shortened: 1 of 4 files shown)


pint/testsuite/test_issues.py
```diff
@@ -697,6 +697,15 @@ def test_issue932(self):
         with self.assertRaises(DimensionalityError):
             q.to("joule")
 
+    def test_issue507(self):
+        # leading underscore in unit works with numbers
+        u.define("_100km = 100 * kilometer")
```
hgrecco: `u` is not defined here and therefore all test fails.

pint/util.py
```diff
@@ -876,7 +880,11 @@ def getattr_maybe_raise(self, item):
     """
     # Double-underscore attributes are tricky to detect because they are
     # automatically prefixed with the class name - which may be a subclass of self
-    if item.startswith("_") or item.endswith("__"):
+    if (
+        item.startswith("_")
```
hgrecco: I would add a parenthesis with the first two clauses. It will help to indicate purpose and link to the explanation in the docs. i.e I think the intention would be clear if
```python
item.endswith("__") 
or (item.startswith("_") and not item.lstrip("_")[0].isdigit())
```

Also the above introduces a subtle bug. If `item == "_"`, `IndexError` will be raised.

jni:
@hgrecco oops! Thank you for the speedy review, and apologies for the errors, I mixed up my own interactive testing and the unittests! These run on my machine now. :sweat_smile: 

I've also fixed that logic bug you pointed out.

Hopefully this gets it over the line! :crossed_fingers: Thank you!

hgrecco:
The linter is the only missing thing, after that I will be able to merge.

jni:
Ach! Incidentally, maybe invert the second and third items in the PR checklist, as I ran through them one by one, which means the linter didn't get run on the test. :joy: 

Um, how would you like me to deal with the unused variables? `_ = `? Or adding a comment for flake8 to ignore that issue? (Sometimes I like to have variables defined as a way to better document what a line of code is doing...)

hgrecco:
Just add a coment for flake8 to ignore.

jni:
Done! Thank you for your patience!

PR #993, opened 2020-01-17
WIP: compatibility checks

### LINKED ISSUE #988: checking for unit compatibility
What is the official way to check compatibility between units (or arrays and units)? Right now I'm using
```python
array.check(unit)  # for mixed array / unit
array.dimensionality == unit.dimensionality  # same as above

array.dimensionality == other_array.dimensionality  # for two arrays
unit.dimensionality == other_unit.dimensionality  # for two units
```
but it took me some time to realize this was possible (it is not really obvious to me and I couldn't find it in the docs -- I might have missed it, though).

I usually put this in a function named something like `is_compatible` to make it a little bit more robust. If there is no such function yet, is there any interest in adding one?

keewis:
Continuing from the issue, this is the code from #988. It is missing documentation and tests right now; I will add these later. Also, I am not sure where to put a potential top-level function.

- [x] Closes #988

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -505,6 +505,22 @@ class Quantity(PrettyIPython, SharedRegistryObject):
         return self._REGISTRY.get_compatible_units(self._units)
 
     def is_compatible_with(self, other, *contexts, **ctx_kwargs):
+        """ check if the other object is compatible
+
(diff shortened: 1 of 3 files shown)


keewis:
is this okay without tests?

PR #998, opened 2020-01-21
Add parse_pattern function

### LINKED ISSUE #997: Feature request: Add a `parse_pattern` function which parses strings using regex patterns
I was searching for a library which could parse units based on a pattern given, e.g. with the pattern `{foot}'{inches}` and the input `10'11`, it would apply the pattern to the input and then convert the results to units. It's a shame that none seem to do it, and I think it would be great if pint had such a feature.

abmyii:
- [X] Closes #997.
- [?] Documented in docs/ as appropriate
- [?] Added an entry to the CHANGES file

I don't like to open issues that I am able to implement myself, so I quickly made a POC implementation which seems to work quite well and is sufficient for my purposes.

Parsing with regex isn't as clean as parsing with the `parse` (https://github.com/r1chardj0n3s/parse) library which I initially thought of using, but I decided that after such a long time of being fully independent of third-party libraries, you wouldn't want to begin now. Thus readability has been sacrificed at the cost of upholding the design principles. Personally I don't mind as long as the feature is usable.

Thank you for this great module!

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -251,14 +251,14 @@ or
 
 .. note:: Pint´s rule for parsing strings with a mixture of numbers and
    units is that **units are treated with the same precedence as numbers**.
-   
+
 For example, the unit of
 
 .. doctest::
 
    >>> Q_('3 l / 100 km')
    <Quantity(0.03, 'kilometer * liter')>
-   
+
 may be unexpected first but is a consequence of applying this rule. Use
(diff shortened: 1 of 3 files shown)


abmyii:
https://github.com/hgrecco/pint/pull/998/commits/d0489c32679ab8fe8bfe42148249e1bb8ab7a8e6

Proof that regexs aren't as convenient. Oh well.

PR #999, opened 2020-01-22
Preparing the codebase for a better Decimal support

hgrecco:
This is a series of three commits to prepare the codebase for a better Decimal support. These are mostly and internal clean up and should not affect any user.

- [ ] ~~Closes # (insert issue number)~~
- [ ] ~~The change is fully covered by automated unit tests~~
- [ ] ~~Documented in docs/ as appropriate~~


hgrecco:
@jthielen @keewis  I would appreciate if you can take a look to this PR. Appart from the changes in the code you can read the idea behind each in the commit message.

hgrecco:
I am merging this on Jan 29th if there are no objections.

keewis:
looks good to me, although reading `ParsedDefinition.from_string` feels a little bit odd. I'd suggest moving that to a toplevel `parse` function, but I guess it was added for consistency with the other `Definition` classes?

pint/definitions.py
```diff
@@ -8,17 +8,79 @@
...
+    """
+
+    @classmethod
+    def from_string(cls, definition):
+        name, definition = definition.split("=", 1)
+        name = name.strip()
+
+        rhs_parts = tuple(res.strip() for res in definition.split("="))
+
+        value, aliases = rhs_parts[0], tuple([x for x in rhs_parts[1:] if x != ""])
```
keewis:
```suggestion
        value, aliases = rhs_parts[0], tuple(x for x in rhs_parts[1:] if x != "")
```
you can just use the generator expression instead of a list comprehension converted to a tuple

pint/definitions.py
```diff
@@ -8,17 +8,79 @@
...
+        symbol, aliases = (aliases[0], aliases[1:]) if aliases else (None, aliases)
+        if symbol == "_":
+            symbol = None
+        aliases = tuple([x for x in aliases if x != "_"])
```
keewis: same:
```suggestion
        aliases = tuple(x for x in aliases if x != "_")
```

pint/definitions.py
hgrecco: agreed. I will fix this.

pint/definitions.py
hgrecco: agreed, I will fix this.

hgrecco:
A module level `parse` method would be equivalent of `Definition.from_string`. In general, I rather have classmethods constructors.

The purpose of `ParsedDefinition` was to avoid reparsing the common structure while still having the possibility in `Definition` to access the information needed to dispatch to the corresponding subclass. In retrospective, I think it would be better to call it `DefinitionTuple`.

keewis:
true, `parse` is not really a good name. What about a static method named `Definition.preprocess` that returns that `DefinitionTuple`?

hgrecco:
> true, parse is not really a good name. What about a static method named Definition.preprocess that returns that DefinitionTuple?

I like that.

keewis:
I guess you'd call the class `PreprocessedDefinition`, then?

hgrecco:
I wil merge this as is and then include your proposed changes in another PR

Issue #967, opened 2020-01-02
Floating Point Error with decimal


smiddy:
I use the pint library with decimal to avoid floating point errors. I found a case in which an error still occurs. I use Python 3.7.1

```python
import pint
import decimal

ureg = pint.UnitRegistry()

delay = decimal.Decimal('200000e-9') * ureg.second

print('Pint Version:', pint.__version__)
print('decimal Version:', decimal.__version__)
print(delay)
print(delay.to('nanosecond'))

```
results in the output

```python
Pint Version: 0.9
decimal Version: 1.70
0.000200000 second
199999.9999999999800000 nanosecond
```

jules-ch:
Closing since it has been fixed in last version of pint.

```python

>>> import pint
>>> import decimal
>>> ureg = pint.UnitRegistry(non_int_type=decimal.Decimal) 
>>> delay = decimal.Decimal('200000e-9') * ureg.second
>>> print('Pint Version:', pint.__version__)
Pint Version: 0.17
>>> print('decimal Version:', decimal.__version__)
decimal Version: 1.70
>>> print(delay)
0.000200000 second
>>> print(delay.to('nanosecond'))
200000 nanosecond

Issue #794, opened 2019-04-17
use_decimal produces floats


cpitclaudel:
Here's a repro:

```
>>> ureg = UnitRegistry()
>>>  type(ureg('1.0 m / s', use_decimal=True).magnitude)
<class 'decimal.Decimal'>
>>>  type(ureg('1 m / s', use_decimal=True).magnitude)
<class 'float'>
```

This causes the following error (see #788):

```
>>> ureg('1 m / s / (1.5 s)', use_decimal=True)
*** TypeError: unsupported operand type(s) for /: 'float' and 'decimal.Decimal'
```

cc @franciscouzo

franciscouzo:
A possible fix is to always convert to `Decimal` when `use_decimal` is `True`, currently, the conversion is only done when the token cannot be represented as an `int`

https://github.com/hgrecco/pint/blob/7b9f1ba3683c4ff0156088e4541db753cf3ea1d5/pint/util.py#L429-L435

cpitclaudel:
Indeed, I saw that snippet.  I wonder why the code was written like this?

cpitclaudel:
I think this might be the only reasonable trick; if we don't do that, then there's no difference between `ureg('1')` and `ureg('1', use_decimal=True)`, which means that then the `*` and `/` operator can't know which unit to use.

cpitclaudel:
Ah, this solution doesn't work: it breaks exponents.

```
>>> ureg('m ** 2', use_decimal=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../pint/pint/registry.py", line 865, in parse_expression
    return build_eval_tree(gen).evaluate(lambda x: self._eval_token(x,
  File ".../pint/pint/pint_eval.py", line 85, in evaluate
    return bin_op[op_text](left, self.right.evaluate(define_op, bin_op, un_op))
  File ".../pint/pint/quantity.py", line 76, in wrapped
    result = f(self, *args, **kwargs)
  File ".../pint/pint/quantity.py", line 1164, in __pow__
    units = new_self._units ** other
  File ".../pint/pint/util.py", line 366, in __pow__
    d[key] *= other
TypeError: unsupported operand type(s) for *=: 'float' and 'decimal.Decimal'
```

cpitclaudel:
Constructing a container doesn't work either:

```
>>> ureg.Unit(pint.util.UnitsContainer({'m': Decimal(1.3)})) ** Decimal(2)
[...]

Issue #452, opened 2016-11-06
Decimal conversion with offset


txxia:
Hello,
I started using pint as it supports Decimal types, but seems like it's not handled in conversion with offset.

Here is an example:
```python
>>> from pint import UnitRegistry
>>> from decimal import Decimal
>>> UR = UnitRegistry()
>>> q = UR.Quantity(Decimal('0'), 'K')
>>> q
<Quantity(0, 'kelvin')>
>>> q.to('celsius')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Program Files (x86)\Python35-32\lib\site-packages\pint\quantity.py", line 263, in to
    magnitude = self._convert_magnitude_not_inplace(other, *contexts, **ctx_kwargs)
  File "C:\Program Files (x86)\Python35-32\lib\site-packages\pint\quantity.py", line 231, in _convert_magnitude_not_inplace
    return self._REGISTRY.convert(self._magnitude, self._units, other)
  File "C:\Program Files (x86)\Python35-32\lib\site-packages\pint\unit.py", line 1026, in convert
    return self._convert(value, src, dst, inplace)
  File "C:\Program Files (x86)\Python35-32\lib\site-packages\pint\unit.py", line 1134, in _convert
    value = self._units[u].converter.from_reference(value, inplace)
  File "C:\Program Files (x86)\Python35-32\lib\site-packages\pint\converters.py", line 80, in from_reference
    value = (value - self.offset) / self.scale
TypeError: unsupported operand type(s) for -: 'decimal.Decimal' and 'float'
```

hgrecco:
This should be fixed! #

onealj:
What is the expected behavior? If a quantity has a Decimal magnitude, what should be the type of the quantity returned by a converter function?
# Decimal, unsafely converting floats to Decimal as necessary
# Decimal, duplicating the converter code to handle Decimal math and float math differently
# float, drop down to lowest common denominator

hgrecco:
We already support `float` and `Fraction` for multiplicative convertors. We could do the same for offset convertors. Another question is if we should reparse the definition (in a Lazy way) to do it properly.

pinkynrg:
Any progress on this issue?

hgrecco:
Not really. A PR would be welcomed.

hgrecco:
This has been solved by the PR, feel free to reopen if needed.

PR #1000, opened 2020-01-22
WIP: Add support for other numerical types at the registry level

### LINKED ISSUE #968: Decimal Parsing Issue
In version 0.9 the following code doesn't work with python 3.7.5.

```python
import decimal
import pint

registry = pint.UnitRegistry()

mps = decimal.Decimal('65000.0') * registry('m/s')
```

I get the following exception:
```
  File "<console>", line 1, in <module>
  File "env/lib64/python3.7/site-packages/pint/quantity.py", line 900, in __mul__
    return self._mul_div(other, operator.mul)
  File "env/lib64/python3.7/site-packages/pint/quantity.py", line 75, in wrapped
    result = f(self, *args, **kwargs)
  File "env/lib64/python3.7/site-packages/pint/quantity.py", line 60, in wrapped
    result = f(self, *args, **kwargs)
  File "env/lib64/python3.7/site-packages/pint/quantity.py", line 866, in _mul_div
    magnitude = magnitude_op(self._magnitude, other_magnitude)
TypeError: unsupported operand type(s) for *: 'float' and 'decimal.Decimal'
```

The work around is to use quantities:

```python
Q_ = registry.Quantity                                                                                                                                                                                                           
Q_(decimal.Decimal('65000.0'), 'm/s')
```
I would expect that both would work?  Is this not correct?

hgrecco:
Until this commit, `int` and `float` were special types in Pint
(just as they are in Python). Numbers were parsed from strings
as int/float from the definition files and from user provided
strings; and exponents of units were also stored as int/float.

This commit change this by adding a new argument (`non_int_type`)
to classes and methods. It indicates how numeric values will be parsed
and defaulted. Any numerical class can be used such as `float` (default),
Decimal, Fraction.

This argument will be found in the following places
1. UnitRegistry: used for parsing the definition files and any 
  value provided as a string. 
2. UnitsContainer: used to compare equality with strings, multiply and
  divide by strings (which is equivalent to parse the string)
3. All methods OUTSIDE the UnitRegistry/Quantity that can parse
   strings have a `non_int_type` argument. (e.g. Definition.from_string)

Tests have been added for by duplicating most cases in `test_quantity.py`.
(Some tests have been deleted such as those dealing with numpy.)
The new file `test_non_int.py` run the tests for `Decimal`, `Fraction` and
`float` (which is redundant but is kept as a crosscheck for the implementation
of this testsuite)

BREAKING CHANGE: `use_decimal` is deprecated. 
Use `non_int_type` keyword argument when instantiating the registry.
```python
 >>> from decimal import Decimal
 >>> ureg = UnitRegistry(non_int_type=Decimal)
```

- [x] Closes #968,  #967, #794, #452 


hgrecco:
Please notice that some of these commits are already included in #999 as they are really independent of decimal. 

Testing, ideas, comments and criticism are welcomed and desired. Importantly, this PR changes the idea about decimal support. It is a commitment of the `UnitRegistry` instance and no longer an option at each call to a parse method. Also, if you choose decimal (or any numerical type), you are then responsible to ensure that operations between those quantities and other values your provide are valid. 

Also, full support for Decimal comes at PR # 1000 🎊

hgrecco:
I have just released 0.11 and during the weekend and I will merge this commit into master. The idea is to have a full month to play with master until we release 0.12. This is a rather important commit as it changes the internal type of the definitions and there might be some flaws.

PR #1023, opened 2020-02-11
Make `__str__` and `__format__` locale aware

### LINKED ISSUE #984: [PROPOSAL] Make `__str__` and `__format__` locale aware
Since #944, the format locale of the registry can be defined using [set_fmt_locale](https://github.com/hgrecco/pint/blame/master/pint/registry.py#L294). 

I propose here to continue in this path and change `__str__` and `__format__` to automatically use `format_babel` under the hooed when the registry locale is not `None`. To ensure backwards compatibility, `None` (which means that no locale is used) will be the default value for the registry locale (as it is today).

hgrecco:
With this commit, all string formatting operations are locale aware

    >>> ureg.set_fmt_locale('fr_FR')
    >>> str(accel)
    '1.3 mètre par seconde²'
    >>> "%s" % accel
    '1.3 mètre par seconde²'
    >>> "{}".format(accel)
    '1.3 mètre par seconde²'

It should not break any code as the default locale value for the
Registry is `None` (meaning do not localize).

Close #984

- [x] Closes #984 (insert issue number)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Pint Changelog
 0.11 (unreleased)
 -----------------
 
+- Make `__str__` and `__format__` locale aware
 - Remove `default_en_0.6.txt`
 - Reorganize long_description.
 - Moved Pi to defintions files.
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -419,9 +419,26 @@ Finally, if Babel_ is installed you can translate unit names to any language
    >>> accel.format_babel(locale='fr_FR')
    '1.3 mètre par seconde²'
 
-You can also specify the format locale at u
+You can also specify the format locale at the registry level either at creation:
 
     >>> ureg = UnitRegistry(fmt_locale='fr_FR')
+
+or later:
+
+.. doctest::
+
+    >>> ureg.set_fmt_locale('fr_FR')
+
+and by doing that, string formatting is now localized:
+
+.. doctest::
+
+    >>> str(accel)
+    '1.3 mètre par seconde²'
+    >>> "%s" % accel
+    '1.3 mètre par seconde²'
+    >>> "{}".format(accel)
+    '1.3 mètre par seconde²'
 
 
 Using Pint in your projects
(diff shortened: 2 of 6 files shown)


hgrecco:
@dalito @bors-ltd @alexbodn  Can you provide some feedback on this PR and the idea behind?

docs/tutorial.rst
```diff
@@ -419,10 +419,27 @@ Finally, if Babel_ is installed you can translate unit names to any language
...
+and by doing that, string formatting is now localized:
+
+.. doctest::
+
+    >>> str(accel)
+    '1.3 mètre par seconde²'
+    >>> "%s" % accel
+    '1.3 mètre par seconde²'
+    >>> "{}".format(accel)
+    '1.3 mètre par seconde²'
```
bors-ltd: If that's actual Babel output, they got the French wrong. :joy:

CHANGES
```diff
@@ -4,6 +4,7 @@ Pint Changelog
 0.11 (unreleased)
 -----------------
 
+- Make `__str__` and `__format__` locale aware
```
bors-ltd: You say "locale aware" but it's not actually influenced by the locale environment by default.

pint/registry.py
```diff
@@ -216,7 +219,7 @@ def __init__(
         self.auto_reduce_dimensions = auto_reduce_dimensions
 
         #: Default locale identifier string, used when calling format_babel without explicit locale.
-        self.fmt_locale = self.set_fmt_locale(fmt_locale)
+        self.set_fmt_locale(fmt_locale)
```
bors-ltd: Changes to this file are independent of this PR. Have you considered using pylint to detect calling this function that doesn't return any value.

CHANGES
hgrecco: True, you can still use `set_fmt_locale("sys") which will use the system locale. The rationale behind being unaffected by the default locale (i.e. keeping it in the same language that the definition file) is to avoid affecting existing code.

pint/registry.py
hgrecco: They are independent but this was a bug that need to be fixed to make this PR work.

docs/tutorial.rst
hgrecco: Indeed this is the babel output.

PR #1027, opened 2020-02-14
Implements Logarithmic Units for units such as dB, dBm, octave or decade

### LINKED ISSUE #71: log and exponential unit conversions
In optics, the power of light is often measured in dBm. The definition of dBm is:

dBm = 10 \* log(mW)

It means that 0 dBm is 1 mW, 10 dBm is 10 mW and -10 dBm is 0.1 mW.

The opposite conversion is:

mW = 10**(dBm / 10)

5igno:
- [ ] Closes #71 


5igno:
@hgrecco: do you have any remarks? Do you see anything missing?

5igno:
Summary of missing points:
- [ ] Documentation: how to we reshape it?
- [ ] Change delta creation logic for LogUnits
- [ ] Change `autocorrect_offset_to_baseunit` to `autocorrect_nonmul_to_base`
- [ ] Warn user that `autocorrect_nonmul_to_base` should be true if LogUnits are to be used in the context of multiplications

terikin:
@5igno Thanks for the work on this; just wanted to add my name as another person interested in having this capability.

One pedantic comment on your additions to `default_en.txt` - "decibel" (and "bel") should be spelled with a single 'L' (see, e.g., [wikipedia](https://en.wikipedia.org/wiki/Decibel)).

5igno:
Oups ... Thank you for pointing this out 😅

terikin:
One additional comment - I've tried using your changes, and encountered an error when trying to construct a Quantity from a string, for example:
- `ureg.Quantity('33 dBm')`

fails with the error:
 - `pint.errors.OffsetUnitCalculusError: Ambiguous operation with offset unit (decibelmilliwatt, )`
raised from `quantity.py`, line 1121

I can create quantities fine if I pass the units separately, such as:
 - `ureg.Quantity(33, ureg.dBm)`

I've dug into it a bit, but I'm new to Pint and thus slow, at least for now...

If I understand it correctly, it seems to be trying to multiply a dimension-less '33' with '1 dBm', and fails when multiplication isn't defined for something with logarithmic dimensions (`Quantity._ok_for_muldiv(...)` returns `false`).  I don't know what the right way to deal with that is, since multiplication doesn't make sense for dB quantities with a reference to another unit (as far as I'm aware).

However, un-referenced dB quantities can make sense in at least one context - attenuation of radio energy in cables, where the attenuation is often specified with dimensions of "dB / km" or even "dB / 100 ft".  In that case, it's entirely correct to be able to multiply the attenuation (say dB/km) by a distance (km), yielding a dimension of dB, which can then be added to other un-referenced dB values (or a single referenced value).

5igno:
Hi @terikin,

Thanks for checking out the suggested changes :)
Indeed you are pointing your finger in the right spot.
The question is what kind of functionality we want to have four our dB units.

The problem you face with "1 dBm"  is most shared with quantities that are non multiplicative, such as temperature. If I am not mistaken, if you try to do operations with "degC" you bump in the same kind of issues.
The problem is solved if you allow pint to transform units to base units, for which operations are well defined. That is done with the flag autocorrect_offset_to_baseunit = True.

In my opinion, there are two ways to go ahead:
- use pint to express the units but leave the work to the user/programmer to make sure that the computation is correct
- have pint do all the heavy lifting and do the computation.
So far, the changes I have suggested address the first point but not the second.

Let me comment on your reply:
> I don't know what the right way to deal with that is, since multiplication doesn't make sense for dB quantities with a reference to another unit (as far as I'm aware).

What I put together in the PR is a proposal for how it could work ... the question is: how would you want it to work? Are you happy with the error that you get now? Or would you prefer pint to give you as output quantity with a unit that may or may not make sense?

> multiplying/dividing un-referenced dB quantities can make sense in at least one context - attenuation of radio energy in cables, where the attenuation is often specified with dimensions of "dB / km" or even "dB / 100 ft".

For pint to provide a result, one has to understand what it means when we do operations with logarithmic units. Let me give an example:

Q(10, ureg.dBm) - Q(20, ureg.dB)

The operation above could pop-up when you have a transmitter that generates 10 dBm and a signal goes through a channel that has -20 dB losses.

Under the hood, that sum (difference really) is in reality a product (ratio) in linear units.
So in this case the parser should: 
- be aware that you are trying to make the difference between log units
- convert them in base units, which are mW and dimensionless respectively
- change the sum into a product, a difference into a ratio, a product into an exponentiation
- compute the result
- go back into a logarithmic unit, if that is available

Q(-10, ureg.dB / ureg.km) * Q(1, ureg.km)

The operation above could pop-up when you have a lossy channel with -10 dB/km losses, which attenuates the power of a signal by 0.1 for every km of propagation, and the channel is 1 km long.

So in this case the parser should: 
- be aware that you are trying to make the product between a mixed log and base units
- convert them log unit into base unit, which is from dB/km to dimensionless/km
- change the product into an exponentiation
- compute the result
- go back into a logarithmic unit, if that is available

Let us make one last example for fun.

Q(10, ureg.dBm) + Q(20, ureg.dBm)

What should the result of this be? What do we want pint to support?
Let's try to see what this operation means.
The parser should: 
- convert the units to base units:
Q(10, ureg.dBm) -> 10 mW
Q(20, ureg.dBm) -> 100 mW
- change the sum into a product
- compute the result
Q(10, ureg.dBm) + Q(20, ureg.dBm) -> 10 mW * 100 mW -> 1000 mW^2 
- go back into a logarithmic unit, if that is available

As it is know the last step cannot be done, because mW^2 is not defined in the list.
You might say yourself: but mW^2 does not make sense!
My perspective on this is that this is not pint's problem: it's the engineer problem to make sense of the output.

So now the question is: do we want pint to be able to generate dBmW^2?
As it is coded now, pint cannot do this job.

The only way i see to achieve something like that is that dB becomes a non-linear prefix, such as k in km or kg (see the comment of [oogieoogieful](https://github.com/hgrecco/pint/issues/71#issuecomment-323438139) ).
I asked @hgrecco if we wanted to go this route, but he preferred to leverage the work done on non_mul units (aka temperatures). Since he is the maintainer of the project, i think it's important we contribute something that he would be happy to live with.

Summarizing:
Please try using the fork of pint further and comment what you like and what you don't like.
Try also to do some operations with Temperatures (dB will follow the same patterns) and comment what you like and what you don't like.
We can then decide together with @hgrecco how to wrap this PR up.

clarkgwillison:
Hi @5igno - thanks for your work on this PR so far, I'm excited to see it because it's something that would make pint very useful for electrical engineering calculations :)

I have a user story I'd like to submit for this feature, I'm happy to try and provide a failing test if you like it/think it's possible/advisable to implement. I think I follow your discussion above, but may be missing some subtleties re: the temperature offset example, please forgive if that's the case (happy to submit updated docs if so).

My use case is for calculating RMS voltage from dBm, and I think it puts my vote more on the side of "have pint do all the heavy lifting and do the computation (wherever it's not actually nonsensical or clearly wrong)"

If I'm trying to get the RMS Voltage amplitude of a signal in a 50 Ohm transmission line, I can use pint like so:
```
>>> ((1 * ureg.mW * 50 * ureg.ohm)**0.5).to(ureg.volt)
0.22360679774997896 𝑣𝑜𝑙𝑡
```
Which is the correct answer (for the VRMS of a sinewave at least).

And I can now use your PR to compute `dBm` from `mW` correctly also:
```
>>> (1 * ureg.mW).to(ureg.dBm)
0.0 𝑑𝑒𝑐𝑖𝑏𝑒𝑙𝑙𝑚𝑖𝑙𝑙𝑖𝑤𝑎𝑡𝑡
```
But if I want to compute amplitude on the same 50 Ohm line using `dBm` as a starting unit (or any other calculation that converts `dBm` back to `mW`) I get an error.

What I would like to get instead is:
```
>>> (0.0 * ureg.dBm).to(ureg.mW)
1.0 milliwatt
>>> ((0.0 * ureg.dBm * 50 * ureg.ohm)**0.5).to(ureg.volt)
0.22360679774997896 𝑣𝑜𝑙𝑡
```
Without using the `Quantity(...).to(...)` idiom to do the conversion, as its syntax isn't as clean.

5igno:
Hi Clark,

Indeed, feedback on use cases is very much appreciated and the reason for
discussing in this PR.
The way the PR works now, as I have tried to explain in the previous
section, is that pint allows to express whatever measurement unit into a
logarithmic equivalent.
For your case, pint allows to transform dBm to mW and back.
What you are trying to do in your example is using dBm for a operation (a
product) that should rather be done in base units.
If you want to use the system as it is now, you should convert dBm to the
base unit mW first, and then perform the normal linear-unit operations you
know and love:

>>> (((0.0 * ureg.dBm).to(ureg.mW) * 50 * ureg.ohm)**0.5).to(ureg.volt)

 That should give you the result you need.

The question for you now is: what kind of exception/ Error message would
you like to receive so that we point you to this solution?

clarkgwillison:
Oops, sorry, hadn't refreshed the page to see your reply before updating my original comment.

If this worked, that would be most of the way there:
```
>>> (0.0 * ureg.dBm).to(ureg.mW)
1 milliwatt
```
As it is right now, I get the following:
```
>>> (0.0 * ureg.dBm).to(ureg.mW)
...
OffsetUnitCalculusError: Ambiguous operation with offset unit (decibellmilliwatt, ). See https://pint.readthedocs.io/en/latest/nonmult.html for guidance.
```
At least as of `aa73ab7`

As for error messages, the most helpful to me would have been just what you said, something like:
```
LogarithmicConversionError: Pint does not yet support automatic conversions between logarithmic and base units. Try using .to_base_units() to convert from (decibellmilliwatt, ) before continuing your calculation. See https://pint.readthedics.io/en/latest/{a-dB-specific-help-page}.html for guidance.
```
That would give me an immediate suggestion of how to fix the problem as well as a place to get further information and examples specific to calculating with dB quantities.

hgrecco:
I think this is maturing nicely and it might be a good moment to include it in the next version of Pint. Log units are a long awaited feature.

As this would be included in a beta like quality, I think we need to focus on managing the user expectations:
1. Improve the docs regarding logarithmic units. 
 - What works, what does not work
 - How does it interacts with other features (i.e. `autoconvert_offset_to_baseunit`)
 - Provide examples
2. Improve the error messages

And then we can include this in the next release.

clarkgwillison:
I've merged the 0.13 release and got some of that documentation work done in this branch: https://github.com/clarkgwillison/pint/tree/new-log-units

@5igno: would you like a PR from there into your branch to update this?

That branch does include a couple failing tests, as things like this are still happening:
```python
# concerning behavior
>>> (-161 * ureg("dBm/Hz")).to("mW/Hz")
<Quantity(-202.686991, 'milliwatt / hertz')>

# expected behavior
>>> (Q_(-161, "dBm") / ureg.Hz).to("mW/Hz")
<Quantity(7.94328235e-17, 'milliwatt / hertz')>

# still broken
>>> (-161 * ureg("dBm/Hz")).to("dBm/Hz")
Traceback (most recent call last)
...
AssertionError:

# mostly working
>>> Q_(-161, "dBm/Hz").to("dBm/Hz")
<Quantity(-161, 'delta_decibellmilliwatt / hertz')>

# still broken
>>> Q_(-161, "dBm/Hz").to("mW/Hz")
Traceback (most recent call last)
...
AssertionError:
```
and I figured some of them may be easy fixes (not for me, I'm still unfamiliar with the internals, but perhaps for @hgrecco)

I can switch gears to documenting the gotchas and working examples if you want, but that calculation error above gave me pause; it still felt a little too alpha.

5igno:
I forgot to mention one detail. The way I normally use pint is with the following flag on:

```
from pint import UnitRegistry
ureg = UnitRegistry(autoconvert_offset_to_baseunit=True)
Q_ = ureg.Quantity
```

then I get this:
```
In [10]: from pint import UnitRegistry

In [11]: ureg = UnitRegistry(autoconvert_offset_to_baseunit=True)

In [12]: Q_=ureg.Quantity

In [13]: Q_(0,ureg.dBm/ureg.Hz).to(ureg.mW/ureg.Hz)
Out[13]: 1.0 <Unit('milliwatt / hertz')>

In [14]: (0*ureg.dBm/ureg.Hz).to(ureg.mW/ureg.Hz)
Out[14]: 1.0 <Unit('milliwatt / hertz')>

In [15]: (0*ureg.dBm).to(ureg.mW)
Out[15]: 1.0 <Unit('milliwatt')>

In [16]: (-161*ureg.dBm/ureg.Hz).to(ureg.mW/ureg.Hz)
Out[16]: 7.943282347242738e-17 <Unit('milliwatt / hertz')>

In [17]: (Q_(-161, "dBm") / ureg.Hz).to("mW/Hz")
Out[17]: 7.943282347242738e-17 <Unit('milliwatt / hertz')>
```

Indeed when using the string parsing, I also get the different result as you state:

````
In [25]: (-161 * ureg("dBm/Hz")).to("mW/Hz")
Out[25]: -202.68699129886093 <Unit('milliwatt / hertz')>
````

**Comment by 5igno** [comment]
OK forget about what I just said: I see you have that setup too from the documentation.

**Comment by 5igno** [comment]
Interesting:

```
In [34]: (-161 * ureg("dBm/Hz"))
Out[34]: -202.68699129886093 <Unit('gram * meter ** 2 / hertz / second ** 3')>
```

**Comment by hgrecco** [comment]
> Interesting
> In [34]: (-161 * ureg("dBm/Hz"))
> Out[34]: -202.68699129886093 <Unit('gram * meter ** 2 / hertz / second ** 3')>

This is related to something that I wanted to suggest to all the people involved with the (awesome) PR. Fix things work in the following order

First for single (non-compound) units

A. 
1. Unit by attribute. e.g. `u = ureg.dBm`
2. Quantity by constructor. e.g. `q = Q_(1.0, u)`
3. Conversion `q.to('milliwatt')`

B.
1. (as before)
2. Quantity by magnitude times Unit. e.g. `q = 1.0 * u`
3. (as before)

C.
1. Unit by parsing. e.g. `u = ureg.parse_units('dBm')`
2. (as before in A or B)
3. (as before in A or B)

Make it work without surprises for both `autoconvert_offset_to_baseunit` settings). Then repeat for compound units (e.g. `dBm/Hz`)

**Comment by 5igno** [comment]
Thank you @hgrecco for the input: it helps to structure the tests more thoroughly.

As far as my understanding of pint goes, it seems like there are issues related to pint's algorithm that decides how to perform the unit conversion. Somewhere the algorithm's intent is to be "smart" and simplify operations when units are the same. Units are the same by definition when they have the same unit.

In this case we are facing units that are "the same" (mW) but are not the same, since they are related to one-another by the log scale. Pint however does not know that and is happy serve you the base unit.

Unfortunately I am not sure I will be able to put any effort in solving the problem in the short term.
Meanwhile I have observed that people are starting to fork my fork of pint to be able to play around with log units.
Others like @clarkgwillison may want to push this forward but don't know if they should submit a PR to my for or pint iteself.

@hgrecco do you think it would be possible to create a feature branch, where multiple people are contributing to? In this way, people will still fork the root pint and contributing becomes more easy.
I have myself copied the content of @dimapu to place the PR.
Solving all the problems in one PR seems a little challenging and there may be a better way to handle this problem. A feature branch would help in my opinion.

**Comment by 5igno** [comment]
I mean ... if you create a log_units feature branch, we can all contribute PR to that branch, until we see we have reached maturity and we merge into master.

**Comment by scasagrande** [comment]
+1 would love to see this merged in, even if the functionality is behind some "beta" warnings, opt-in feature flag, etc. Its the piece we need at InstrumentKit to convert our system over to using `pint`.

Worst case we could specify a commit hash as our install dependency.

**Comment by 5igno** [comment]
@hgrecco I am not sure how to solve the linter problem: I do not get any issue in my machine, but it seems to be a problem in the CI pipeline.

> All done! ✨ �🍰✨✨
> 47 files would be left unchanged.

**Comment by keewis** [comment]
the issue is `isort`:
```
ERROR: /home/travis/build/hgrecco/pint/pint/compat.py Imports are incorrectly sorted and/or formatted.
```
running `python -m isort .` should fix that for you.

5igno:
@hgrecco, code is ready to merge if needed. I was wandering ... what still needs to be improved?

Thanks to the contribution of @clarkgwillison, we made progress on two points:

- [X] Documentation: drafted by @clarkgwillison
- [X] Warn user that autocorrect_nonmul_to_base should be true if LogUnits are to be used in the context of multiplications: this is now covered in the documentation and error that points to it

We still need to implement/improve on:
- [ ] Change delta creation logic for LogUnits
- [ ] Change autocorrect_offset_to_baseunit to a dedicated flag
- [ ] Handle dB for amplitude (20 log10) with [context](https://github.com/hgrecco/pint/issues/71#issuecomment-585290868)

Before the introduction of logarithmic units there was no distinction between non_multiplicative and offset units.
With the introduction of LogUnits, the distinction between non_multiplicative and offset units becomes real: in my view, non_multiplicative becomes a shared (True) feature of both logarithmic and offset units. Logically log_units are non_multiplicative but behave differently from offset units.

As it is written in the documentation: many of the examples fail without supplying the autoconvert_offset_to_baseunit=True flag. 

In my opinion, we should have the logic for Logarithmic units decoupled from the one of offset units, creating a separate flag that will most probably be True all the time. This would enable the user to skip the initial phase:
https://github.com/5igno/pint/blob/log_units/docs/log_units.rst#setting-up-the-unitregistry

PR #1031, opened 2020-02-19
Rename `ParsedDefinition` to `PreprocessedDefinition`

### LINKED ISSUE #1010: Rename `ParsedDefinition` to `PreprocessedDefinition`

hgrecco:
Close #1010

- [x] Closes #1010 (insert issue number)

### DIFF
--- a/pint/definitions.py
+++ b/pint/definitions.py
@@ -15,8 +15,8 @@ from .errors import DefinitionSyntaxError
 from .util import ParserHelper, UnitsContainer, _is_dim
 
 
-class ParsedDefinition(
-    namedtuple("ParsedDefinition", "name symbol aliases value rhs_parts")
+class PreprocessedDefinition(
+    namedtuple("PreprocessedDefinition", "name symbol aliases value rhs_parts")
 ):
     """Splits a definition into the constitutive parts.
 
@@ -125,7 +125,7 @@ class Definition:
 
         Parameters
         ----------
-        definition : str or ParsedDefinition
+        definition : str or PreprocessedDefinition
 
         Returns
         -------
@@ -133,7 +133,7 @@ class Definition:
         """
 
         if isinstance(definition, str):
-            definition = ParsedDefinition.from_string(definition)
+            definition = PreprocessedDefinition.from_string(definition)
 
         if definition.name.startswith("@alias "):
             return AliasDefinition.from_string(definition)
@@ -184,7 +184,7 @@ class PrefixDefinition(Definition):
     @classmethod
     def from_string(cls, definition):
         if isinstance(definition, str):
-            definition = ParsedDefinition.from_string(definition)
+            definition = PreprocessedDefinition.from_string(definition)
 
         aliases = tuple(alias.strip("-") for alias in definition.aliases)
         if definition.symbol:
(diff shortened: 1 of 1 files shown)


keewis:
will you also move `PreprocessedDefinition.from_string` to `Definition.preprocess`, or will that be another PR?

hgrecco:
I tried that and it does not look nice. Here is why. Basically you end up having
```python
class Definition:

    @classmethod
    def preprocess(cls, definition):
        name, definition = definition.split("=", 1)
        name = name.strip()

        rhs_parts = tuple(res.strip() for res in definition.split("="))

        value, aliases = rhs_parts[0], tuple([x for x in rhs_parts[1:] if x != ""])
        symbol, aliases = (aliases[0], aliases[1:]) if aliases else (None, aliases)
        if symbol == "_":
            symbol = None
        aliases = tuple([x for x in aliases if x != "_"])

        return PreprocessedDefinition(name, symbol, aliases, value, rhs_parts)
```

which is then used (for example), here:

```python
class UnitDefinition(Definition):

    def from_string(cls, definition):
        if isinstance(definition, str):
            definition = cls.preprocess(definition)

        # here more code
```

What I do not like is the following:
1. `preprocess` becomes overridable at each `Definition` subclass. But there is no need because there is only one way to write a definition string that can be parsed into a `PreprocessedDefinition`.
2. Even if there were multiple ways, you will still need `PreprocessedDefinition` (or similar) to know to which `Definition` subclass to dispatch.
3. You need use `Definition.preprocess` to create a `PreprocessedDefinition` which you might want for debugging purposes.

This could be fixed with a module level function. But I do not like because it seems disconnected.

Therefore, I think that a classmethod constructor at `PreprocessedDefinition` is the best and `from_string` seems to be the best name mimicking the ones at definition and other examples as for example here: 

https://stackoverflow.com/questions/51943892/how-to-define-constructor-for-pythons-new-namedtuple-type

hgrecco:
So I will merge this and if we get any better solution or rationale we can still fix it later without much concern as this is part of the internal API.

keewis:
First of all, I don't feel strongly about this and I don't really disagree with the points you raised, so if what I'm suggesting does not convince you, feel free not to use it (you will end up looking at the code way more than me).

---

My itch with `ParsedDefinition.from_string` or `PreprocessedDefinition.from_string` was that the name feels somewhat redundant and from the name of the class I'd expect to call something like a `parse` or `preprocess` function or method to create it from a string (for debugging, you might also use the normal constructor of `namedtuple` that does not do the parsing).

If you agree with that, the question is then where to put such a function and how to name it. We seem to have two options: module-level and as a method of `Definition`.

Module-level has the advantage that it is independent of classes, so normally I'd go with that, as long as the name (including the way it is called) makes it clear what it does. Since you feel `definitions.preprocess` is not ideal, especially if you tend to call it just by name: `preprocess` instead of `definitions.preprocess`, what about `definitions.preprocess_definition`?

My original suggestion was to use a `staticmethod` in `Definition`:
```python
class Definition:
    @staticmethod
    def preprocess(string):
        # ...
        return PreprocessedDefinition(name, symbol, aliases, value, rhs_parts)
```
which has then to be used with `definitions.Definition.preprocess` (or `self.preprocess` or `cls.preprocess`). We should make it a `staticmethod` since we don't use `cls` (or `self`) in the function. Of course now you can override it in subclasses, but you don't have to (I might be wrong, but I think this is normal in OOP).

hgrecco:
I understand the rationale and I do not completely disagree with you. But I have decided to defer this to release 0.11. I would like to have a month of testing before we release 0.12 which will hopefully address the decimal issue.

PR #1056, opened 2020-03-13
Fixing bug when multiplying quantity by dict

### LINKED ISSUE #1032: Multiplication of quantity by dictionary-like object
I want to multiply a `quantity` by a `dict`-like object which has its own multiply defined. I would like for `pint` to return a `NotImplemented` when it tries to multiply a `quantity` by an object of my type, forcing the multiply to be done using my object's multiply. Here's what currently happens:

```python
import collections
import pint

ureg = pint.UnitRegistry()
q = 5 * ureg.s
c = collections.Counter({1:2, 3:4})
q * c
AttributeError: 'NotImplementedType' object has no attribute '_REGISTRY'
```
I understand that `collections.Counter` doesn't have its own multiply defined, but the error I get is the same if I attempt to multiply the `quantity` by my object with the `quantity` on the left (if the `quantity` is on the right, my object's multiply is used). The actual error here occurs in `pint.quantity.ireduce_dimensions`, when the result of calling the input function `f` is `NotImplemented`.

bmlivin:
- [ x ] Closes #1032
- [ x ] Executed ``black -t py36 . && isort -rc . && flake8`` with no errors
- [ x ] The change is fully covered by automated unit tests
- [ x ] Added an entry to the CHANGES file


bmlivin:
I can't see the details of the bors check (I get a permission denied page). Is there something I should change because it failed?

hgrecco:
There is a conflict with CHANGES likely because another merge happend before.

PR #1057, opened 2020-03-14
Changed to relative import in 2 positions

eevleevs:
I suggest to use only relative imports, so pint can also work as a subpackage.


hgrecco:
Please run `black -t py36 . && isort -rc . && flake8`

PR #1059, opened 2020-03-19
implement quantile

keewis:
This implements `numpy.quantile` and `numpy.nanquantile`.

### DIFF
--- a/pint/numpy_func.py
+++ b/pint/numpy_func.py
@@ -260,7 +260,10 @@ def implement_func(func_type, func_str, input_units=None, output_unit=None):
 
     # Handle functions in submodules
     func_str_split = func_str.split(".")
-    func = getattr(np, func_str_split[0])
+    func = getattr(np, func_str_split[0], None)
+    # If the function is not available, do not attempt to implement it
+    if func is None:
+        return
     for func_str_piece in func_str_split[1:]:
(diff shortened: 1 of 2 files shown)


keewis:
Is there a reason the output in `numpy.ipynb` was not stripped? If the notebook does not need any manual interaction, `nbsphinx` should be able to execute it but it won't if the notebook is "pre-executed" (if I didn't misunderstand anything).

The `getattr` guard is necessary because `numpy=1.14` does not provide `quantile` / `nanquantile`.

jthielen:
@keewis The only reason for the output not being stripped was that when I had tested it out, the sphinx build only rendered what was in the notebook rather than doing any execution. But, perhaps I had something configured incorrectly, so feel free to update that if you can get it working!

keewis:
turns out I added the `getattr` guard to the wrong function. Should I undo that?

notebook: It does work once you cleared every output (although the graphviz output looked cropped). Since there are quite a few new dependencies in `numpy.ipynb`, we would need to expand `requirements_docs.txt` a lot. That seems a bit out of scope for this PR, though.

keewis:
this is ready for review and merge.

The docs update for this PR should be automatic (it is not right now, but that is for a new PR to fix), so nothing to do here.

PR #1106, opened 2020-06-08
Fix exponent superscripting in HTML

clarkgwillison:
- [ ] Closes # (no issue filed)

This PR fixes two issues with exponent rendering of decimal quantities. The first was causing units with exponents that were longer than 1 character not to superscript properly in HTML views, such as used by Jupiter, and the second was causing rendered exponents in the terminal to look less than ideal.

## HTML View
In electronics, noise density is often expressed in units of Volts per sqrt(Hertz), but prior to this change that unit rendered like so:

image

By adding brackets around the exponent in the format string, we achieve the following:

image

## Terminal Pretty Print
Using the unicode dot operator instead of a plain period (.) makes decimal exponents looks slightly more natural.

Before:
```
nV/Hz⁰.⁵
```
After:
```
nV/Hz⁰⋅⁵
```

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -37,6 +37,7 @@ Pint Changelog
 - Replace pkg_resources.version to importlib.metadata.version. (Issue #1083)
 - Fix typo in docs for wraps example with optional arguments. (Issue #1088)
 - Add momentum as a dimension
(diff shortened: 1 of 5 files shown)


clarkgwillison:
@hgrecco - I think this is ready for review/merge, thoughts?

PR #1119, opened 2020-06-18
synchronize the Quantity.prod method with np.prod(q)

### LINKED ISSUE #1118: synchronize Quantity.prod with np.prod(q)
Continuing from #867: #1087 fixed the calculated units for `prod` when the `axis` or `where` (not both) parameters are passed, but only for the ufunc and not for the method (also, right now this will calculate the units using input and output sizes, but instead we could just use `q.shape[axis]`).

I looked into implementing this but it seems this is somewhat complicated: the way it works right now, `__getattr__` will get the method from the magnitude and wrap it using `_numpy_method_wrap` which in turn will call `get_op_output_unit`. To make this work I think we'd need to move the units calculation of `_prod` into a new function that accepts the shape and the args (`axis` and `where`) and try to modify `get_op_output_unit` to accept functions in addition to names. @jthielen, thoughts?

keewis:
As suggested in #1118, this makes `q.prod()` an alias of `np.prod(q)`.

- [x] Closes #1118

### DIFF
--- a/pint/numpy_func.py
+++ b/pint/numpy_func.py
@@ -416,7 +416,6 @@ matching_input_copy_units_output_ufuncs = [
 copy_units_output_ufuncs = ["ldexp", "fmod", "mod", "remainder"]
 op_units_output_ufuncs = {
     "var": "square",
-    "prod": "size",
     "multiply": "mul",
     "true_divide": "div",
     "divide": "div",
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -113,6 +113,9 @@ def check_implemented(f):
 
 
 def method_wraps(numpy_func):
+    if isinstance(numpy_func, str):
+        numpy_func = getattr(np, numpy_func, None)
+
     def wrapper(func):
         func.__wrapped__ = numpy_func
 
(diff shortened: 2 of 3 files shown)


keewis:
hmm, it seems `numpy==1.14` used `numpy.ndarray.prod` to implement `numpy.prod`, so this change would make this recursive. Also, while for `np.prod` we require the array function protocol (which by default is only activated for `numpy>=1.17`) we currently don't do that for `Quantity.prod`.

Any ideas on how to solve this except from bumping the `numpy` version?

hgrecco:
In version 0.10 we stated that we are going to adhere to [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html). So that means that only on Jan 13, 2021 we can drop support for Numpy 1.16 (initially released on Jan 13, 2019). 

However, that does not mean that certain specific features require specific numpy versions. My suggestion would be to support it for Numpy >= 1.17 and raise a nice exception otherwise. 

My suggestion

keewis:
@hgrecco, done. I'm not sure about the exception class, though. Is `NotImplementedError` the appropriate class?

jthielen:
@keewis If I recall correctly, this should be supported just fine on NumPy 1.16 so long as array function support is enabled, but also may not be supported on NumPy 1.17 if support is disabled. So, would it make sense to tie the check to that instead of just NumPy version? For what it's worth, `NotImplementedError` sounds correct to me.

keewis:
Sure. After we dropped `1.14` and `1.15`, we would probably need to rework on the tests, though: `where` has been introduced in `1.16`

PR #1123, opened 2020-06-21
make __eq__ always return a duck array

### LINKED ISSUE #351: Array comparisons return a single boolean if arrays are 0
When I compare two numpy array quantities that are both all zeros, I get a single boolean value, but would expect to get an _array_ of booleans.

 **Steps to reproduce**

``` python
x = np.zeros(4) * ureg.meter
y = np.zeros(4) * ureg.meter
return x==y
```

returns `True`

**Expected behavior**<br>
As described in #75, the return value should be `array([ True,  True,  True,  True], dtype=bool)`
We get the expected behavior if the arrays are non-zero, e.g.,

``` python
x = np.ones(4) * ureg.meter
y = np.ones(4) * ureg.meter
x==y
```

This is with Pint 0.7.2 installed from PyPI, Python 2.7.6 [GCC 4.8.2] installed with apt-get on Ubuntu 14.04. I've also observed this with pint 0.6.

keewis:
When comparing quantities with different units / dimensionalities but containing duck arrays for equality, the returned value tends to change between boolean arrays and python boolean objects depending on the input. This makes it somewhat difficult to write code / tests against it and with this I'd propose to change that to always return arrays if we're using array magnitudes.

I'm not quite sure where `bool_result` should live, maybe `compat.py`?

- [x] Closes #351

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -1470,11 +1470,16 @@ class Quantity(PrettyIPython, SharedRegistryObject):
     @check_implemented
     def __eq__(self, other):
         def bool_result(value):
+            nonlocal other
+
             if not is_duck_array_type(type(self._magnitude)):
                 return value
 
-            shape = np.broadcast(self._magnitude, other).shape
-            return np.full(shape=shape, fill_value=False)
+            if isinstance(other, Quantity):
+                other = other._magnitude
+
+            template, _ = np.broadcast_arrays(self._magnitude, other)
+            return np.full_like(template, fill_value=False)
 
         # We compare to the base class of Quantity because
         # each Quantity class is unique.
@@ -1498,6 +1503,7 @@ class Quantity(PrettyIPython, SharedRegistryObject):
 
             return bool_result(False)
 
+        # TODO: this might be expensive. Do we even need it?
         if eq(self._magnitude, 0, True) and eq(other._magnitude, 0, True):
             return bool_result(self.dimensionality == other.dimensionality)
 
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -1469,6 +1469,13 @@ class Quantity(PrettyIPython, SharedRegistryObject):
 
     @check_implemented
     def __eq__(self, other):
+        def bool_result(value):
+            if not is_duck_array_type(type(self._magnitude)):
+                return value
+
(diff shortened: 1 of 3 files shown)


pint/quantity.py
```diff
@@ -1486,12 +1498,14 @@ def __eq__(self, other):
...
-            )
+            if self.dimensionless:
+                return eq(self._convert_magnitude(self.UnitsContainer()), other, False)
 
+            return bool_result(False)
+
+        # TODO: this might be expensive. Do we even need it?
         if eq(self._magnitude, 0, True) and eq(other._magnitude, 0, True):
-            return self.dimensionality == other.dimensionality
+            return bool_result(self.dimensionality == other.dimensionality)
```
keewis, line 1508: I can't figure out what this block is for. I tried removing the whole block and rerun the tests to find out why this is necessary, but this didn't break any tests. Since this is potentially expensive (we might end up comparing the values multiple times on huge arrays) I'd suggest removing this:
```suggestion
```

hgrecco:
While I am in principle ok about making the output more predictable and easy to understand, I think it would be good to put here (or in an issue) some examples of current and expected behavior so more people can be aware of this (breaking) change.

keewis:
okay, then here are the examples (mostly from the test cases):
```python
u = np.ones((10,)) * ureg.dimensionless
v = np.zeros_like(u) * ureg.m
w = np.ones_like(u) * ureg.m

# compare a quantity with itself
u == u  # array-like, same as before
# compare dimensionless with non-quantity
u == 1  # array-like, before this was a single bool
# compare quantity with dimensionality with non-quantity
v == 1  # array-like, before this was a single bool

# compare quantities with different units and with the values being all 0
np.zeros_like(u) * ureg.m == np.zeros_like(u) * ureg.s
# array-like, before this returned a single bool

# compare two quantities with equal units
v == w  # array-like, same as before
# compare two quantities with compatible units
v == w.to("mm")  # array-like, same as before
# compare quantities with different units and different values
u == v  # array-like, before this was a single bool
```
as a summary: if the other quantity has a different unit the behaviour changed. Before it returned a single bool, now it returns an array filled with that bool.

hgrecco:
This looks good to me and there has not been any comment. I am merging so it can be included in the next release

bors r+

PR #1125, opened 2020-06-28
Workaround pintpandas iter issue

andrewgsavage:
- [ ] Executed ``black -t py36 . && isort -rc . && flake8`` with no errors

Possible work around for the iter issue
Drops the failing tests from 
====== 21 failed, 329 passed, 29 xfailed, 2 xpassed, 6 warnings in 21.44s ======
to 
====== 4 failed, 346 passed, 28 xfailed, 3 xpassed, 11 warnings in 17.55s ======

https://github.com/hgrecco/pint-pandas/issues/33
see also
https://github.com/hgrecco/pint/issues/753


znicholls:
Super nice @andrewgsavage !

hgrecco:
awesome

hgrecco:
There are a few failing tests.

znicholls:
> There are a few failing tests

One thing I noticed, we'll need a hack to workaround e.g. `Quantity("1", "m")` because strings are iterable :(

andrewgsavage:
> (Tiny typo) magnitide —> magnitude
haha ty, didn't notice that when debugging and was confused

OK sounds like you're happy with this approach, will try to get this working.

hgrecco:
I have to say that I was surprised that it worked. A few years ago I tried doing this with the `__str__` method but it did not work. But I think that `__iter__` is different.

Can you also try in this PR to change the imports from `pintpandas` to `pint_pandas` (I posted the change here https://github.com/hgrecco/pint-pandas/issues/1)

hgrecco:
I was trying to fix the pending issues without success. Likely because Magic methods like __iter__ are looked up on the class, not the instance as described [here](https://stackoverflow.com/questions/19508548/class-with-iter-assigned-in-constructor-not-recognized-as-iterator)

PR #1129, opened 2020-07-03
Implement Dask collection interface

### LINKED ISSUE #883: Integration with Dask (add tests; implement the Dask collection interface on Quantity)

rpmanser:
This pull request implements the [Dask collection interface](https://docs.dask.org/en/latest/custom-collections.html) for the Quantity class and adds convenience methods `compute()`, `persist()`, and `visualize()`. As is, only the convenience methods are wrapped to check that the magnitude of the Quantity is a dask array, which should cover current use cases. Ping @jthielen since we've been discussing this.

I have not tested `persist()` on a HPC cluster, which will have different behavior than when it is called on a single machine (e.g., desktop). Is there a way to integrate tests for distributed cases? If not, I can test this independently if needed.

- [X] Closes #883 
- [X] The change is fully covered by automated unit tests
- [X] Documented in docs/ as appropriate

### DIFF
--- a/pint/testsuite/test_dask.py
+++ b/pint/testsuite/test_dask.py
@@ -4,9 +4,16 @@ import pytest
 
 from pint import UnitRegistry
 
-# Conditionally import NumPy and Dask
+# Conditionally import NumPy, Dask, and Distributed
 np = pytest.importorskip("numpy", reason="NumPy is not available")
 dask = pytest.importorskip("dask", reason="Dask is not available")
+distributed = pytest.importorskip("distributed", reason="Distributed is not available")
+
+from dask.distributed import Client  # isort:skip
+from distributed.client import futures_of  # isort:skip
+from distributed.utils_test import cluster, gen_cluster, loop  # isort:skip
+
+loop = loop  # flake8
 
 ureg = UnitRegistry(force_ndarray_like=True)
 units_ = "kilogram"
@@ -26,26 +33,16 @@ def numpy_array():
     return np.arange(0, 25, dtype=float).reshape((5, 5)) + 5
 
 
-@pytest.mark.parametrize("component", ["__dask_graph__", "__dask_keys__"])
-def test_has_dask_components(dask_array, component):
-    """Test that a pint.Quantity wrapped Dask array has a Dask graph and keys"""
+def test_is_dask_collection(dask_array):
+    """Test that a pint.Quantity wrapped Dask array is a Dask collection."""
     q = ureg.Quantity(dask_array, units_)
-
-    q_method = getattr(q, component)
-    component_q = q_method()
-
-    truth_method = getattr(dask_array, component)
-    component_truth = truth_method()
-
-    assert component_q == component_truth
+    assert dask.is_dask_collection(q)
 
 
-def test_has_no_dask_graph(numpy_array):
-    """Test that a pint.Quantity wrapped NumPy array does not have a Dask graph
-    and that attempting to access it returns None.
-    """
+def test_is_not_dask_collection(numpy_array):
+    """Test that other pint.Quantity wrapped objects are not Dask collections."""
     q = ureg.Quantity(numpy_array, units_)
-    assert q.__dask_graph__() is None
+    assert not dask.is_dask_collection(q)
 
 
 def test_dask_scheduler(dask_array):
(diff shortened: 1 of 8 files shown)


keewis:
`xarray`'s [distributed tests](https://github.com/pydata/xarray/blob/b62a3fb2e96e6d88bb866453958a902fd782adc8/xarray/tests/test_distributed.py) might be useful as an example? They use the `loop` fixture and the `Cluster` and `Client` context managers to set up a fake distributed system.

rpmanser:
I'll take a look and see if I can setup something similar. Thanks!

pint/testsuite/test_dask.py
```diff
@@ -0,0 +1,130 @@
...
+    """Test exception handling for calling visualize() on a pint.Quantity wrapped object
+    that is not a dask.array.core.Array object.
+    """
+    q = ureg.Quantity(numpy_array, units_)
+    comps = add_five(q)
+    with pytest.raises(AttributeError) as excinfo:
+        comps.visualize()
+
+    exctruth = "Method visualize only implemented for objects of <class 'dask.array.core.Array'>, not <class 'numpy.ndarray'>"
+    assert str(excinfo.value) == exctruth
```
keewis, line 130: you can use `pytest.raises(AttributeError, match=exctruth)` to have `raises` compare the message. Also, could you split `exctruth` into shorter chunks? I.e.
```python
exctruth = (
    "Method visualize only implemented for objects of"
    " <class 'dask.array.core.Array'>, not"
    " <class 'numpy.ndarray'>"
)
```

pint/testsuite/test_dask.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def test_visualize():
+    pass
+
+
+def test_compute_exception(numpy_array):
+    """Test exception handling for calling compute() on a pint.Quantity wrapped object
+    that is not a dask.array.core.Array object.
+    """
```
keewis, line 97: All these exception tests have the same body (except from the docstring). You could try using `pytest.mark.parametrize` in combination with `getattr` to avoid the duplicated code.

Keep in mind that this is merely a suggestion, not a request, and if you decide that removing the duplicated code is less important than keeping the docstring that's fine, too.

pint/testsuite/test_dask.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def test_dask_optimize(dask_array):
+    """Test that a pint.Quantity wrapped Dask array can be optimized."""
+    q = ureg.Quantity(dask_array, units_)
+
+    assert q.__dask_optimize__ == dask.array.Array.__dask_optimize__
+
+
+def test_compute(dask_array, numpy_array):
```
keewis, line 65: as with the `exception` tests, the `compute` and `persist` tests can be rewritten to use `pytest.mark.parametrize` and `getattr`.

rpmanser:
I made all of the suggested changes and added tests for visualize and explicitly checking that `compute()` and `persist()` return the same results for the single machine case. I will be working on distributed tests next.

hgrecco:
This is looking really good, please feel free to ping me when you want me to review it and merge it. 

I would like to take the opportunity to ask a question: When are we going to provide an external package (e.g. `pint-pandas`) and when are we going to include it in `pint`. I am not suggesting to create `pint-dask` but rather to start building a rationale about pint extensions.

dopplershift:
My concern with extension packages is that they won't necessarily receive the same level of maintenance as the core pint library. As a maintainer of a downstream library, I'm always concerned about adding more dependencies.

What do people see as the benefits of having external packages?

jthielen:
@rpmanser As far as implementing the collection interface and testing it, this looks great! One small suggestion I would have is referring to https://github.com/dask/dask/issues/4583 in the NumPy doc example, since it is Dask holding us back from the usual "multiply on the right" construction. Also, I think this gets enough of #883 taken care of to consider it closed, but I think a follow-up issue should be raised as a reminder to finish up tests once https://github.com/dask/dask/issues/4583 is resolved (masked arrays and https://github.com/numpy/numpy/issues/15200 are in the same position).

One bigger-picture question/discussion point: in https://github.com/pydata/xarray/issues/525 and https://github.com/hgrecco/pint/issues/883 (and in other discussions), I was thinking of Pint Quantities "acting like a Dask Array" or "being a duck Dask Array" would occur *if and only if* the Quantity is wrapping a Dask Array. That is, a library like xarray could recognize a suitable Dask Array-like by a test like `is_duck_array(data) and hasattr(data, "__dask_graph__")` or similar, and this test could be `True` if the magnitude of the Quantity was a Dask Array and `False` if not. However, with the implementation here, a Pint Quantity will always be a duck Dask Array since the Dask interface methods will always be available (though of course not always usable). Also, in seeing the implementation here, I've realized that a hacky implementation of these methods through attachment in `__new__` or routing through `__getattr__` would be needed to achieve my initial idea (rather than a simple decorator check).

So, with that in mind, is it okay for xarray (and any other libraries depending on the duck Dask Array interfacing of Quantity) to have Quantity *always* be a duck Dask Array, or do these interface methods need to be "hidden" and only available when the magnitude of a Quantity is a Dask Array? While there ostensibly is another option (a "DaskQuantity" subclass that has these methods available), I think that would make the situation worse, since too much magic/special casing may be required to get a DaskQuantity as the result of arbitrary operations between Dask Arrays and regular Quantities.

rpmanser:
This is ready for your review @hgrecco. Looks like the travis CI has been failing due to the `-rc` flags given to `isort`, but otherwise `black -t py36 . && isort -rc . && flake8` runs without error.

dopplershift:
@jthielen In the absence of documentation on some kind of protocol, this might be one of those things that has to be worked out in practice. Having said that, personally I don't consider `__getattr__` a hack for this use case.

jthielen:
> @jthielen In the absence of documentation on some kind of protocol, this might be one of those things that has to be worked out in practice. Having said that, personally I don't consider `__getattr__` a hack for this use case.

Sounds good. I've also asked over in xarray (https://github.com/pydata/xarray/issues/4208), so we'll see if we get any feedback there.

hgrecco:
Congratulations @rpmanser for the awesome work!

rpmanser:
> Congratulations @rpmanser for the awesome work!

Thanks, happy I could help!

Issue #765, opened 2019-02-22
Elide large Latex-formatted arrays


natezb:
Semi-related to #654.

Jupyter will stall for several seconds when trying to display quantities containing large numpy arrays, since it tries to display and render the whole thing. It might be nice to elide this display beyond a certain practical size, like 1000 elements.

jondoesntgit:
I know that tqdm has some features that detect if it’s in a Jupyter notebook or not.

Numpy and pandas both display ellipses if the array is larger than 60 elements.

natezb:
I don't think that this has to be Jupyter-specific, but it is where I notice the issue, since I otherwise don't tell pint to generate LaTeX. It seems reasonable to me to always elide long arrays, regardless of environment.

hgrecco:
I think that the right way would be to mimic (or use) numpuy formatting options. Recently #759 was merged. I think we can use it here as well

hgrecco:
Any PR for this?

ipcoder:
I would like to emphasize that this is quite an issue for practical use of the package.
Thanks

hgrecco:
Linking to #1757

PR #1133, opened 2020-07-14
Change HTML repr away from LaTeX to a Sparse/Dask-like repr

### LINKED ISSUE #654: Turn off latex rendering in jupyter notebooks
I have been wrapping some pretty large numpy arrays with pint lately. Unfortunately, viewing these arrays with mathjax rendering ends up crashing my web browser. I searched the docs for a way to turn off the default rendering, but I could not find anything. Is there an easy way to do this?

jthielen:
While I've never been a big fan of Pint's LaTeX repr in JupyterLab/Jupyter Notebook due to working with large arrays that would crash the renderer, I've recently been working a lot with various wrapping combinations of xarray, Pint, Dask, and Sparse, in which Pint's LaTeX repr performs particularly poorly. Even if Pint's repr were elided (#765), it would still poorly nest with the other HTML reprs.

This motivated me to implement an alternative HTML repr for Pint using a simple HTML table similar to what both Dask and Sparse use. As shown in [this demonstration notebook](https://nbviewer.jupyter.org/urls/dl.dropbox.com/s/jlgqm6s92kzvagi/pint_html_repr_demo.ipynb), this allows much better nesting of HTML reprs. A quick example image is below:

This doesn't quite take care of [xarray's unit repr discussions](https://github.com/pydata/xarray/issues/2773) (since this only shows up when expanding the data in a dataset), but I still think it is a worthwhile improvement for the expanded view. @keewis do you have any thoughts on this particular interaction?

Marking as a WIP/draft for now to see if this is a welcome format for the HTML repr first before I dig too much into optimizing the implementation and writing tests for it.

- [x] Closes #654 (even though it was already closed in favor of #765, I think the are related but separate issues)

Tagging @nbren12, @hgrecco, @jondoesntgit, @natezb, @ipcoder, @keewis, and @TomNicholas for input based on past issues (#654, #765, and https://github.com/xarray-contrib/pint-xarray/issues/6).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ test/
 
 # test csv which should be user generated
 notebooks/pandas_test.csv
+
+# dask stuff
+dask-worker-space
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,9 @@ Pint Changelog
 0.15 (unreleased)
 -----------------
 
+- Change `Quantity` and `Unit` HTML (i.e., Jupyter notebook) repr away from LaTeX to a
+  simpler, more performant pretty-text and table based repr inspired by Sparse and Dask.
+  (Issue #654)
 - Implement Dask collection interface to support Pint Quantity wrapped Dask arrays.
 - Started automatically testing examples in the documentation
 - Fixed right operand power for dimensionless Quantity to reflect numpy behavior. (Issue #1136)
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -319,10 +319,10 @@ Pint also supports `f-strings`_ from python>=3.6 :
    >>> print(f'The str is {accel:~.3e}')
    The str is 1.300e+00 m / s ** 2
    >>> print(f'The str is {accel:~H}')      # HTML format (displays well in Jupyter)
-   The str is \[1.3\ m/{s}^{2}\]
+   The str is 1.3 m/s<sup>2</sup>
 
-But Pint also extends the standard formatting capabilities for unicode and
-LaTeX representations:
+But Pint also extends the standard formatting capabilities for unicode, LaTeX, and HTML
+representations:
 
 .. doctest::
 
@@ -335,7 +335,7 @@ LaTeX representations:
    'The LaTeX representation is 1.3\\ \\frac{\\mathrm{meter}}{\\mathrm{second}^{2}}'
    >>> # HTML print - good for Jupyter notebooks
    >>> 'The HTML representation is {:H}'.format(accel)
-   'The HTML representation is \\[1.3\\ meter/{second}^{2}\\]'
+   'The HTML representation is 1.3 meter/second<sup>2</sup>'
 
 If you want to use abbreviated unit names, prefix the specification with `~`:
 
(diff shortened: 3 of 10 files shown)


keewis:
this would definitely be a big improvement of the expanded HTML repr of xarray objects (I never really looked at the repr of quantities in a notebook so I can't comment on that), although the repr for big `numpy` arrays looks a bit strange (to fix that, I guess numpy needs its own `_repr_html_`?). :+1: for the new HTML repr

To properly format the collapsed or text repr we'd need to first modify `xarray`, so I guess we can ignore them for now.

jthielen:
Also, just to add a note here: assuming the `_repr_short_` or `_repr_inline_` of https://github.com/pydata/xarray/pull/4248 goes through as brought up by @keewis, that would be important to add to Pint here or in a follow-up PR.

jthielen:
Pinging @hgrecco for your thoughts on this in particular as maintainer, since it's been three weeks since the initial inquiry? Thanks!

keewis:
I've been wondering whether the `_repr_inline_` would be a good fit for `pint`, or whether it should be provided by `pint-xarray`. Until it is adopted by other packages this would be something specific to `xarray`, so it would probably be better to implement in `pint-xarray`. However, that would mean we're relying on monkeypatching, which is generally not a good idea. Thoughts?

jthielen:
@keewis That's a good point.

I wouldn't be too concerned about monkeypatching for something like `_repr_inline_`, since it is a single method that can be injected into the global Quantity class, and as you said, is just used in this one place for xarray at the moment. I guess what it comes down to for me is: do we expect all users of `xarray` with `pint` to be always using `pint-xarray`? If so, then it definitely seems reasonable to monkeypatch this in `pint-xarray`. If not, it could be surprising behavior to the unaware user to have "nice" inline reprs when `pint-xarray` is available/imported and "ugly" reprs when not.

In any case, since it isn't quite the simple thing I initially thought, I think it would be best to defer implementation of `_repr_inline_` to another PR (either direct implementation in Pint or monkeypatch in pint-xarray) so as to not cloud the more general HTML repr discussion here.

keewis:
agreed. I'd say let's implement this in `pint-xarray` for now and if desired we can move it to `pint` later.

jthielen:
Just wanted to ping @hgrecco again on this. I'll be working on some xarray & Pint integration docs for MetPy in the near future, and it would be great to have this in (or at have the general direction approved) by then.

hgrecco:
Sorry for the late reply,. pandemic times! I was also never really happy by the way pint represents this. So I agree with the change as long as an scalar is kept "normal" (as you also showed)!. Thanks for the PR (and pinging again!

jthielen:
> Sorry for the late reply,. pandemic times! I was also never really happy by the way pint represents this. So I agree with the change as long as an scalar is kept "normal" (as you also showed)!. Thanks for the PR (and pinging again!

Sounds good, I'll work on the tests and then mark as ready for review!

jthielen:
Hopefully I've finally gotten the uncertainties stuff fixed, so (assuming it is) this should be ready for review and merge.

hgrecco:
Thanks for this PR!

PR #1137, opened 2020-07-29
Avoid warning when setting a masked value

dopplershift:
When the underlying data is a masked array, __setitem__ can be given a sentintel np.ma.masked or masked values. This avoids passing them to math.isnan which causes a warning:
```
miniconda3/envs/py38/lib/python3.8/site-packages/pint/quantity.py:1782: UserWarning: Warning: converting a masked element to nan.
    if math.isnan(value):
```


dopplershift:
Rebased to address conflict in `CHANGES`.

PR #1147, opened 2020-08-16
Add case (in)senstive option to registry

### LINKED ISSUE #1145: Add case (in)senstive option to registry and extend to Unit and Quantity parsing
[In our quest to handle CF compliant/UDUNITS style units in MetPy](https://github.com/Unidata/MetPy/issues/1362), we have come across a couple issues regarding case sensitivity. For example, we need to be able to parse `Meter` as `meter` and [`Celcius` as an alias of `degrees_Celcius`](https://github.com/hgrecco/pint/issues/1081). To enable this within Pint (as an option of course and not the default [to maintain standards](https://github.com/hgrecco/pint/issues/1081#issuecomment-615238512)), I would like to propose the following to extend the case insensitive handling already existing in Pint:

1) Add a `case_sensitive` option to `BaseRegistry` which defaults to `True`
2) Modify the existing `case_sensitive` options on `get_name`, `parse_unit_name`, `parse_pattern`, and `parse_expression` to default to `None` instead of `True`, and have `None` be replaced within the methods with the registry's `case_sensitive` option
3) Add analogous `case_sensitive` kwarg to `parse_units`
4) Optional: add analogous `case_sensitive` kwarg to any other public registry method that ends up using `parse_unit_name` (or something that uses it, recursively) so that it can be appropriately passed along.

This would allow something like the following to become valid, so that a greater extent of CF unit compliance can be achieved:

```python
import pint

ureg = pint.UnitRegistry(case_sensitive=False)

q = ureg("1 Meter")

print(q)
```

```
1 meter
```

@hgrecco Is this compatible enough with your philosophy for Pint in https://github.com/hgrecco/pint/issues/1081#issuecomment-618681303 and seem like a reasonable enough way forward in implementation?

jthielen:
Implements the case sensitive registry option discussed in #1145

- [x] Closes #1145

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,8 @@ Pint Changelog
 - Change `Quantity` and `Unit` HTML (i.e., Jupyter notebook) repr away from LaTeX to a
   simpler, more performant pretty-text and table based repr inspired by Sparse and Dask.
   (Issue #654)
+- Add `case_sensitive` option to registry for case (in)sensitive handling when parsing
+  units (Issue #1145)
 - Implement Dask collection interface to support Pint Quantity wrapped Dask arrays.
 - Started automatically testing examples in the documentation
 - Fixed right operand power for dimensionless Quantity to reflect numpy behavior. (Issue #1136)
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -167,6 +167,8 @@ class BaseRegistry(metaclass=RegistryMeta):
         locale identifier string, used in `format_babel`
     non_int_type : type
         numerical type used for non integer values. (Default: float)
+    case_sensitive : bool, optional
+        Control default case sensitivity of unit parsing. (Default: True)
 
     """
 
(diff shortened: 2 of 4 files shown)


hgrecco:
This looks good. Just one question. Why you decided to put the 
```python
case_sensitive = (
            self.case_sensitive if case_sensitive is None else case_sensitive
        )
```

in `parse_unit_name` and not in `_parse_unit_name`?

jthielen:
> Why you decided to put the `...` in `parse_unit_name` and not in `_parse_unit_name`?

I believe that was because I started with it at the highest level (in each public function with the `case_sensitive` kwarg), and then realized that they all ended up being routed through `parse_unit_name`, so that was the only one needed. Indeed, I can move it to `_parse_unit_name` just as well. Let me know if you'd want me to move it there or not.

hgrecco:
I think it makes sense to put it in `_parse_unit_name` so that any future method that uses internal functions can rely on that and avoid code duplication.

jthielen:
> I think it makes sense to put it in `_parse_unit_name` so that any future method that uses internal functions can rely on that and avoid code duplication.

Sounds good, the change has been made. Thanks!

PR #1148, opened 2020-08-16
Update docs with guidelines on when to create extension packages

### LINKED ISSUE #1130: guidelines for when to create extension packages
In order to keep the `dask` / `distributed` PR (#1129) focused on that, let's continue the discussion here.

@hgrecco:
> I would like to take the opportunity to ask a question: When are we going to provide an external package (e.g. pint-pandas) and when are we going to include it in pint. I am not suggesting to create pint-dask but rather to start building a rationale about pint extensions.

@dopplershift:
> My concern with extension packages is that they won't necessarily receive the same level of maintenance as the core pint library. As a maintainer of a downstream library, I'm always concerned about adding more dependencies.
>
> What do people see as the benefits of having external packages?

First of all, external packages do have the advantage of a separate release cycle.

If I'm not missing any, all the extension packages we have right now (`pint-pandas` and `pint-xarray`) are for *wrapping* `pint`, not the other way around. While the `dask` interface is not really part of it, `pint` wrapping other libraries is usually done using the `numpy` interface (the only exception is `measurements` / `uncertainties`).

jthielen:
Implement the doc changes suggested in #1130, and updates `pint-xarray` link.

- [x] Closes #1130
- ~~Executed ``black -t py36 . && isort -rc . && flake8`` with no errors~~
- ~~The change is fully covered by automated unit tests~~
- ~~Added an entry to the CHANGES file~~

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -63,6 +63,32 @@ To build the documentation, invoke Sphinx from the ``docs`` directory::
     $ cd docs
     $ make html
 
+Extension Packages
+------------------
+
+Pint naturally integrates with other libraries in the scientific Python ecosystem, and
+a small number of
+`extension/compatibility packages<numpy.html#Compatibility-Packages>`_ have arisen to
+aid in compatibility between certain packages. Pint's rule of thumb for integration
+features that work best as an extension pacakage versus direct inclusion in Pint is:
+
+* Extension (separate packages)
+
+  * Duck array types that wrap Pint (come above Pint in
+    `the type casting hierarchy<numpy.html#Technical-Commentary>`_)
+
+  * Uses features independent/on top of the libraries
+
+  * Examples: xarray, Pandas
+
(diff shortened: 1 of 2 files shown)


docs/contributing.rst
```diff
@@ -63,6 +63,32 @@ To build the documentation, invoke Sphinx from the ``docs`` directory::
...
     $ make html
 
+Extension Packages
+------------------
+
+Pint naturally integrates with other libraries in the scientific Python ecosystem, and
+a small number of
+`extension/compatibility packages<numpy.html#Compatibility-Packages>`_ have arisen to
+aid in compatibility between certain packages. Pint's rule of thumb for integration
+features that work best as an extension pacakage verses direct inclusion in Pint is:
```
hgrecco, line 73: typo: versus

jthielen: Thank you catching that! Update has been made.

PR #1168, opened 2020-09-04
Mark code-block as python

nicoddemus:
Besides providing syntax-highlighting, I believe this might be the cause why it is not showing up at the docs currently.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -431,7 +431,7 @@ want to avoid creating multiple instances of the unit registry.
 The best way to do this is by instantiating the registry in a single place. For
 example, you can add the following code to your package ``__init__.py``
 
-.. code-block::
+.. code-block:: python
 
    from pint import UnitRegistry
    ureg = UnitRegistry()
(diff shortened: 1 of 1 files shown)


nicoddemus:
Screenshot of the current docs, which are missing the changed code blocks:

image

I assume this PR will fix it.

keewis:
I don't think that will fix it, the language identifier should only control syntax highlighting.

Actually, @hgrecco, the RTD builds are failing since quite a long time ago (6da2d66c4ae02160d46bb5a50fecdc44bf9d97b3 is the first failing commit). In order to check that without having to remember to regularly check either the [builds](https://readthedocs.org/projects/pint/builds/) or the badge in the readme, you could enable the PR feature of RTD? That way, each PR gets an additional check that makes sure the docs also build on RTD.

hgrecco:
It is weird because it was failing due to pint-pandas, which in turn was failing due to an older version of pandas. Somehow, it was not able to resolve properly the installation. I fixed this, and enforced pandas>1.0.4 and now is working. Thanks!

See  #1170

PR #1196, opened 2020-10-28
Fix scalar ndarray tolist function

### LINKED ISSUE #1195: `Quantity.tolist()` method fails for zero-dimenional arrays
# Description
As I understand, `Quantity.tolist()` is meant to mirror numpy's `ndarray.tolist()` method; it does this successfully with arrays of at least dimensionality 1, but throws an exception for 0-dimensional arrays (i.e., scalar ndarrays).

*Pint version: 0.16.1*

# Example
Some setup code:
```pycon
>>> import pint
>>> import numpy as np
```

When I run `tolist()` on a 1-dimensional ndarray, it works as expected:
```pycon
>>> a = np.arange(5)
>>> a.ndim
1
>>> a.tolist()
[0, 1, 2, 3, 4]
>>> (a * ureg.meter).tolist()
[<Quantity(0, 'meter')>, <Quantity(1, 'meter')>, <Quantity(2, 'meter')>, <Quantity(3, 'meter')>, <Quantity(4, 'meter')>]
```

But when I run `tolist()` on a 0-dimensional ndarray, it raises an exception:
```pycon
>>> a = np.array(5)
>>> a.ndim
0
>>> a.tolist()
5
>>> (a * ureg.meter).tolist()
  File "C:\___\venv\lib\site-packages\pint\quantity.py", line 1886, in tolist
    for value in self._magnitude.tolist()
TypeError: 'int' object is not iterable
```

jules-ch:
- [x] Closes #1195

Returning scalar Quantity if magnitude.tolist() is not Iterable.
Better AttributeError messaging.

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -1879,12 +1879,22 @@ class Quantity(PrettyIPython, SharedRegistryObject):
 
     def tolist(self):
         units = self._units
-        return [
-            self.__class__(value, units).tolist()
-            if isinstance(value, list)
-            else self.__class__(value, units)
-            for value in self._magnitude.tolist()
-        ]
+
+        try:
+            values = self._magnitude.tolist()
+            if not isinstance(values, list):
+                return self.__class__(values, units)
+
+            return [
+                self.__class__(value, units).tolist()
+                if isinstance(value, list)
+                else self.__class__(value, units)
+                for value in self._magnitude.tolist()
+            ]
+        except AttributeError:
+            raise AttributeError(
+                f"Magnitude '{type(self._magnitude).__name__}' does not support tolist."
+            )
 
     # Measurement support
     def plus_minus(self, error, relative=False):
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ Pint Changelog
 0.17 (unreleased)
 -----------------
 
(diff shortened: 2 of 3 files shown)


pint/quantity.py
```diff
@@ -1879,12 +1879,22 @@ def __setitem__(self, key, value):
...
+            return [
+                self.__class__(value, units).tolist()
+                if isinstance(value, list)
+                else self.__class__(value, units)
+                for value in self._magnitude.tolist()
+            ]
+        except AttributeError:
+            raise AttributeError(
+                f"Magnitude '{type(self._magnitude).__name__}' does not support tolist."
+            )
```
keewis, line 1897: Wouldn't it be much easier to just do
```suggestion
        values = self._magnitude.tolist()
        if not isinstance(values, list):
            values = [values]

        return [
            self.__class__(value, units).tolist()
            if isinstance(value, list)
            else self.__class__(value, units)
            for value in values
        ]
```
?

That way, we only call `tolist` once per dimension and the code is a little bit simpler

jules-ch: Yeah thanks, I knew performance wise calling twice wasn't that good thx :). True I'll make the changes

PR #1248, opened 2021-02-16
Fix numpy.linalg.solve units output.

### LINKED ISSUE #1246: numpy.linalg.solve returns wrong units
`numpy.linalg.solve` returns units that are inverted with respect to what is expected. E.g.

```python
A = np.random.rand(5, 5)*ureg.m
b = np.random.rand(5)*ureg.s
np.linalg.solve(A, b)
```

returns `<Unit('meter / second')>` for `x`, whereas `<Unit('second / meter')>` is expected, since the matrix equation

equation

expressed in units, corresponds to `meter * second / meter = second`.

(Note also, that the wrong behaviour is explicitly tested [here](https://github.com/hgrecco/pint/blob/b36d0a63ee24e7dada0cbf8b0499695ff57d17a8/pint/testsuite/test_numpy.py#L452).)

jules-ch:
Update get_op_output_unit with new type invdiv.
It outputs the product of the following units over the first one in the args list.

Update tests with values & np.dot(A, x) == b.
Where x = np.linalg.solve(A, b) just like numpy documentation https://numpy.org/doc/stable/reference/generated/numpy.linalg.solve.html

- [x] Closes #1246

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,7 @@ Pint Changelog
 - UnitsContainer returns false if other is str and cannnot be parsed
   (Issue #1179, thanks rfrowe)
 - Add Github Actions CI. (Issue #1236)
+- Fix numpy.linalg.solve unit output. (Issue #1246)
 
 0.16.1 (2020-09-22)
 -------------------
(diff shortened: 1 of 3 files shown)


hgrecco:
awesome, thanks!

jules-ch:
Seems like bors timed out

hgrecco:
Let's try again

bors r+

hgrecco:
I now the reason!, bors toml still points to travis. I will merge this manually.

bors r-

PR #1251, opened 2021-03-01
Add np.linalg.norm implementation

### LINKED ISSUE #1250: Missing np.linalg.norm implementation
Hi all,

Firstly, thank you to the authors & maintainers of `Pint`, it makes working with units both easy and fun.

I noticed there was no implementation for `np.linalg.norm` available:
```python
>>> import numpy as np
>>> print(np.__version__)
1.19.3
>>> import pint
>>> print(pint.__version__)
0.16.1
>>> q = np.array([[[3, 4], [5, 12], [8, 15]], [[3, 4], [5, 12], [8, 15]]]) * ureg.m
>>> np.linalg.norm(q, axis=2)
<ipython-input-6-4c2d06c14435> in <module>()
      5 ureg = pint.UnitRegistry()
      6 q = np.array([[[3, 4], [5, 12], [8, 15]], [[3, 4], [5, 12], [8, 15]]]) * ureg.m
----> 7 np.linalg.norm(q, axis=2)

<__array_function__ internals> in norm(*args, **kwargs)

TypeError: no implementation found for 'numpy.linalg.norm' on types that implement __array_function__: [<class 'pint.quantity.build_quantity_class.<locals>.Quantity'>]
```

As the feature seemed fairly straightforward from previous examples, I've implemented the `np.linalg.norm` routine ready for PR here.

arjavtrivedi:
- [x] Closes #1250 

Doc entry in _docs/numpy.ipynb_ already covered by print statement here:
```python
from pint.numpy_func import HANDLED_FUNCTIONS
print(sorted(list(HANDLED_FUNCTIONS)))
```

### DIFF
--- a/pint/testsuite/test_issues.py
+++ b/pint/testsuite/test_issues.py
@@ -817,6 +817,12 @@ class TestIssues(QuantityTestCase):
             np.array((0.04, 0.09)),
         )
 
+    @helpers.requires_numpy
+    def test_issue_1250(self):
+        q = np.array([[3, 4], [5, 12], [8, 15]]) * self.ureg.m
+        expected = np.array([5, 13, 17]) * self.ureg.m
+        helpers.assert_quantity_equal(np.linalg.norm(q, axis=1), expected)
+
 
 if np is not None:
 
--- a/CHANGES
+++ b/CHANGES
@@ -9,12 +9,13 @@ Pint Changelog
 - Fix comparisons between Quantities and Measurements.
    (Issue #1134, thanks lewisamarshall)
 - Implemented benchmarks based on airspeed velocity.
-- Fix tolist function with scalar ndarray. 
+- Fix tolist function with scalar ndarray.
   (Issue #1195, thanks jules-ch)
 - UnitsContainer returns false if other is str and cannnot be parsed
   (Issue #1179, thanks rfrowe)
 - Add Github Actions CI. (Issue #1236)
 - Fix numpy.linalg.solve unit output. (Issue #1246)
+- Add numpy.linalg.norm implementation. (Issue #1250)
 
 
 0.16.1 (2020-09-22)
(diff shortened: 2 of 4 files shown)


pint/testsuite/test_issues.py
```diff
@@ -817,6 +817,12 @@ def test_issue_1185(self):
             np.array((0.04, 0.09)),
         )
 
+    @helpers.requires_numpy
+    def test_issue_1250(self):
```
keewis, line 821: I would put this in `test_numpy.py` and call it `test_linalg_norm`, but otherwise this looks good to me.

keewis (approved):
looks good to me, but someone with the permissions would have to rerun the tests so we can investigate the failures (the logs have expired since then).

pint/testsuite/test_issues.py
arjavtrivedi: @keewis - thanks for the feedback. I've removed this test in favour of the implementation in `test_numpy.py` ([here](https://github.com/hgrecco/pint/blob/cc56b12ca5e07ea756a71847d30635f370415344/pint/testsuite/test_numpy.py#L1429)).

arjavtrivedi:
I've managed to come back to this PR after quite a while. As such, I've updated merged upstream changes and addressed the previous PR feedback. The actions on the forked repository ([here](https://github.com/arjavtrivedi/pint/actions)) appear to be passing, but otherwise not sure why the PR workflows here are failing.

AdVetter:
We are using Pint and `np.linalg.norm` a lot in our calculations and had to write helper a function to use it properly. Looking really forward to when this is merged! :-)

keewis:
@arjavtrivedi (or @hgrecco / @jules-ch), could you merge in `master`? That should get RTD to build, and we'd need a maintainer to once again approve the workflow runs. Once tests pass I think this should be ready.

arjavtrivedi:
Thank you @hgrecco for the great software!

keewis:
> unfortunately I don't have write access else I'd merge!

This is not important anymore since the PR was merged (but I guess is good to know for any future PRs). Still, what I meant was to merge `master` into *this* branch to make use of some changes in `master` that fixed RTD.

hgrecco:
@keewis I think that the current master has everything ok, right?

PR #1253, opened 2021-03-01
remove pintpandas import

### LINKED ISSUE #1252: Remove pint_pandas import from pint
I think we should remove the pint_pandas import from pint. 

This should prevent the circular import issues that are preventing a fix to https://github.com/hgrecco/pint-pandas/issues/37. This should simplify registry handling somewhat, so I think this is worth writing an extra line of code  `import pint_pandas`  (which is in 2/3 examples in the documentation anyway)

andrewgsavage:
- [x] Closes #1252 

### DIFF
--- a/docs/pint-pandas.ipynb
+++ b/docs/pint-pandas.ipynb
@@ -43,9 +43,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


hgrecco:
Maybe you can brifly comment while the import is no longer needed. It will be serve as future reference

andrewgsavage:
2/3 of the example uses require importing pint_pandas to access PintType or PintArray, so in many cases the import is still necessary.

Removing the import from pint will allow pint_pandas to use get_application_registry(), which should prevent issues caused by using multiple registries.

Issue #1166, opened 2020-08-29
Better annotations support


hgrecco:
With [PEP560](https://www.python.org/dev/peps/pep-0560/) we could now try to have a better annotations experience for Pint. Briefly, my proposal would be to do something like this

```python
class Model:

    value: Quantity['m/s']
```

or 

```python
class Model:

    value: Quantity['[length]/[time]']
```

and the provide to a nice API to check for this.

What do you think?

hgrecco:
and also these [examples](https://pint.readthedocs.io/en/stable/wrapping.html) would be:
```python
>>> @ureg.awrap
... def mypp(length: Quantity['meter]') -> Quantity['second]':
...     return pendulum_period(length)

```

and 

```python
>>> @ureg.acheck
... def pendulum_period(length: Quantity['[length]'):
...     return 2*math.pi*math.sqrt(length/G)
```

where `awrap`and `acheck` is are annotated equivalents to `wrap` and `check`.

jmuhlich:
It would also be fantastic to have Mypy support for checking these annotations statically! Happy to contribute where I can.

dopplershift:
I haven't started using annotations in MetPy (yet), so I don't have any practical experience to rely on to see any obvious gotchas. In general, though, those look reasonable.

hgrecco:
I was playing with this concept. Some things to discuss:

1. Can annotations be done with units (e.g. `m/s`) and dimensions (e.g. `[length]/[seconds]`)?
   Yes. As there are valid use cases for both (e.g. wrapping vs checking)
2. What is the output type of Quantity['m/s']?
  - A str? No.
  - A Quantity?
  - A new class (e.g. TypedQuantity)?
  - A UnitContainer or similar?
3. Can we annotate with `ureg.meter`?

jules-ch:
It would be nice to set the expected type `magnitude` should return either np.array, float or any supported types. It is sometimes confusing for the user to guess it when it is simply typed with Quantity.

claudiofinizio:
Hello,
I am writing a [webapp](https://claudiofinizio.github.io/ruralwater/) for designing rural water supplies and I made extensive use of both pint and mypy. I would therefore be **glad to contribute exposing Quantity to mypy**.

My objective is to write code like follows:

```
class WaterPipeline:
    @property`
    def get_pipeline_pathlength(self): -> Quantity['length']
        """"""
```
[...]

PR #1259, opened 2021-03-07
pint typing

jules-ch:
Following the discussion started here #1166

This PR introduces Type annotation support for Quantity.

Quantity is a Generic (https://docs.python.org/3/library/typing.html#user-defined-generic-types).

Magnitude type changes based on input type when constructing Quantity.
This will ease type checking for libraries & application using types and developers who wish to statically type check their code.

- Quantity as Generic Type
- Properly type annotate widely used functions from public API (Registry, Unit, Quantity)

If readability is becoming an issue we can export type annotations to stub files.

After the merge of this PR we should comply to [PEP561](https://www.python.org/dev/peps/pep-0561/).
We can also introduce [Annotated Quantity](https://github.com/hgrecco/pint/issues/1166#issuecomment-755754727) types with dimensionality & unit.

After running mypy, we will be able to track some bugs since we have a lot of type checking with `isinstance`, switches etc.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,19 @@
 [run]
 omit = pint/testsuite/*
 
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
+    # Have to re-enable the standard pragma
+    pragma: no cover
+
+    # Don't complain about missing debug-only code:
+    def __repr__
+
+    # Don't complain if tests don't hit defensive assertion code:
+    raise AssertionError
+    raise NotImplementedError
+    AbstractMethodError
+
+    # Don't complain if non-runnable code isn't run:
+    if TYPE_CHECKING:
--- a/CHANGES
+++ b/CHANGES
@@ -19,6 +19,8 @@ Pint Changelog
 
 - pint no longer supports Python 3.6
 - Minimum Numpy version supported is 1.17+
+- Add supports for type hints for Quantity class. Quantity is now a Generic (PEP560).
+- Add support for [PEP561](https://www.python.org/dev/peps/pep-0561/) (Package Type information)
 
 
 0.17 (2021-03-22)
--- a/pint/_typing.py
+++ b/pint/_typing.py
@@ -1,0 +1,18 @@
+from typing import TYPE_CHECKING, Any, Callable, Tuple, TypeVar, Union
+
+if TYPE_CHECKING:
+    from .quantity import Quantity
+    from .unit import Unit
+    from .util import UnitsContainer
+
+UnitLike = Union[str, "UnitsContainer", "Unit"]
+
+QuantityOrUnitLike = Union["Quantity", UnitLike]
+
+Shape = Tuple[int, ...]
+
+_MagnitudeType = TypeVar("_MagnitudeType")
+S = TypeVar("S")
+
+FuncType = Callable[..., Any]
+F = TypeVar("F", bound=FuncType)
--- a/pint/context.py
+++ b/pint/context.py
@@ -8,13 +8,21 @@
     :license: BSD, see LICENSE for more details.
 """
 
+from __future__ import annotations
+
 import re
 import weakref
 from collections import ChainMap, defaultdict
+from typing import TYPE_CHECKING, Any, Callable, Optional, Tuple
 
 from .definitions import Definition, UnitDefinition
 from .errors import DefinitionSyntaxError
 from .util import ParserHelper, SourceIterator, to_units_container
+
+if TYPE_CHECKING:
+    from .quantity import Quantity
+    from .registry import UnitRegistry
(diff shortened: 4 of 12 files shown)


jules-ch:
PR is almost ready.

```python

from pint import UnitRegistry
import numpy as np

ureg = UnitRegistry()

float_quantity = ureg.Quantity(150.0, "m**2")
int_quantity = ureg.Quantity(150, "m**2")
array_quantity = ureg.Quantity(np.array([1, 2, 3, 4]), "m**2")
list_quantity = ureg.Quantity((1, 2, 3, 4), "m**2")
str_quantity = ureg.Quantity("50*2 m")

sequence = [ureg.Quantity(2, "m"), ureg.Quantity(3, "m")]

array_quantity_2 = ureg.Quantity.from_list(sequence)

reveal_type(float_quantity)
reveal_type(int_quantity)
reveal_type(float_quantity.m)
reveal_type(float_quantity.to("km**2"))
reveal_type(array_quantity)

reveal_type(array_quantity_2)
reveal_type(str_quantity)
reveal_type(list_quantity)
reveal_type(list_quantity)
```

With mypy : 

```
note: Revealed type is 'pint.quantity.Quantity[builtins.float*]'
note: Revealed type is 'pint.quantity.Quantity[builtins.int*]'
note: Revealed type is 'builtins.float*'
note: Revealed type is 'pint.quantity.Quantity[builtins.float*]'
Revealed type is 'pint.quantity.Quantity[numpy.ndarray*]'
Revealed type is 'pint.quantity.Quantity[numpy.ndarray]'
Revealed type is 'pint.quantity.Quantity[Any]'
Revealed type is 'pint.quantity.Quantity[numpy.ndarray]'
```

jules-ch:
If some people can test this in their IDE & with mypy. I'd be grateful.

I had trouble working with dynamically generated classes (like quantity is), it seems to work with another TypeVar in the overloaded `__new__` constructor, I'll probably file issues in the mypy tracker.

For further steps I'll file an issue where we can discuss further steps.

- I'll link this also to the issue where we discussed Array sublassing (it can ease type hints)
- Annotated Types with PEP561
- Better mypy support will come with a plugin I think or with dedicated types just like numpy does with [DTypes](https://github.com/numpy/numpy/blob/336cda183332006ce57750af6b14642a79e3b3f4/numpy/typing/__init__.py#L252)
It might be really difficult to implement, so I don't know if we want to go into this rabbithole.

@hgrecco as I said if type hints is becoming a burden in the code, we can extract them to stub files (*.pyi) if needed

hgrecco:
Cool. What do you think about adding `from __future__ import annotations` to have a uniform typing experience among all python versions. This is not supported in Python 3.6 but I think we can drop it as it is not supported by current NumPy version, right?

keewis:
right, but that's not what `pint` has to look out for. Rather, NEP29 allowed dropping support for py36 about a year ago so this is definitely okay.

jules-ch:
I agree with @keewis , we can drop support for python 3.6 according to NEP 29.
Python 3.6 is going to the End of Life phase at the end of this year.
We can bump minimum numpy version aswell.

According to pypistats.org, python 3.6 represents roughly 10% of the download.

We can pin an issue announcing last compatible version & that version 0.18 will drop python 3.6 support. See if this raise flags.

hgrecco:
@keewis You are right. I actually looked at NEP29. 

@jules-ch I agree with the protocol you have proposed.

jules-ch:
I'm fixing CI with #1356, after that I think we can merge this.

Type hints will be experimental, I made typing module private.
I haven't had a proper answer to the issue I raised in mypy https://github.com/python/mypy/issues/10611, the workaround is working to have type inference on the constructor.

jules-ch:
@hgrecco It is ready for review

hgrecco:
This is awesome. Excelent job. My only question is the purpose of this

```python
if TYPE_CHECKING:
    from .quantity import Quantity
    from .registry import UnitRegistry
    from .util import UnitsContainer
```
in files like `context.py` My understanding is that when annotations is imported from future everyhing is treated as an string and therefore this should not be necessary. But my knowledge of typing is not so great. Can you comment on this?

jules-ch:
It is used to avoid circular imports.
Even with `__future__.annotations` import statements are still there & therefore circular imports can still throw errors

hgrecco:
I get tge circular imports, and I get that these imports are only used when typechecking. I was wondering if there was no other place to put it (i.e. a typing imports file which make somethings available). In any case, I really like this and I will be happy to merge it

conversy:
Hi,

sorry for what might appear as a very dumb question, but I couldn't find any doc on how to use this.
Could you provide me with some examples of annotations that could be verified with `mypy` please?
That would be very helpful, thanks in advance!

PR #1270, opened 2021-03-22
Do not install numpy in CI when not instructed to

jules-ch:
Fixing CI

When installing with `pip install .[test]`, this installs pytest-mpl which depends on matplotlib which depends on numpy.
We are then testing with numpy installed even though we have some tests that requires numpy not to be installed.

I added an extra step with pytest-mpl if matplotlib is in the `extras` dependencies


hgrecco:
Good catch

Issue #1273, opened 2021-03-23
ValueError: Cannot operate with Quantity and Quantity of different registries makes it challenging to use pint in packages


edmundsj:
I understand from reading the basic documentation that raising this ValueError when doing unit comparison between different registries is intended behavior. However, this makes pint extremely difficult to use as someone trying to write a package people will use in conjunction with pint. The example I am running into is the following:

Internally, I want to do some unit parsing. To do this, I use ureg.parse_expression() on the expression of interest. Works great, and returns a quantity. But, when someone tries to use my package (let's call it my_package) to pass in quantities of their own, I cannot compare those to internal units (or even parse them after converting them to strings). 

```
from my_package import Experiment # Internally uses pint for unit parsing
from pint import UnitRegistry
import numpy as np

ureg = UnitRegistry()
voltages = ureg.V * np.array([1, 2, 3])
exp = Experiment(voltage=voltages) # problem - internal and external use of pint are incompatible

```

This seems like a serious limitation of pint, and discourages me (as someone writing packages) from using it. I have to ask the user of my package not only to define their quantities of interest with pint but also when using those quantities to pass in their UnitRegistry so I can use it internally to do comparisons and arithmetic on their units. And I have to make sure, if the user does not pass in a unit registry, that does not break my package (which it will without some serious contortions).

hgrecco:
> I have to ask the user of my package not only to define their quantities of interest with pint

This part I do not get. If the user 

This is not only an intended behaviour but I would argue that is the correct behavior as is the only way to ensure correctnes. Pint is fully configurable so different packages might add , remove or modify units and constants. A single, shared registry is the way to go. 

To share a registry Pint has the concept of **application_registry** (that should be documented better!). Let me define "package" as the software component that only exposes functionality to be used by other packages / applications; and application as the software that the user actually uses. (it is bad definition but stay with me for a while)

The idea is that package writer/maintainer use this at their packages:

```python
ureg = pint.get_application_registry()
```

which will  return the current **application_registry** or return a default one if none was created.

The application writer/maintainer do the following at their init

```python
ureg = pint.UnitRegistry()
# configure the registry the way you like
pint.set_application_registry(ureg)

import other_package
```

In this way you get, loose coupling with as many packages as you want but also correctness.


> I have to ask the user of my package not only to define their quantities of interest with pint 

Here I do not follow you. The package user need to define the quantities in certain way, there is no way to avoid it (except using strings).

edmundsj:
[...]

PR #1275, opened 2021-03-24
Use application registry in docs

cpascual:
The application registry feature is not adequately advertised.
Improve this by using get_application_registry() instead of UnitRegistry()
throughout the docs (including README and docstrings) when that makes sense.
Also change the tutorial to recommend using it whenever possible.

See https://github.com/hgrecco/pint/issues/1273 for the motivation.


cpascual:
Hi, I am getting [some doctest errors](https://github.com/hgrecco/pint/pull/1275/checks?check_run_id=2183721477), but I am not sure if they are related to my changes.

hgrecco:
I am not sure that putting `ureg.get_application_registry` everywhere is a good idea but I might biased by my own history with pint. 

What I already know is that we should improve the correspondending [subsection](https://pint.readthedocs.io/en/stable/tutorial.html#using-pint-in-your-projects) (or even move it to its own section). This should include the rationale that I gave in #1273 and maybe a small example (e.g. project 1, project 2 ..)

Going back to the readme and docs, my fear is that we are changing the suggested way to use it too quickly and confuse existing users. My suggestions would be to do a gradual rollout. Example: we put the `get_application_registry` in the readme and cos as a note after we show how to instantiate the registry and I see how it plays off. But, as I said, I might be dead wrong so I would be happy to hear others peoples thoughts.

cpascual:
I completely defer it to your judgement, @hgrecco since I have only a very biased perspective on the use of pint. 

I created this PR just because it was simpler than trying to explain my proposed changes, but I don't mind at all if it is partially or completely edited.

My point of view: I like the "radical" change (as proposed) because the full benefit of the application registry comes, IMHO, when its use is generalized (i.e. when most pint-enabled projects play along well with the rest)

andrewgsavage:
the docs have been reorganised since this so I'll close it

PR #1301, opened 2021-05-10
Fix negative exponents with force_ndarray=True

### LINKED ISSUE #1295: force_ndarray is mutually exclusive with negative integer exponents
It seems that due to the way `parse_expression` works, using negative exponents with `force_ndarray=True` results in an error:
```python
In [2]: ureg = pint.UnitRegistry()
   ...: ureg("m ** -1")
Out[2]: 1.0 <Unit('1 / meter')>

In [3]: ureg2 = pint.UnitRegistry(force_ndarray=True)
   ...: ureg2("m ** -1")
ValueError                                Traceback (most recent call last)
<ipython-input-3-e15cfb26c53e> in <module>
      1 ureg2 = pint.UnitRegistry(force_ndarray=True)
----> 2 ureg2("m ** -1")

.../lib/python3.9/site-packages/pint/registry.py in parse_expression(self, input_string, case_sensitive, use_decimal, **values)
   1261         gen = tokenizer(input_string)
   1262 
-> 1263         return build_eval_tree(gen).evaluate(
   1264             lambda x: self._eval_token(x, case_sensitive=case_sensitive, **values)
   1265         )

.../lib/python3.9/site-packages/pint/pint_eval.py in evaluate(self, define_op, bin_op, un_op)
     91                 raise DefinitionSyntaxError('missing binary operator "%s"' % op_text)
     92             left = self.left.evaluate(define_op, bin_op, un_op)
---> 93             return bin_op[op_text](left, self.right.evaluate(define_op, bin_op, un_op))
     94         elif self.operator:
     95             # unary operator

.../lib/python3.9/site-packages/pint/quantity.py in wrapped(self, *args, **kwargs)
    113         elif isinstance(other, list) and other and isinstance(other[0], type(self)):
    114             return NotImplemented
--> 115         return f(self, *args, **kwargs)
    116 
    117     return wrapped

.../lib/python3.9/site-packages/pint/quantity.py in __pow__(self, other)
   1494                     units = new_self._units ** exponent
   1495 
-> 1496             magnitude = new_self._magnitude ** exponent
   1497             return self.__class__(magnitude, units)
   1498 

ValueError: Integers to negative integer powers are not allowed.
```
To be clear, that's a `numpy` issue: `np.array(1) ** -1` will raise the same error.

Not sure what to do about this: I would like to get this fixed somehow but this seems tricky (we seem to just evaluate the parsed syntax tree).

xref xarray-contrib/cf-xarray#197

cc @jthielen

keewis:
This is option 1 from https://github.com/hgrecco/pint/issues/1295#issuecomment-827791585. The only reason why I chose to implement this instead of any other option is that it is the smaller fix. If we decide on a different option or to do a combination I can modify this PR.

- [x] Closes #1295

### DIFF
--- a/pint/compat.py
+++ b/pint/compat.py
@@ -211,6 +211,10 @@ def is_duck_array_type(cls) -> bool:
     )
 
 
+def is_duck_array(obj):
+    return is_duck_array_type(type(obj))
+
+
 def eq(lhs, rhs, check_all: bool):
     """Comparison of scalars and arrays.
 
--- a/pint/pint_eval.py
+++ b/pint/pint_eval.py
@@ -25,8 +25,24 @@ _OP_PRIORITY = {
     "-": 0,
 }
 
+
+def _power(left, right):
+    from .compat import is_duck_array
+    from .quantity import Quantity
+
+    if (
+        isinstance(left, Quantity)
+        and is_duck_array(left.magnitude)
+        and left.dtype.kind not in "cf"
+        and right < 0
+    ):
+        left = left.astype(float)
+
+    return operator.pow(left, right)
+
+
 _BINARY_OPERATOR_MAP = {
-    "**": operator.pow,
+    "**": _power,
     "*": operator.mul,
     "": operator.mul,  # operator for implicit ops
     "/": operator.truediv,
--- a/pint/testsuite/test_unit.py
+++ b/pint/testsuite/test_unit.py
@@ -345,6 +345,12 @@ class TestRegistry(QuantityTestCase):
             kelvin=1, meter=1
         )
 
+    @helpers.requires_numpy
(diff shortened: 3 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           - python-version: 3.8
             numpy: "numpy"
             uncertainties: "uncertainties"
-            extras: "sparse xarray netCDF4 dask[complete] graphviz babel==2.8"
+            extras: "sparse xarray pooch netCDF4 dask[complete] graphviz babel==2.8"
```
dopplershift, line 20: Why are you including pooch? I don't see it used in the PR (or used by the repo at all).

PR #1321, opened 2021-05-28
Add root exception for all pint errors

c-w:
Many libraries follow the pattern where there's a single root exception that can be caught to identify any exception the library throws which makes it very easy to isolate issues related to the library. This pull request adds such a root exception.

Usage example:

```py
import pint
ureg = pint.UnitRegistry()

try:
  # do something involving ureg
except pint.PintError:
  # here we know for sure that the error came from pint
```

  - This is a developer ergonomics improvement that came to mind while working with the library so I didn't open an issue for it.
  - I didn't see any existing references to exception handling in the docs so I didn't add anything related to the new root exception. 


hgrecco:
I think this a good idea. What do you think @jules-ch ?

jules-ch:
That's a good idea ! 👍

PR #1324, opened 2021-05-31
add a CI to automatically update the `pre-commit` hook versions

keewis:
As mentioned in #1302 there's a way to automatically get PRs to autoupdate the `pre-commit` hooks (once a week or on explicit request).

### DIFF
--- a/.github/workflows/lint-autoupdate.yml
+++ b/.github/workflows/lint-autoupdate.yml
@@ -1,4 +1,4 @@
-name: "pre-commit autoupdate"
+name: pre-commit
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


jules-ch:
Looks nice, I think we can integrate this, if your experience with this on xarray has been good.

keewis:
it has been. Something to keep in mind, however, is that the `pre-commit` action apparently has been deprecated in favor of `pre-commit.ci`, which seems to support updates as well (at the expense of some flexibility).

hgrecco:
Let's merge this and then see how the `pre-commit.ci` evolves.

PR #1328, opened 2021-06-03
fix conversion involving negative exponents with `force_ndarray=True`

### LINKED ISSUE #1327: more trouble with `force_ndarray_like=True`
Apparently, `force_ndarray_like=True` results in `q._units` containing `numpy` scalar values, which causes unit conversions to fail:
```python
In [1]: import pint
   ...: 
   ...: ureg = pint.UnitRegistry(force_ndarray_like=True)
   ...: q = ureg.Quantity(1, "1 / hours") ** 2
   ...: display({k: type(v) for k, v in q._units.items()})
   ...: q.to("1 / s ** 2")
{'hour': numpy.int64}
ValueError                                Traceback (most recent call last)
<ipython-input-1-fd499ff53bb2> in <module>
      3 ureg = pint.UnitRegistry(force_ndarray_like=True)
      4 q = 1 / ureg.Quantity(1, "hours") ** 2
----> 5 q.to("1 / s ** 2")

.../pint/quantity.py in to(self, other, *contexts, **ctx_kwargs)
    655         other = to_units_container(other, self._REGISTRY)
    656 
--> 657         magnitude = self._convert_magnitude_not_inplace(other, *contexts, **ctx_kwargs)
    658 
    659         return self.__class__(magnitude, other)

.../pint/quantity.py in _convert_magnitude_not_inplace(self, other, *contexts, **ctx_kwargs)
    604                 return self._REGISTRY.convert(self._magnitude, self._units, other)
    605 
--> 606         return self._REGISTRY.convert(self._magnitude, self._units, other)
    607 
    608     def _convert_magnitude(self, other, *contexts, **ctx_kwargs):

.../pint/registry.py in convert(self, value, src, dst, inplace)
    945             return value
    946 
--> 947         return self._convert(value, src, dst, inplace)
    948 
    949     def _convert(self, value, src, dst, inplace=False, check_dimensionality=True):

.../pint/registry.py in _convert(self, value, src, dst, inplace)
   1828                 value, src = src._magnitude, src._units
   1829 
-> 1830         return super()._convert(value, src, dst, inplace)
   1831 
   1832     def _get_compatible_units(self, input_units, group_or_system):

.../pint/registry.py in _convert(self, value, src, dst, inplace)
   1431 
   1432         if not (src_offset_unit or dst_offset_unit):
-> 1433             return super()._convert(value, src, dst, inplace)
   1434 
   1435         src_dim = self._get_dimensionality(src)

.../pint/registry.py in _convert(self, value, src, dst, inplace, check_dimensionality)
    982         # Here src and dst have only multiplicative units left. Thus we can
    983         # convert with a factor.
--> 984         factor, _ = self._get_root_units(src / dst)
    985 
    986         # factor is type float and if our magnitude is type Decimal then

.../pint/registry.py in _get_root_units(self, input_units, check_nonmult)
    822 
    823         accumulators = [1, defaultdict(int)]
--> 824         self._get_root_units_recurse(input_units, 1, accumulators)
    825 
    826         factor = accumulators[0]

.../pint/registry.py in _get_root_units_recurse(self, ref, exp, accumulators)
    871                 accumulators[1][key] += exp2
    872             else:
--> 873                 accumulators[0] *= reg._converter.scale ** exp2
    874                 if reg.reference is not None:
    875                     self._get_root_units_recurse(reg.reference, exp2, accumulators)

ValueError: Integers to negative integer powers are not allowed.

In [2]: q._units = pint.util.to_units_container({k: getattr(v, "item", lambda: v)() for k, v in q._units.items()})
   ...: print(q.to("1 / s ** 2"))
7.71604938271605e-08 / second ** 2
```
I'm pretty sure this is unrelated to #1301 (although they share a common cause: `numpy.int64(1) ** -2` raises) because this also fails if I go back to the release commit of `0.17`.

keewis:
again, I'm not sure if I'm breaking something, but the test suite doesn't fail locally and it fixes my issue.

- [x] Closes #1327

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -1486,7 +1486,7 @@ class Quantity(PrettyIPython, SharedRegistryObject):
                     raise DimensionalityError(other._units, "dimensionless")
                 else:
                     exponent = _to_magnitude(
-                        other, self.force_ndarray, self.force_ndarray_like
+                        other, force_ndarray=False, force_ndarray_like=False
                     )
                     units = new_self._units ** exponent
 
--- a/pint/testsuite/test_quantity.py
+++ b/pint/testsuite/test_quantity.py
@@ -1563,6 +1563,18 @@ class TestOffsetUnitMath(QuantityTestCase):
                 helpers.assert_quantity_almost_equal(op.pow(in1, in2), expected)
 
     @helpers.requires_numpy
+    def test_exponentiation_force_ndarray(self):
+        self.ureg.force_ndarray_like = True
+        q = self.Q_(1, "1 / hours")
+
+        q1 = q ** 2
+        assert all(isinstance(v, int) for v in q1._units.values())
+
+        q2 = q.copy()
+        q2 **= 2
+        assert all(isinstance(v, int) for v in q2._units.values())
+
+    @helpers.requires_numpy
     @pytest.mark.parametrize(("input_tuple", "expected"), exponentiation)
     def test_inplace_exponentiation(self, input_tuple, expected):
         self.ureg.default_as_delta = False
--- a/pint/testsuite/test_quantity.py
+++ b/pint/testsuite/test_quantity.py
@@ -1564,8 +1564,8 @@ class TestOffsetUnitMath(QuantityTestCase):
 
     @helpers.requires_numpy
     def test_exponentiation_force_ndarray(self):
-        self.ureg.force_ndarray_like = True
-        q = self.Q_(1, "1 / hours")
+        ureg = UnitRegistry(force_ndarray_like=True)
+        q = ureg.Quantity(1, "1 / hours")
 
         q1 = q ** 2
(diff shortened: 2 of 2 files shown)


keewis:
the test failures were created by a faulty test isolation (apparently, `self.ureg` is shared across all tests and if it is mutated all later tests get the modified test version). Which makes me think that maybe there's more to be done to get `ipow` to work.

jules-ch:
@hgrecco has made the transition to pytest but we're not using fixture everywhere & end up with QuantityTestCase & class based UnitRegistry which triggered your problem.
I'll merge your changes

hgrecco:
Indeed, we migrated to pytest but it will be good to use fixtures everywhere. Moreover, during the migration I started to reuse a registry (instead of creating a new one) to speed up tests. This worked very nicely but it would be good to do it more consistently.

PR #1336, opened 2021-06-21
Add Quantity conversion from timedelta

ikseek:

Balance to_timedelta() interface of Quantity with from_timedelta.
Maybe such a conversion should be allowed directly in Quantity.__new__ to happen automatically?


hgrecco:
I wonder if a typecheck or similar is needed. Not sure

PR #1341, opened 2021-06-28
Fix string formatting for numpy array scalars

dopplershift:
Unlike normal arrays, the scalars do not respond to `set_printoptions`, so instead normal formatting needs to be used. (See numpy/numpy#11048 for ongoing confusion about this.) Also, for HTML formatting, they are flagged as `iterable` even though they should be formatted like single values.


jules-ch:
Looks good !

PR #1347, opened 2021-07-14
NEW: Add ability to parse alternate micro-prefix

### LINKED ISSUE #574: Add U+00B5 into the default registry for micro prefix

### LINKED ISSUE #1030: micro prefix incompatibility with babel
The mu prefix returned by babel cannot be interpreted by pint. 

```python
import pint
import unicodedata

ureg = pint.UnitRegistry()

# symbol used internally by pint, this works
ureg(unicodedata.lookup('MICRO SIGN'))

# symbol returned by babel, fails
ureg(unicodedata.lookup('GREEK SMALL LETTER MU'))
```

carterbox:
Due to legacy reasons, there are two unicode characters for Greek lowercase mu: U+03BC and U+00B5. Both are valid prefixes for micro. This commit ensures that both Unicode characters are included in the unit definitions file.

This means that both characters will be parsed by Pint without error. It should not affect how Pint prints anything.

More information at Wikipedia: https://en.wikipedia.org/wiki/Micro-#Symbol_encoding_in_character_sets

- [x] Closes #1030, Closes #574
~~Documented in docs/ as appropriate~~ No changes made to docs; I see this PR as a bug fix.


carterbox:
Retroactively, I realize that this PR is a duplicate of previous efforts. However, for reasons unclear to me, the unicode character for mu never seems to make into the definition file.

carterbox:
Related to #680, #603, #597, #222

hgrecco:
I think you need to distinguish between the ability to parse and to print. When parsing adding an alias is easy. Just append at the new alias at the end of the definition or add a replacement rule.

Printing is trickier from the user experience perspective. Right now we support printing the full name of a unit (or prefix) or the symbol (i.e. abbreviated version). We do not have a way to let the user choose if they want to use a unicode version or not (or which version to use).  Most discussions related to unicode in Pint has been due to this. 

In any case, I am ok with supporting multiple parsing options but for printing we should choose one and stick to it

carterbox:
I'm only advocating for adding additional parsing ability. Pint should not raise an error if the user provides the "wrong" μ. I agree that there should be one printing option, but I have no opinions about which one (However Wikipedia claims that SI and Unicode both prefer the Greek mu instead of micro character).

I'll try to get all the boxes checked on this PR in the next 20 hours.

carterbox:
Ready for review.

jules-ch:
Can you add a comment in this line so we know why we have the same glyph twice.
Just as a reminder, you just have to add a comment with `#` on the same line.

I'll look into which one should be prefered for printing, I think we can keep the one defined in pint as not to break compatibility.

Otherwise we will introduce a breaking change if necessary

pint/default_en.txt
```diff
@@ -68,7 +68,7 @@ atto- =  1e-18 = a-
 femto- = 1e-15 = f-
 pico- =  1e-12 = p-
 nano- =  1e-9  = n-
-micro- = 1e-6  = µ- = u-
+micro- = 1e-6  = µ- = μ- = u-
```
carterbox, line 71:
```suggestion
# The micro (U+00B5) and Greek mu (U+03BC) are both valid prefixes,
# and they often use the same glyph.
micro- = 1e-6  = µ- = μ- = u-
```

cgevans:
Would there be anything that would be helpful in getting this included, perhaps in the next release?  It seems like a very small change, as it is only adding the symbol to parsing, and the lack of the symbol makes using pint with a Greek keyboard surprisingly frustrating.

carterbox:
Ready for review

rjgildea:
Is there anything preventing this from being merged? We would very much like support for both micro-prefixes added to pint as we are using in the parsing of [NeXus](https://manual.nexusformat.org/index.html) files which [defer](https://manual.nexusformat.org/datarules.html#nexus-data-units) to the [UDUNITS](https://www.unidata.ucar.edu/software/udunits/udunits-2.2.28/udunits2-prefixes.xml) package for valid unit formats, which includes both micro-prefixes.

carterbox:
Thanks, @hgrecco!

PR #1349, opened 2021-07-14
Allow string parsing of offset units

### LINKED ISSUE #386: OffsetUnitCalculusError on String Parsing of degC, degF
I encountered the problem by:

`>>> from pint import UnitRegistry`
`>>> ureg = UnitRegistry()`

and these following commands 
`>>> 1*ureg['celsius']`
`>>> 1*ureg['degC']`
`>>> 1*ureg.parse_expression('degC')`
`>>> 1*ureg('degC')`
`>>> ureg.parse_expression('1*degC')`
all will throw error:

```
  File "/Library/Python/2.7/site-packages/pint/quantity.py", line 727, in __mul__
    return self._mul_div(other, operator.mul)
  File "/Library/Python/2.7/site-packages/pint/quantity.py", line 682, in _mul_div
    getattr(other, 'units', ''))
pint.errors.OffsetUnitCalculusError: Ambiguous operation with offset unit (degC).
```

using ureg.degC or ureg.degF will not cause the problem

```
>>> 1*ureg.degC
<Quantity(1, 'degC')>
>>> 1*ureg.degF
<Quantity(1, 'degF')>
```

using ureg.Quantity will not cause the problem

```
>>> q_ = ureg.Quantity
>>> q_(1,'celsius')
<Quantity(1, 'degC')>
```

My python version 

```
python --version
Python 2.7.10
```

My Pint version

```
pip show pint

---
Metadata-Version: 1.1
Name: Pint
Version: 0.7.2
```

My System version

```
sw_vers
ProductName:    Mac OS X
ProductVersion: 10.11.4
BuildVersion:   15E65
```

burnpanck:
- [x] Closes #386

The fix implements special handling for implicit multiplications in the string parser when the left operand is not a quantity:
Then, it creates a quantity directly, instead of multiplying the two operands.
For this to work, the `string_preprocessor` had to be modified not to replace implicit multiplications with explicit multiplications.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -35,6 +35,9 @@ Pint Changelog
 - Improved the application registry.
   (Issue #1366, thanks keewis)
 - Improved testing isolation using pytest fixtures.
+- Fix string formatting of numpy array scalars
+- Allow string parsing of offset units
+  (Issue #386)
 
 ### Breaking Changes
 
--- a/CHANGES
+++ b/CHANGES
@@ -47,6 +47,7 @@ Pint Changelog
 0.17 (2021-03-22)
 -----------------
 
+- Do not build "universal" wheel as Python 2 is not supported.
 - Add the Wh unit for battery capacity measurements
   (PR #1260, thanks Maciej Grela)
 - Fix issue with reducable dimensionless units when using power (Quantity**ndarray)
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -85,7 +85,7 @@ from .errors import (
     RedefinitionError,
     UndefinedUnitError,
 )
-from .pint_eval import build_eval_tree
+from .pint_eval import _BINARY_OPERATOR_MAP, build_eval_tree
 from .systems import Group, System
 from .util import (
     ParserHelper,
@@ -1232,6 +1232,11 @@ class BaseRegistry(metaclass=RegistryMeta):
         else:
             raise Exception("unknown token type")
 
+    def _eval_implicit_mul(self, left, right):
+        if isinstance(left, self.Quantity):
+            return left * right
+        return self.Quantity(left, right)
+
     def parse_pattern(
         self,
         input_string: str,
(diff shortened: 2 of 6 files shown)


burnpanck:
This fix will probably not work for compound offset units such as `degC m`. However, I believe that these things are neither fully supported by the rest of pint, nor should they be; see #1350.

burnpanck:
Not completely sure what I see in the CI results here (haven't worked with GitHub CI recently).
Locally on my machine, a number of tests would fail already in the master branch where I branched this PR off.

dalito:
Would e.g. `ureg.Quantity("42 degC/m")` raise an error with your PR? In other words, is the idea to allow parsing of strings with an offset unit only if there is no other unit present in the string and if the unit´s exponent is 1? I think with such limitations parsing of strings with offset units would be Ok.

burnpanck:
I believe `42 degC/m` would indeed fail. All this PR does is treating `<number> <unit>` combinations not as a multiplication, but as "forming of a quantity" (notice the absence of a `*` - multiplications remain untouched). I believe that is what a human does when reading such expressions. I haven't touched the grouping priority or associativity of operations, so all product units will remain to be parsed from left to right, such that this "forming of quantity" only happens between the number and the first unit, followed by multiplications by further units.

Generally: As I tried to stipulate in #1350, products involving offset are generally questionable, as they lack a physical meaning - IHMO they should be completely disallowed. However, this PR simply follows the rules set by pint; it makes the expression parser accept "forming of quantity" involving offset units, and then continues evaluating the expression given in the string according to pint's rules. I think that is the least surprising behaviour.

burnpanck:
It seems that #386 is still unsolved. At our company, we have been using a fork of pint for the past two years to get faster access to some fixes that we care for particularly. One of these things is the ability to serialise quantities to json or yaml and back elegantly as a single, human-readable string rather than an ugly `[magnitude, units]` construct. We are currently updating all our dependencies, and I was hoping to see this fixed now. 

@hgrecco, @dalito: What was speaking against merging this two years ago? Obviously, by now, this PR has become stale, but I would be willing to revive it if there is any chance that it would be merged.

PR #1356, opened 2021-07-25
Fix dask tests

jules-ch:
Fix Dask tests

xref https://github.com/pydata/xarray/pull/5610


jules-ch:
timeout = None is not possible anymore on new versions of dask

PR #1361, opened 2021-08-02
Fix parsing units with same exponents but different sign

### LINKED ISSUE #1360: Cannot parse units with pretty form & same exponent (both negative & positive)
Just encountered this, we can not parse strings like `'kg²·s⁻²'`.

String processing is replacing all `²` with `**2` resulting in `kg**2*s-**2` which fails.

https://github.com/hgrecco/pint/blob/e591f8c42e7b451c02d30cc7a4f9ac03fa66f840/pint/util.py#L777-L792

jules-ch:
- [x] Closes #1360


hgrecco:
LGTM!

PR #1366, opened 2021-08-05
improve the application registry

### LINKED ISSUE #1365: improve the application registry for use in packages
xref xarray-contrib/pint-xarray#129

The currently recommended way to use the application registry is:
```python
# application
ureg = pint.UnitRegistry()
# configuration
pint.set_application_registry(ureg)

import package_using_pint

# package_using_pint
registry = pint.get_application_registry()
```

There are a few issues with this:
1. there's code before an import so tools like `isort` will complain
2. since all `set_application_registry` does is assigning its argument to a module-global, all other references to the previous object are not updated.
3. as a consequence of 2, packages providing new registries either need to be imported before packages using them, or they need to provide a setup function that needs to be called before the import of those packages

I don't really like setup functions, though, especially if the package can clearly separate the unit registry from other functionality. E.g. `import cf_xarray` does not even try to import `pint`, but `import cf_xarray.units` will create and set a custom unit registry.

To solve this, I'd like to propose exposing a `UnitRegistry` wrapper object as `pint.application_registry`:
```python
class ApplicationRegistry:
    def __init__(self, registry):
        self._registry = registry

    def set(self, new):
        self._registry = new

    def __getattr__(self, name):
        return getattr(self._registry, name)

    def __dir__(self):
        return dir(self._registry)

    # other __*__ methods

application_registry = ApplicationRegistry(_DEFAULT_REGISTRY)
```
with this, packages using the application registry can just `import pint` or `from pint import application_registry`, and other packages providing a registry can call `.set` to have all other packages get the new registry, without having to care when/in which order they are imported.

Also note that this is completely independent of the `get/set_application_registry` functions.

Not sure if I'm missing something obvious, but if this sounds like a good idea I'll send in a PR.

keewis:
This adds the application registry wrapper object proposed in #1365 to make the application registry easier to use between packages. `get_application_registry` and `set_application_registry` were changed to use the new wrapper class.

No docstrings, docs or extra tests, yet, and the testsuite still fails.

cc @jthielen

- [x] Closes #1365, closes xarray-contrib/pint-xarray#129

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,7 @@ Pint Changelog
 - Fix string formatting of numpy array scalars
 - Fix default format for Measurement class (Issue #1300)
 - Fix parsing of pretty units with same exponents but different sign. (Issue #1360)
+- Convert the application registry to a wrapper object (Issue #1365)
 
 ### Breaking Changes
 
--- a/pint/__init__.py
+++ b/pint/__init__.py
@@ -48,7 +48,6 @@ _DEFAULT_REGISTRY = LazyRegistry()
 
 #: Registry used for unpickling operations.
 application_registry = ApplicationRegistry(_DEFAULT_REGISTRY)
-_APP_REGISTRY = application_registry
 
 
 def _unpickle(cls, *args):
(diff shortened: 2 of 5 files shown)


keewis:
This still needs to be documented, but the code should be ready for review.

kthyng:
Looking forward to this to simplify my files! Thanks for your work!

hgrecco:
I agree with the implementation. I think we need to document this very clearly as has raised multiple questions in the past. Great work!

keewis:
@hgrecco, @jules-ch, would it be fine to merge this without the docs and I'll add documentation in a new PR?

hgrecco:
I agree

jules-ch:
Sorry I haven't had time to review PRs pr issues lately. I'm okay to merge this, @keewis  can you create a PR for documentation for this feature. I think we will do this with the rework on documentation.

jules-ch:
Merging this :+1:

Issue #972, opened 2020-01-07
Improve the docs


hgrecco:
This is just a placeholder issue to obtain feedback about the current state of the docs:
1. Are the docs clear enough?
2. Are the examples useful and real?
3. Is the information there for a newbie?
4. Is the information there for an advanced user?
5. Is the organization appropiate?
6. What can we do to improve it.

Meistermagier:
There seems to be an issue regarding the Dataframe Integration 
```
df_ = df.pint.quantify(level=-1)
df_
```
This Code Snippet given by the Docs does not work and returns the error; 'DataFrame' object has no attribute 'pint'

Meistermagier:
i just checked more to the Dataframe Integration this part also Produces Errors:
```
import pandas as pd
import pint
df = pd.DataFrame({
    "torque": pd.Series([1, 2, 2, 3], dtype="pint[lbf ft]"),
    "angular_velocity": pd.Series([1, 2, 2, 3], dtype="pint[rpm]"),
})
df
```
Produces the error: data type 'pint[lbf ft]' not understood

Im now assuming there is just an import or a setup missing in the Documentation and that fails the Code snippets. Which makes the Pandas Integration part very unintuitive or rather not at all understandable because it is not reproducable.

jthielen:
Here are some of the thoughts I had (from a high-level perspective, not getting into finer details of the individual pages):

- **Content coverage**
  - I think the breadth and depth of coverage of material in the docs is pretty good!
  - Perhaps a clearer organization of the user guide into sections (or even just a getting started/user guide/help & reference split like [xarray](http://xarray.pydata.org/en/stable/index.html)) would be worthwhile
- **Navigation**
  - I wish there was a better sidebar menu so that you didn't have to jump back to the homepage, click through multiple sections, or use search to navigate and browse the docs
- **Small-screen friendliness**
  - Right now, the pages are somewhat hard to use on a mobile device or restricted screen width (lack of good top menu, content overflowing device width, etc.)
  - The "Fork me on GitHub" corner banner gets in the way
  - The search experience is especially poor
- **API Reference**
  - The current API reference, the "Developer Reference" page https://pint.readthedocs.io/en/latest/developers_reference.html is somewhat difficult to use (with its current state, I find it easy to just refer to the source code itself than use that portion of the docs)
  - The formatting does not establish a clear separation between components, and the lack of a summary view makes it hard to browse
  - Not sure what is considered public vs. private API?
[...]

PR #1371, opened 2021-08-14
custom unit formats

### LINKED ISSUE #1357: formats, documentation and custom formats
I think there are several issues with unit formatting right now:

Related to xarray-contrib/pint-xarray#121: whenever I need to format units, rather than the documentation I tend to look through the source code of `pint.formatting`, especially `_FORMATS` is useful. This, I think, is because the documentation of the format values is hidden pretty well (the only mention I can find is in the tutorial which is incomplete, i.e. some of the formats and the quantity modifer `"#"` are never mentioned).

I also think the default format (`""` in `_FORMATS`) should get a actual name: if I set `UnitRegistry.default_format = "P"`, the "default format" as defined in `_FORMATS` is not accessible anymore until I set the registry's default back to `""`.

Finally, it would be great to have some kind of mechanism to register custom formats (we already can modify `pint.formatting._FORMATS`, but that's a implementation detail, right?). The syntax could be
```python
@pint.register_unit_format("xyz")
def format_units_xyz(units, modifier):
    ...
```
where the register function should make sure it does not overwrite already existing formats.

Not sure if it would be better to open separate issues for each of these, but I though it might make sense to consider them together.

keewis:
This adds a `register_unit_format` function and implements the currently available formats using that. I'd also like to switch the `siunitx` code (`"Lx"`) to use the new format, but because that is implemented in a function that expects a `Unit` object instead of a `UnitContainer` I still need to figure out a way to make it compatible.

I also renamed the `""` format to `"D"`.

- [x] Closes #1357 (possibly more)

### DIFF
--- a/pint/formatting.py
+++ b/pint/formatting.py
@@ -83,6 +83,81 @@ def register_unit_format(name):
     return wrapper
 
 
+@register_unit_format("P")
+def format_pretty(unit):
+    return formatter(
+        unit.items(),
+        as_ratio=True,
+        single_denominator=False,
+        product_fmt="·",
+        division_fmt="/",
+        power_fmt="{}{}",
+        parenthesis_fmt="({})",
+        exp_call=_pretty_fmt_exponent,
+    )
+
+
+@register_unit_format("L")
+def format_latex(unit):
+    return formatter(
+        unit.items(),
+        as_ratio=True,
+        single_denominator=True,
+        product_fmt=r" \cdot ",
+        division_fmt=r"\frac[{}][{}]",
+        power_fmt="{}^[{}]",
+        parentheses_fmt=r"\left({}\right)",
+    )
+
+
+@register_unit_format("Lx")
+def format_latex_siunitx(unit):
+    return formatter(
+        unit.items(),
+        siopts="",
+        pm_fmt=" +- ",
+    )
+
+
+@register_unit_format("H")
+def format_html(unit):
+    return formatter(
+        unit.items(),
+        as_ratio=True,
+        single_denominator=True,
+        product_fmt=r" ",
+        division_fmt=r"{}/{}",
+        power_fmt=r"{}<sup>{}</sup>",
+        parentheses_fmt=r"({})",
+    )
+
+
+@register_unit_format("D")
+def format_default(unit):
+    return formatter(
+        unit.items(),
+        as_ratio=True,
+        single_denominator=False,
+        product_fmt=" * ",
+        division_fmt=" / ",
+        power_fmt="{} ** {}",
+        parentheses_fmt=r"({})",
+    )
+
+
+@register_unit_format("C")
+def format_compact(unit):
+    return formatter(
+        unit.items(),
+        as_ratio=True,
+        single_denominator=False,
+        product_fmt="*",  # TODO: Should this just be ''?
+        division_fmt="/",
+        power_fmt="{}**{}",
+        parentheses_fmt=r"({})",
+    )
+
+
 def formatter(
     items,
     as_ratio=True,
--- a/pint/formatting.py
+++ b/pint/formatting.py
@@ -70,52 +70,8 @@ def _pretty_fmt_exponent(num):
     return ret
 
 
-#: _FORMATS maps format specifications to the corresponding argument set to
-#: formatter().
-_FORMATS: Dict[str, dict] = {
-    "P": {  # Pretty format.
-        "as_ratio": True,
-        "single_denominator": False,
-        "product_fmt": "·",
-        "division_fmt": "/",
-        "power_fmt": "{}{}",
-        "parentheses_fmt": "({})",
-        "exp_call": _pretty_fmt_exponent,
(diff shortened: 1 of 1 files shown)


keewis:
this currently fails because `Measurements.__format__` assumes `dict`s in `_FORMATS`. I guess we should probably move that code to a branch in `formatter`.

keewis:
actually, maybe this is a case for the `Quantity` format (which would require `register_quantity_format`): `Measurement` is a subclass of `Quantity`, so the code should not relate to the unit format.

Since it's pretty difficult to do that without breaking anything, I renamed the `dict` of `callable`s to `_FORMATTERS` and added back the `dict` of `dict` structure for `_FORMATS`.

keewis:
this still needs proper documentation (e.g. there should be a complete list of formats available, with examples), but I guess that depends on #972

keewis:
I just noticed that `_parse_spec` really is for quantity format specs, so `format_unit` doesn't really need the additional complexity. 

The tests fail because `siunitx_format_unit`, which I decided to call from the registered `format_latex_siunitx`, needs a registry. Not sure how to fix that, maybe by adding a `_registry` field to `UnitsContainer`?

hgrecco:
Regarding `siunitx_format_unit` The way we have solve this issues in the past is having a registry parameter. It might make sense here.

pint/formatting.py
```diff
@@ -117,6 +117,135 @@ def _pretty_fmt_exponent(num):
...
+    ...
+    >>> ureg = pint.UnitRegistry()
+    >>> u = ureg.m / ureg.s ** 2
+    >>> f"{u:custom}"
+    '<formatted unit>'
+    """
+
+    def wrapper(func):
+        if name in _FORMATTERS:
+            raise ValueError(f"format {name:!r} already exists")  # or warn instead
```
keewis, line 155: should this be a warning or an error?

pint/formatting.py
```diff
@@ -117,6 +117,135 @@ def _pretty_fmt_exponent(num):
...
+    name : str
+        The name of the new format (to be used in the format mini-language). A error is
+        raised if the new format would overwrite a existing format.
+
+    Examples
+    --------
+    >>> @pint.register_unit_format("custom")
+    ... def format_custom(unit, registry, **options):
+    ...     result = "<formatted unit>"  # do the formatting
+    ...     return result
```
keewis, line 145: this is executed twice (once for `pint.formatting.register_unit_format` and once for `pint.register_unit_format`), and because there's no isolation in between the runs, the second run will try to register a pre-existing name. Not sure how to avoid that... does `sphinx` have a test isolation feature, or should we switch to `python -m pytest --doctest-modules` and the `ipython` / `jupyter` `sphinx` directives?

keewis: as far as I can tell, there's not much we can do here, except switching the doctest runner to `pytest` (and the `doctest` blocks in the `rst` files to `ipython` / `jupyter-execute`), to replace `autodoc` with `autosummary`, or to make sure `autodoc` does not generate pages for both `pint.register_unit_format` and `pint.formatting.register_unit_format`.

Since all of these are out-of-scope for this PR, I'll convert the example to a code block and we can discuss further changes in a dedicated issue / PR.

pint/formatting.py
```diff
@@ -117,6 +117,135 @@ def _pretty_fmt_exponent(num):
...
+
+@register_unit_format("Lx")
+def format_latex_siunitx(unit, registry, **options):
+    if registry is None:
+        raise ValueError(
+            "Can't format as siunitx without a registry."
+            " This is usually triggered when formatting a instance"
+            ' of the internal `UnitsContainer` with a spec of `"Lx"`'
+            " and might indicate a bug in `pint`."
+        )
```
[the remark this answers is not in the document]
hgrecco: I think is fine. But I think in any case this should not happen if it is used internally.

pint/formatting.py
hgrecco: I think it should produce an error ... or provide a mechanism like we do for the registry with `on_redefinition`.

pint/formatting.py
hgrecco: Yes, there is no way out.

pint/formatting.py
jules-ch: We can start with raising an error for now.
If possible we should protect default formats.

jules-ch:
Good work @keewis looks promising. 👍

keewis:
I didn't bring this up before, but I basically disallowed passing quantity modifiers (just `#` for now) to `Unit.__format__`. So
```python
f"{u:#C}"
```
is a error, but
```python
ureg.default_format = "#C"
f"{u}"
```
is fine.

This might be a breaking change (it caused a few tests to fail in `pint-xarray`, but I think I'll just remove those) so it would be good to discuss this. For reference, ignoring those is as simple as calling `extract_custom_formats` on `spec` in `Unit.__format__`, just as we do with `default_format`.

PR #1372, opened 2021-08-15
don't fail all CI jobs of a matrix if one job fails

keewis:
Currently, all jobs of a matrix will be cancelled / skipped if a single one fails. This is problematic if e.g. there's a failure for both the `uncertainties`- and the `numpy`-only CI, but the `uncertainties`-only CI finishes first: in that case, all other CI is cancelled, including the `numpy`-only CI. This in turn will severely reduce the usefulness of the CI.

The `fail-fast: false` setting disables that, with the disadvantage that now all jobs will complete (or time out).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,8 +5,8 @@ on: [push, pull_request]
 jobs:
   test-linux:
     strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


hgrecco:
Lookgs good to me

PR #1373, opened 2021-08-20
Implement `Quantity.to_no_prefix` for stripping SI-prefix

nikosavola:

This PR adds a method for Quantity to return a copy with no SI-prefixes, e.g., nanometers to meters w/o specifying that meters are the desired unit.

```python
import pint
Q_ = pint.Quantity
(Q_('8 um')).to_no_prefix()
```
returns 
```console
<Quantity(8e-6, 'meter')>
```

### DIFF
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -883,6 +883,28 @@ class Quantity(PrettyIPython, SharedRegistryObject, Generic[_MagnitudeType]):
 
         return self.to(new_unit_container)
 
+    def to_no_prefix(self) -> Quantity[_MagnitudeType]:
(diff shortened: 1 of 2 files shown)


hgrecco:
I think the implementation still needs work. Converting to a string and then replacing by `""` is fragile. It would be better to iterate over units, parse each individually and then remove the prefix. Take a look at the Unit api.

jules-ch:
I agree with @hgrecco, there is already a function that provides parsing of prefix and units. I'll provide a better review soon.
Difficult for me to provide a review at the moment.

PR #1375, opened 2021-08-21
string formatting guide

### LINKED ISSUE #1357: formats, documentation and custom formats

keewis:
As expressed in #1357, the tutorial does not cover everything on the formats (nor should it!) and that section is pretty difficult to find when trying to look up the available formats (also, I didn't know where to put the `register_*_format` functions)

Edit: maybe we should add a `show_formats` function together with the `register_*_format` functions?

The `Quantity` formatting is still left because I don't yet know enough about that.

@hgrecco, RTD has a pretty neat PR preview feature which can be enabled through the admin settings. Could you enable this?

- [x] partially closes #1357


hgrecco:
@keewis I enabled `Build pull requests for this project ` I guess that is all, right?

keewis:
indeed. You should have gotten a message from github requesting the access. If so, we should be able to close/reopen this to see the effect. Let's try:

hgrecco:
I don't see it in the status. I am checking if the webhook is ok

keewis:
for reference, here's the docs on that feature: https://docs.readthedocs.io/en/stable/pull-requests.html

hgrecco:
Somehow there has not been

keewis:
it's there now! Thanks!

hgrecco:
Just for the record, the read the docs documentation [instructs](https://docs.readthedocs.io/en/stable/webhooks.html#github) to enable the webhook for `Branch or tag creation, Branch or tag deletion and Pushes events` It seems that you also need to enable for `Pull requests` when this feature is used.

hgrecco:
@keewis We also need to check that the visibility of the generated docs are fine. ping me to let me know if something is not ok.

keewis:
> We also need to check that the visibility of the generated docs are fine.

Not sure if I understand this correctly. Could you elaborate?

hgrecco:
Ohh, when I was going to link to a reference I found (`Privacy levels are only supported on Read the Docs for Business.`) So do not worry, I was afraid that the outcome of the preview was only visible to me!

hgrecco:
This preview feature is great! How about dividing the index? It seems too crowded. This is a proposal
```
Getting started
    Installation
    Tutorial

User Guide
    Defining Quantities
    String formatting
    Temperature conversion
    Logarithmic Units

Rolling your own conversions and making a better use of those included (find a short title)
    Defining units
    Different Unit Systems (and default units)
    Contexts

Interaction with other packages
    NumPy Support
    Serialization
    Plotting with Matplotlib
    Pandas support

Other cool stuff (find a title)
    Buckingham Pi Theorem
    Optimizing Performance
    Wrapping and checking functions
    Using Pint for currency conversions
    Using Measurements (This will go to Interaction with other packages, I put this here because I want to extract measurements to another package)

More information
    Command-line script
    Developer reference
    Contributing to Pint
    Frequently asked questions


```

keewis:
indeed. I think @jules-ch wanted to do a reorganization of the docs and introduce a new theme (see #972), which is where I would do this, too.

keewis (commented):
this should be ready for a review. I'd especially be interested in the opinions of people learning about string conversion in `pint` for the first time, since that's the best way for me to find things I wrongly assumed to be common knowledge. Not sure if you still fit in that category, but still: cc @TomNicholas, @jbusecke

docs/formatting.rst
```diff
@@ -0,0 +1,112 @@
...
+String formatting
+=================
+The conversion of :py:class:`Unit` and :py:class:`Quantity` objects to strings (e.g.
+through the :py:class:`str` builtin or f-strings) can be customized using :ref:`format
+specifications <formatspec>`. The basic format is:
+
+
+.. code::
+
+   [magnitude format][modifier][unit format]
```
keewis, line 20: it appears that this is invisible after building

keewis: well, that's weird. I can not reproduce locally, so this might have something to do with RTD. In any case, I worked around this using the `::`-at-the end-of-a-text-block syntax (the only disadvantage is that it has syntax highlighting)

keewis:
now that the last issue is resolved this should be ready for a final review

cpascual:
Hi @keewis , I just read the proposed formatter docs: I found them great! 
Just one minor thing: I found that the difference between the `C`  format type specifier for the units and the `#` modifier for quantities is not easy to grasp on a quick reading (on my first glance I actually assumed that there was a typo and that you meant `C` where you wrote `#`). Perhaps a note or some clarification in the text could help?

Maybe something like adding *"(see also the related `#` modifier)"* when first mentioning `C`?

keewis:
they actually do different things, despite their naming: `#` will convert to a unit that makes the quantity have as few decimal digits as possible while `C` will reduce the space taken up by operators (e.g. by not padding them with spaces).

I guess we should try to make the naming distinct, and if that's not possible I can adding a more detailed explanation to both (maybe by replacing the format type's name with a "description" column?)

cpascual:
@keewis you are right! I was still confused even after a not-so-quick read (but the blame is on me, not on the text! )

keewis:
well, texts can be written in a way that is less misleading, even if the reader is less attentive. I will try to see if I can make the difference a bit more visible.

jules-ch:
@keewis I'm gonna merge this.
It's a good addition to the docs.
If you want to amend this, open a PR I'll be happy to merge it.

PR #1376, opened 2021-08-23
fix examples section (numpy docstyle compatible)

marscher:
https://numpydoc.readthedocs.io/en/latest/format.html#examples


marscher:
the failed doc build seems unrelated.

keewis (approved):
indeed. There's an issue with how RTD sets up the environment (system packages are enabled), which keeps a unsupported version of `numpy`. Given that this is the same for `master`, I think we can ignore this here.

In any case, thanks for the PR, @marscher

jules-ch (approved):

PR #1377, opened 2021-08-23
fix the RTD builds

keewis:
as described in https://github.com/hgrecco/pint/pull/1376#pullrequestreview-736375619, the RTD builds are currently failing because the environment includes a old version of `numpy` (via the system packages). To fix that, we can set `system_packages` to `false`.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -10,3 +10,4 @@ python:
      - requirements: requirements_docs.txt
      - method: pip
        path: .
+   system_packages: false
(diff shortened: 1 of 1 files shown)


hgrecco:
LGTM

keewis:
we should merge this as soon as possible because RTD is currently failing on `master`

jules-ch:
Thats weird, It should be as `false` by default https://docs.readthedocs.io/en/stable/config-file/v2.html#python-system-packages

I'll merge like this for the CI yo work, no side effects I guess

PR #1378, opened 2021-08-23
add more `pre-commit` hooks

keewis:
This adds a few basic hooks:
- `trailing-whitespace` (remove trailing whitespace)
- `end-of-file-fixer` (make sure there's a newline before EOF)
- `check-yaml` (make sure the `.yml` / `.yaml` file is valid yaml)

---


hgrecco:
Just to check ... the effect of these hooks cannot be obtained with the existing hooks?

keewis:
It's true that `black` will reformat python files to not contain trailing whitespace and to end with a newline, but the current hooks only work with python files, while the new hooks would be applied to every text file (obviously, the `check-yaml` hook is only applied to `yaml` files)

hgrecco:
@keewis Sounds good to me then.

Issue #1277, opened 2021-03-26
Suggestion: Add unitless dimensions


edmundsj:
Not sure if this would be difficult to implement internally, but one common thing I do with units is convert between various unitless representations (i.e. I want to convert a percent to ppm).

Suggested additional dimensionless units:
- percent
- ppthousand
- ppm
- ppb
- ppt
- ppq

hgrecco:
I would be ok with thi.

keewis:
note that "ppb", "ppt", and "ppq" are ambiguous ("billion" might mean either 10<sup>9</sup> or 10<sup>12</sup>, depending on the [language](https://en.wikipedia.org/wiki/Long_and_short_scales)), and that "ppthousand" should probably be an alias of "permille"

hgrecco:
@keewis True, this is ambiguos. I think we should not add it.

jules-ch:
> The internationally recognized symbol % (percent) may be used with the SI. When it is
used, a space separates the number and the symbol %. The symbol % should be used rather
than the name “percent”. In written text, however, the symbol % generally takes the
meaning of “parts per hundred”. Phrases such as “percentage by mass”, “percentage by
volume”, or “percentage by amount of substance” shall not be used; the extra information
on the quantity should instead be conveyed in the description and symbol for the quantity.
The term “ppm”, meaning 10e−6 relative value, or 1 part in 10e6
, or parts per million, is also
used. This is analogous to the meaning of percent as parts per hundred. The terms “parts per
billion” and “parts per trillion” and their respective abbreviations “ppb” and “ppt”, are also
used, but their meanings are language dependent. For this reason the abbreviations ppb and
ppt should be avoided. 

From the SI Brochure BIPM : https://www.bipm.org/utils/common/pdf/si-brochure/SI-Brochure-9-EN.pdf

cpascual:
In view of the above info, in my opinion percent  and ppm should definitely be supported out of the box (and ppthousand / ppmille maybe too), while ppb, ppt, ppq are better left out.

I also think that the `%` symbol should be supported for percent. I am aware that thanks to #911 it is now possible to add a custom preprocessor to support it, but IMHO, it should be available out of the box

hCraker:
Any updates on adding ppm and other concentration units to MetPy?

dalito:
@hCraker - Note that "ppm" is not a unit for concentration accoding to the citation from the SI Brochure BIPM above.

[...]

PR #1392, opened 2021-09-30
Added units for atmospheric chemistry applications

vinodkatmos:
Added volume mixing ratio units like ppt, ppm and ppb

### DIFF
--- a/pint/default_en.txt
+++ b/pint/default_en.txt
@@ -336,6 +336,12 @@ particle = 1 / N_A = _ = molec = molecule
 [concentration] = [substance] / [volume]
 molar = mole / liter = M
 
+# Volume mixing ratio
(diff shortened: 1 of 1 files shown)


cpascual:
I think this should be evaluated taking into account the discussion in #1277 .
Specifically, the doubts regarding the ambiguous ones such as ppb

cpascual:
Also, this PR proposes adding `ppmV = 1e-6 mole / mole = ppm`. I would not associate ppm (or ppb and ppt, etc) specifically to *Volume* ratios, since it can be used in so many other contexts (mass, particles,...). Instead, I would prefer an approach like that in #1277 where they are simply considered unitless dimensions

andrewgsavage:
this would need some discussion, please open a discussion/issue

PR #1397, opened 2021-10-20
PR: Avoid looping on large numpy arrays

impact27:
Some numpy arrays have a dtype (such as int32) that excludes any objects. In that case, looping the array looking for a quantity is not helpful, and can waste a lot of time if the array is large.


andrewgsavage:
done this in #1987

PR #1399, opened 2021-10-26
Add support for python 3.10

jules-ch:


keewis:
> remove the hardcoded steps

this doesn't work, those are hard-coded by RTD's build system. The official way to override those (according to [their docs](https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#don-t-rely-on-implicit-dependencies)) is to pin the dependencies in our requirements file.

jules-ch:
We'll do that then

PR #1403, opened 2021-11-01
Fix mutating the application registry

keewis:
When adding the `ApplicationRegistry` class I apparently forgot to forward `__setattr__`. This means that trying to mutate the application registry does not change the wrapped registry, but rather adds a new attribute to `ApplicationRegistry`:
```python
from pint import application_registry as ureg

ureg.force_ndarray_like = True
assert ureg.force_ndarray_like is True
assert ureg._registry.force_ndarray_like is True  # AssertionError
```

Since this is included in `v0.18`, we might need a new release soon. What do you think, @hgrecco, @jules-ch

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ Pint Changelog
 0.19 (unreleased)
 -----------------
 
-- Nothing changed yet.
+- Fix setting options of the application registry (Issue #1403).
 
 
 0.18 (2021-10-26)
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -2328,6 +2328,8 @@ class LazyRegistry:
 class ApplicationRegistry:
     """A wrapper class used to distribute changes to the application registry."""
 
+    __slots__ = ["_registry"]
+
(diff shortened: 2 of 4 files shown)


keewis:
I'm not sure if that was a random failure on RTD, but pinning `sphinx>4` seems to have worked

jules-ch:
I've explained the error in RTD build here https://github.com/hgrecco/pint/pull/1399#issuecomment-952393441.
We need to get rid of those hard coded dependencies. We can pin sphinx version like you did in the meantime

Looks good to me.
We can do a 0.18.1 release

keewis:
great. Should I add a note to `CHANGES`?

keewis:
oh, and is it fine to keep the pin in this PR?

jules-ch:
Yes Add this fix in CHANGES & keep the pin I'll pin other docs dependencies in another PR.

keewis:
> add this fix in CHANGES

done

jules-ch:
Thanks @keewis

PR #1409, opened 2021-11-06
Fix use of offset units for higher dimensional units (e.g. gauge pressure #1066)

### LINKED ISSUE #1066: Absolute vs. Relative pressure
Dear all,

I could not find a differentiation between absolute and relative pressures in the list of units proposed by pint. Is this feature already implemented? If not, I think it would make sense to have it, as it is very common to differentiate between these two pressures. One could have units looking like `bar_relative` and `bar_absolute` for instance.

Thank you in advance and best regards,

dalito:
- [x] Closes #1066

The problem was limited to offset units of higher dimension like pressure (M L<sup>-1</sup> T<sup>-2</sup>). The offset units included in pint are all temperature units with just one dimension and were thus not affected.


jules-ch:
@dalito can you rebase & push to resolve conflict on changes

jules-ch:
Great, merging this.
Thanks @dalito

Issue #981, opened 2020-01-10
Numpy Supported functions


hgrecco:
This will be a tracking issue for numpy supported functions.

hgrecco:
numpy: 1.17.4
pint: 0.10.1
- [x] alen
- [x] all
- [ ] allclose
- [ ] alltrue
- [x] amax
- [x] amin
- [ ] angle
- [x] any
- [x] append
- [ ] apply_along_axis
- [ ] apply_over_axes
- [x] argmax
- [x] argmin
- [ ] argpartition
- [x] argsort
- [ ] argwhere
- [x] around
- [ ] array2string
- [ ] array_equal
- [ ] array_equiv
- [ ] array_repr
- [ ] array_split
- [ ] array_str
- [ ] asfarray
- [ ] asscalar
- [x] atleast_1d
- [x] atleast_2d
- [x] atleast_3d
- [x] average
- [ ] bincount
- [x] block
- [ ] broadcast_arrays
- [x] broadcast_to
- [ ] busday_count
- [ ] busday_offset
- [ ] can_cast
- [ ] choose
- [x] clip
- [x] column_stack
- [ ] common_type
- [x] compress
- [x] concatenate
- [ ] convolve
- [x] copy
- [x] copyto
- [ ] corrcoef
- [ ] correlate
- [x] count_nonzero
- [ ] cov
- [x] cross
- [x] cumprod
- [x] cumproduct
- [x] cumsum
- [ ] datetime_as_string
- [ ] delete
- [ ] diag
- [ ] diag_indices_from
- [ ] diagflat
- [x] diagonal
- [x] diff
- [ ] digitize
- [x] dot
- [ ] dsplit
- [x] dstack
- [x] ediff1d
- [x] einsum
- [ ] einsum_path
- [x] empty_like
- [x] expand_dims
- [ ] extract
- [ ] fill_diagonal
- [x] fix
- [ ] flatnonzero
- [x] flip
- [ ] fliplr
- [ ] flipud
- [x] full_like
- [ ] fv
- [ ] geomspace
- [x] gradient
- [ ] histogram
- [ ] histogram2d
- [ ] histogram_bin_edges
- [ ] histogramdd
- [ ] hsplit
- [x] hstack
- [ ] i0
- [ ] imag
- [ ] in1d
- [ ] inner
- [x] insert
- [x] interp
- [ ] intersect1d
- [ ] ipmt
- [ ] irr
- [ ] is_busday
- [x] isclose
[...]

PR #1410, opened 2021-11-06
Implement broadcast arrays

djpugh:
- [x] Improves #981

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ Pint Changelog
 0.19 (unreleased)
 -----------------
 
-- Nothing changed yet.
+- Implement numpy broadcast_array (Related to issue #981)
 
 
 0.18 (2021-10-26)
--- a/pint/numpy_func.py
+++ b/pint/numpy_func.py
@@ -715,7 +715,7 @@ def _broadcast_arrays(*args, **kwargs):
     input_units = []
     unitless_args = []
     for x in args:
-        if hasattr(x, 'units'):
+        if hasattr(x, "units"):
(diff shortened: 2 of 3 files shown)


MichaelTiemannOSC:
I think I need this.  I want to be able to broadcast a quantity to a dataframe thusly:

x['foo'] = Q_(1.0, ureg('m')

Presently I get this error message:

`TypeError: object of type 'int' has no len()`

If I manually create a PintArray based on the length of X and do some extra work I can cause my scalar quantity to become part of the X dataframe.  But a broadcast feature would be wonderful.

djpugh:
@hgrecco  - the previous tests failed on install of uncertainties==3.0.1

keewis:
> `TypeError: object of type 'int' has no len()`

Note that you can work around this by setting `ureg.force_ndarray_like = True` or `ureg.force_ndarray = True`

djpugh:
@hgrecco this seems to be passing and ready to merge - any other comments?

andrewgsavage:
> @hgrecco this seems to be passing and ready to merge - any other comments?

could you change the line in CHANGES to be for the newest version?

andrewgsavage:
I implemented broadcast_arrays after this PR (didn't see this one, my bad) 
Looks like the implementation in this PR doesn't work with the test case I made, so this will need looking into

djpugh:
@andrewgsavage, I've not had to use this functionality in a while, although I'm now working on something similar to last time I hit this error (and implemented this). Happy to look at it again if it's still an issue

PR #1434, opened 2021-12-09
Define microns as 1e-6 m

### LINKED ISSUE #1433: Automatically alias definitions where 1 X = 1 Y, and hash-equivalent aliases
Example: microns and micrometers are the same thing: i.e. `Quantity(1, "micron") == Quantity(1, "micrometer")` because in the default registry configuration, there's the line `micron = micrometer`. However, `Unit("micron") != Unit("micrometer")`, i.e. they are not aliases, unless you set it in your local registry.

Would it be possible to have unit definitions for `X` and `Y` where `N X = N Y` automatically be added as an alias as well?

Additionally, while aliases affect `__eq__`, they don't affect `__hash__`, which means they are treated differently in sets and dicts. Would it be too breaking a change to implement this, possibly by determining a normative name for all unit aliases (in the same way that abbreviations map to the full unit name) to hash against?

clbarnes:
This resolves #1433, the non-equality with micrometers.
Also includes a test which previously failed and now passes.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@ Pint Changelog
 - Upgrade min version of uncertainties to 3.1.4
 - Fix setting options of the application registry (Issue #1403).
 - Fix Quantity & Unit `is_compatible_with` with registry active contexts (Issue #1424).
+- Fix equivalence between `Unit("micron")` and `Unit("micrometer")`
 
 
 0.18 (2021-10-26)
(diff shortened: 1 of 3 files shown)


clbarnes:
This passes locally. However, I did find that I had to `pip install` the package again every time I changed the definitions file, even when the package was installed in editable mode. Without doing that, the test would pass in VScode, but fail when I called `pytest` (even in the terminal within VScode). This may be due to how package data files are handled, and this failure may relate to that by way of the pip cache.

jules-ch:
We're install pint before testing with `pip install .`, I'll look into it.

dalito:
I just tried this in a standard console (cmder/cmd on Win10). There it is not necessary to reinstall pint after changing the default registry file (`default_en.txt`). May it be related to VScode? A colleague had a somewhat similar problem with terminals in VScode this week.

hgrecco:
Yes, but I am anyway keeping part of the test.

PR #1436, opened 2021-12-09
Typehint wrapping

### LINKED ISSUE #1431: Type checking problem using function wrapping
Hello,

using pint 0.18, I found that writing a wrapper with mulitple arguments and ignoring some of them (following section "Ignoring an argument or return value" in [https://pint.readthedocs.io/en/stable/wrapping.html](url)),
```
from typing import Tuple
import numpy as np
from pint import UnitRegistry, Unit

ureg = UnitRegistry()

@ureg.wraps((None, Unit(ureg.mm)), (None, Unit(ureg.mm)))
def f(x: np.ndarray, y: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
    return x + y, x * y
```
static code checking via `mypy` for both arguments of the wrapper raises an error:
> Argument has incompatible type "Tuple[None, Unit]"; expected "Union[str, Unit, Iterable[str], Iterable[Unit], None]"

I think the type definition should maybe look like
`Union[str, Unit, None, Iterable[Union[str, Unit, None]]]`
to match the documentation's example.

**Note:** Actually, when setting up the wrapper *exactly* as illustrated in the examples without an additional `Unit()` around the `ureg.mm`, `@ureg.wraps((None, ureg.mm), (None, ureg.mm))`, the error would be
> Argument has incompatible type "Tuple[None, **Any**]"; expected "Union[str, Unit, Iterable[str], Iterable[Unit], None]".

However, using e.g. `@ureg.wraps((ureg.mm, ureg.mm), (ureg.mm, ureg.mm))` does not raise a `mypy` error. Just included that to make sure, the `None` was the source of the error raised, not the `Any` instead of a `Unit`.

Suseko:
- [x] Closes #1431  

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -45,6 +45,7 @@ Other contributors, listed alphabetically, are:
 * Ryan Dwyer <ryanpdwyer@gmail.com>
 * Ryan Kingsbury <RyanSKingsbury@alumni.unc.edu>
 * Ryan May
+* Sebastian Kosmeier
 * Sigvald Marholm <sigvald@marebakken.com>
 * Sundar Raman <cybertoast@gmail.com>
 * Tiago Coutinho <coutinho@esrf.fr>
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Pint Changelog
 0.19 (unreleased)
 -----------------
 
+- Fix type hints of function wrapper (Issue #1431)
 - Upgrade min version of uncertainties to 3.1.4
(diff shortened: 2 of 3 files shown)


jules-ch:
Since it is your first contribution can you add your name in the AUTHORS file.

Suseko:
Ok, did that. Thanks a lot, also for the motivation to submit something :)

jules-ch:
Thanks ! :)

PR #1454, opened 2022-01-23
Update hour symbol to `h`

### LINKED ISSUE #719: "h" should be the abbreviation of hours
"hr" is used as an abbreviation for hours, but this is not according to standard (see https://www.bipm.org/en/CGPM/db/9/7/), and much less common than "h" worldwide.

Here "h" is used for the Planck constant, which I think has much less use than the hours symbol. Both "h" and "hr" should be used for hours, to avoid misunderstanding. Planck should be called something else.

jules-ch:
- Replace h symbol for planck constant to  `ℎ`
- Add breaking change in CHANGES

- [x] Closes #719


CHANGES
```diff
@@ -19,6 +19,8 @@ Pint Changelog
 
 ### Breaking Changes
 
+- Update hour default symobl to `h`. (Issue #719)
```
dopplershift, line 22: "symobl" -> "symbol"

jules-ch:
@hgrecco tell me if that's ok with you.

hgrecco:
I agree with this change. But I think with this change!

hgrecco:
Sorry, I was very tired that day :-D  I was saying that it should be documented as breaking change and then I noticed that was the case!

jules-ch:
Alright I'm gonna merge this then.

PR #1463, opened 2022-02-09
Fix to_compact, infer_base_unit for non-float non_int_type (#1461)

### LINKED ISSUE #1461: `infer_base_unit` and `to_compact` break with `non_int_type=Decimal` and Decimal Quantity, dependent on prior actions.
When using `non_int_type=Decimal`, some unit conversions fail, some don't fail, and some unit conversions that fail initially start working after certain others are made.

Whether conversions fail appears to depend on how the quantities are instantiated.  Using string parsing with the UnitRegistry, eg, `ureg("1000.0 nm")` will often work.  Using a direct instantiation, eg, `ureg.Quantity(Decimal("1000.0")`, will not work, even though the two are equal.  However,  a conversion done on the first will cause future conversions of the second to work.  As a minimal example:

```python
import pint
from decimal import Decimal
print(ureg.Quantity(Decimal("1000.0"), "1.0 nm").to_compact()) # fails with TypeError: unsupported operand type(s) for *: 'float' and 'Decimal'
print(ureg("1000.0 nm").to_compact()) # Doesn't fail
print(ureg.Quantity(Decimal("1000.0"), "1.0 nm").to_compact()) # Now this doesn't fail
```

cgevans:
to_compact used math.log10, which converted non-float types to floats;
this caused exceptions when using a non-integer type, like Decimal,
that does not implement multiplication/division with floats, and
generally changed non-integer types to floats.  This addresses the
problem by checking non_int_type.  If non_int_type != float, it
assumes that the type implements a log10 method, as Decimal does.

infer_base_unit created a generic UnitsContainer, not a UnitsContainer
specific to a registry; this allows specification of a registry, or
uses the registry of the passed-in Quantity/UnitLike.

- [x] Closes #1461
- [ ] ~~Documented in docs/ as appropriate~~ (probably N/A)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,7 @@ Pint Changelog
 - Use default numpy `np.printoptions` available since numpy 1.15.
 - Implement `numpy.nanprod` (Issue #1369)
 - Fix default_format ignored for measurement (Issue #1456)
+- Fix `to_compact` and `infer_base_unit` for non-float non_int_type.
 
 ### Breaking Changes
 
--- a/pint/quantity.py
+++ b/pint/quantity.py
@@ -867,18 +867,10 @@ class Quantity(PrettyIPython, SharedRegistryObject, Generic[_MagnitudeType]):
         else:
             unit_str, unit_power = units[0]
 
-        if self._REGISTRY.non_int_type is float:
-            log10_abs_m = math.log10(abs(magnitude))
-        else:
-            # We don't necessarily know how to compute log10 for an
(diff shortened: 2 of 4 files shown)


OrangeChannel:
Does this break behavior with fractions.Fraction? And should we instead have a try/except block to catch other arbitrary numerical types that don't define that log10 method?

jules-ch:
Thanks @cgevans, looks good ! 👍

jules-ch:
As you said, it is only used to get the best suitable prefix, I think casting to float is enough there.

Issue #1465, opened 2022-02-13
Major refactoring of definition parsing and loading


hgrecco:
**Background**: A separate, text based definition file has been always one of Pint's strong points (and one of the reason it was created). But the definition file evolved from a very simple thing containing prefixes, multiplicative units and constants to a richer albeit more complex one. Extra code was added to the Registry to accommodate these new features. Recent discussions regarding speeding up registry instantiation sparked some discussions about caching to disk. But also, and more importantly, about how are we storing information in the registry. It is time to rethink how we do this and clean things up.

**Current situation**: Upon (default) registry instantiation, the following steps occur:
1. `default_en.txt` and `constants_en.txt` are *parsed*. (~ 70 ms)
2. Definitions are *loaded* into the registry. (~ 10 ms)
3. The *RegistryCache* is built. (~ 60 ms)

Steps (1) and (2) are actually intertwined, as definitions are loaded as they are read. Some definitions are parsed into a definition object (UnitDefinition, AliasDefinition, PrefixDefinition, DimensionDefinition) and then loaded into the Registry, while others (such as systems, groups and contexts) are handled by the final classes that will operate with the quantities. Finally, directives such as `@import` and `@defaults` are hacked into the registry.

In the definition file, there are three type of definitions (this is not explicit in the code)
- statements: units, prefix, constants, dimensions
- single line directives: alias, import (start with and `@`)
- block directive: defaults, systems, groups and contexts (start with and `@` and end with `@end`)

**Proposal**: 
- Separation of concerns: 
   - make a parser that parses into definition objects (like UnitDefinition) but for all type of things found on a definition file.
   - define loaders that take these definition objects and put them in the registry
- Definition object to communicate between parser and loader
   - immutable (as much as we can do this with Python)
   - avoid members that are registry objects (Quantity, Unit, Measurement)
   - avoid `<local> objects`
   - pickable
- Definition file object as a parsed version of a text definition file.
   - contains information about the original file
   - contains definition objects
   - contains line numbers for debugging
   - pickable
- Infrastructure to cache in disk.
   - can generate a filename from information about the source that is going to be cached.
   - invalidates the cache file if the source is newer

**What is done** in #1464:
- new `parser.py` module
   - ImportDefinition class: represents a parsed `@import` directive
   - DefinitionFile class
   - defines Parser class
      - parses and handles the `@import directive`
      - new single line and block directives can be registered
      - can load package resources and files
[...]

PR #1468, opened 2022-02-14
Update testsuite to avoid a complete fail when the UnitRegistry is faulty

hgrecco:
Under no circumstances a registry should be instantiated in a
module outside a fixture to avoid error during collection.
This precludes running simple tests that do not depend on
the registry.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -77,8 +77,12 @@ We use pytest_ for testing. If you contribute code you need to add tests:
 - When using classes, derive from `QuantityTestCase`.
 - Use `parametrize` as much as possible.
 - Use `fixtures` (see conftest.py) instead of instantiating the registry yourself.
-- Checkout `helpers.py` for some convenience functions before reinventing the wheel.
+  Check out the existing fixtures before creating your own.
 - When your test does not modify the registry, use `sess_registry` fixture.
+- **Do not** create a unit registry outside a test or fixture setup.
+- If you need a specific registry, and you need to reuse it create a
+  fixture in your test module called `local_registry` or similar.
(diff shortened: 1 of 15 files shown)


hgrecco:
I run into this when working on #1465 Every problem in the registry rendered the whole test suit useless. This PR fixes this, add some lines to the docs about this problem and leave a few TODOs to finish the migration to pytest fixture usage.

ping @jules-ch

jules-ch:
I'm ok for this.

I think we should rationalize pytest fixture, having a registry fixture for each pytest scope is overkill.

```
@pytest.fixture
def func_registry():
    return pint.UnitRegistry()


@pytest.fixture(scope="class")
def class_registry():
    """Only use for those test that do not modify the registry."""
    return pint.UnitRegistry()


@pytest.fixture(scope="module")
def module_registry():
    """Only use for those test that do not modify the registry."""
    return pint.UnitRegistry()


@pytest.fixture(scope="session")
def sess_registry():
    """Only use for those test that do not modify the registry."""
    return pint.UnitRegistry()
```

But we can adress this in another PR.
Sorry for the late answer.

hgrecco:
Indeed we need to revise how we organize the fixture. The current situation was done too quickly migrate from UnitTest to pytest. Briefly, the `func_registry`, `class_registry` and `module_registry` were used depending if `ureg = UnitRegistry()` was called within a function, on a class setup method or a top level module variable. `sess_registry` was then added and some tests that had no side effects (*) were moved to use resulting in a large speed up.

This could be improved in future PRs.

(*) due to the cache all tests have side effects on the registry, it would be nice to disable this.

Issue #1142, opened 2020-08-10
Quantity __new__ not parsing units string properly if some metric terms are abbreviated and some are not


klindsay28:
When I run the code
```
from pint import UnitRegistry
ureg = UnitRegistry()
unit_str = "cm centimeter cm centimeter"
print(ureg.Quantity(1.0, units=ureg(unit_str)))
print(ureg.Quantity(1.0, units=unit_str))
```
I get the output
```
1.0 centimeter ** 4
1.0 centimeter ** 2
```

I expect the output
```
1.0 centimeter ** 4
1.0 centimeter ** 4
```

It seems like if `units` is a string, then `cm` in `units` is getting ignored if `centimeter` is present.

This is with `pint 0.14`.

PR #1476, opened 2022-03-01
Fix parsing of units string with same canonalized name

### LINKED ISSUE #1441: Unit object gives wrong unit when mixing aliases
When creating a unit object with mixed aliases (`ureg.Unit('hour/hr')`), it returns a wrong unit. It seems like part of it gets ignored. This does not happen when creating a quantity as `ureg.Quantity('1 hour/hr')`.

Replicating code:
```
import pint
ureg = pint.UnitRegistry()
print(ureg.Unit('hour/hr'))       # prints "<Unit('1 / hour')>"

print(ureg.Quantity('1 hour/hr')) # prints "<Quantity(1.0, 'dimensionless')>"
```

In my model, the user sometimes uses units like "Gt CO2/yr" and "trillion USD/year". Costs per gigaton of CO2 become messed up because the "yr/year" doesn't disappear. I cannot switch to `Quantity`, since I use pint through the optimisation package Pyomo, which uses `Unit`.

jules-ch:
- [x] Closes #1441& #1142

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -26,6 +26,7 @@ Pint Changelog
 - Add `pint.testing` with functions to compare pint objects in tests (Issue #1421).
 - Fix handling modulo & floordiv operator in pint_eval (Issue #1470)
 - Fix `to_compact` and `infer_base_unit` for non-float non_int_type.
+- Fix parsing of units string with same canonalized name (Issue #1441 & #1142)
 
 ### Breaking Changes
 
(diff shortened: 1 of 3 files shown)


keewis (approved):
looks good to me

jules-ch:
Mypy support implicit `Optional` but explicit is prefered.

https://stackoverflow.com/questions/62732402/can-i-omit-optional-if-i-set-default-to-none

hgrecco:
Looks good to me!

jules-ch:
Thanks for the quick review !

PR #1478, opened 2022-03-01
Remove most pytest warnings

jules-ch:


keewis:
if you can fix the linting errors this should be ready for merging

jules-ch:
Now docs build is failing because of https://github.com/jupyter/nbconvert/issues/1736

keewis:
yes, to fix that you can pin `jinja2<3.1` until `nbconvert` is updated, but the interesting part is: RTD does not fail. Maybe we should just abandon the `Documentation Build` CI?

PR #1491, opened 2022-03-28
Update black version

### LINKED ISSUE #1490: Update black to stable version

jules-ch:
- [x] Closes #1490 


keewis:
I just noticed the autoupdate CI is failing because the PR creation action changed / broke. I'm not really motivated to update it, though, because I switched my own projects to `pre-commit.ci`. It has been really convenient to work with, so I guess I recommend enabling it here, too?

An additional advantage is that it will automatically commit any changes made by the linters / autoformatters, which, while slightly annoying for someone using `pre-commit` locally, is really useful for people who don't use it (i.e. mostly new developers)

jules-ch:
yeah, I've seen that I haven't had time to look into it.
I let @hgrecco decide about enabling `pre-commit.ci`.
I will look into disabling the doc build ci if it's a duplicate with the read the docs one.

hgrecco:
In principle, I am +1 on adding `pre-commit.ci`, but I want to understand this phrase "while slightly annoying for someone using pre-commit locally".

keewis:
that might have been an overstatement: the only issue I ever noticed was that it adds commits to the PR, and that I have to remember to pull before adding new commits. If I add commits before pulling, git will add merge commits, which is something I try to avoid. However, git allows to rebase instead, so I guess I should just look into configuring git differently?

hgrecco:
Then I agree with enabling this!

jules-ch:
That might be configurable no ?

jules-ch:
https://pre-commit.ci/#configuration-autofix_prs

keewis:
As you say, `pre-commit.ci` also allows disabling the autofix bot, but I do think that's really useful for new contributors. So I really should just look into fixing that for me locally.

Issue #1381, opened 2021-09-02
Wrong calculation when using temperature prefixes 


Christian-Mueller:
I'm doing following conversion:

```
from pint import UnitRegistry

ureg = UnitRegistry()
t = ureg.Quantity(20, ureg.degC)
print(t.to(ureg.millidegC))
```

output:
`293150.0 millidegree_Celsius`

It seems like prefixes for degree Celsius are not handled correctly. The result should be 20000 millidegree_Celsius.
Looks like it changes the unit to Kelvin internally.
It this a known issue? Is there a workaround for my use case?

jules-ch:
It seems millidegree_Celsius is not identified as offset units in the convert stage hence the value does not make sense afterwards.

I'll investigate

Christian-Mueller:
Any progress so far on this bug?

jules-ch:
The issue has been identified but the implementation is tricky with the current implementation:

https://github.com/hgrecco/pint/blob/c99ed1f537dea91af36c756dccef49e26d4f6447/pint/registry.py#L1529-L1550

Here the conversion for offset untis is handled like so:

- Value is converted to the reference unit of the source unit which is Kelvin for all temperature unit
- Conversion for remaining unit is done for non offset dimensions
- Value is converted to destination unit from reference unit of destination unit which is Kelvin for all temperature unit


In the case of prefixed units:
for DegC to mDegC

- Value is converted to the reference unit which is kelvin
- Conversion for remaining unit is done for non offset dimensions
- Value is converted from reference unit which is DegC for mDegC

calculation does output the wrong value, for those step

- 20 -> 293.15
- 293.15 -> 293.15
- 293.15 -> 293150.0

I'll probably end up with a fix for only those prefixed temperature units.

Thoughts @hgrecco.

I find it difficult to work with UnitDefinition, UnitsContainer there

dalito:
I also thought a bit about this: mDegC is the same as 1e-3 * degC which is a multiplication that is forbidden by default. So to be consistent pint should in my opinion raise errors for 

1. any kind of conversion to an SI-prefixed offset unit and
2. creation of quantities with an SI-prefix (`ureg.Quantity(20, ureg.milli_degC)`).

hgrecco:
Moving this to #1754

PR #1509, opened 2022-04-12
WIP: Fix Offset conversion with prefix & conversions using offset units

### LINKED ISSUE #1504: Problems when converting values with offset and prefixes
I am happy to use your library, but I ran into problems when converting values with offset and prefix. Example:
```python
>>>from pint import UnitRegistry
>>>ureg = UnitRegistry(lines)
>>>ureg.Quantity(0, "kilodegC").to("degK")
<Quantity(0.0, 'kelvin')>
>>>ureg.Quantity(1, "kilodegC").to("degK")
<Quantity(1000.0, 'kelvin')>
```
After all, 0 * 1000 degrees Celsius is not equal to 0 Kelvin, just as 1000 degrees Celsius is not equal to 1000 degrees Kelvin. The same applies to user units used with offset. I hope the authors of the project will fix this problem.

jules-ch:
- [x] Closes #1504

Well it works but the implementation is so ugly.

It basically using an iterator to get converter from unit to base unit & making conversion from src to base & then from base to dst.

prefixed_src -> src_referenced -> ... -> base

base -> ... -> dst_reference -> prefixed_dst

It is similar to what happens with multiplicative, but this time I'm using an iterator instead of the recursive function used in get_roots_units with the accumulator.

I had to use some trickery for to identify the prefixed unit as non multiplicative.
I think we need to work on this, since if a converter is non multiplicative in this tree going back to base unit, well conversion will be non multiplicative.

If you guys have insights, I feel like this is clearly not OK in terms of clarity.

### DIFF
--- a/pint/definitions.py
+++ b/pint/definitions.py
@@ -11,11 +11,14 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Callable, Iterable, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Callable, Iterable, Iterator, Optional, Tuple, Union
 
 from .converters import Converter, LogarithmicConverter, OffsetConverter, ScaleConverter
 from .errors import DefinitionSyntaxError
 from .util import ParserHelper, UnitsContainer, _is_dim
+
+if TYPE_CHECKING:
+    from .registry import UnitRegistry
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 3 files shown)


OrangeChannel:
See https://github.com/hgrecco/pint/issues/1381#issuecomment-962446166 . I also don't think prefixing offset units should be allowed at all. Practically I get that a kilodegC is 1000 °C but an object at 1 kilodegC is not a thousand times hotter than something that is 1 degC right? The multiplication implies by prefixes doesn't physically make sense here.

PR #1525, opened 2022-05-21
Control autoconvert in test mul w/ scalar

coder0xff:
test_multiplication_with_scalar has been dependent on the prior state of self.ureg.autoconvert_offset_to_baseunit making the `(10, "deC")` case pass/fail intermittently. Set `self.ureg.autoconvert_offset_to_baseunit = False` at the start of the test. Note that this case will reliably fail now.


hgrecco:
I am not sure about this (not just your change, but the previous line as well). I think this could profit for migrating into a function to be sure that the registry is isolated.

coder0xff:
> I am not sure about this (not just your change, but the previous line as well). I think this could profit for migrating into a function to be sure that the registry is isolated.

My rationale here is that this test would pass or fail based on whether other tests in the fixture were previously run.

PR #1530, opened 2022-05-30
Fix changing number of significant places when converting units with Decimal magnitudes (See #1527)

### LINKED ISSUE #1527: Converting between units changes number of significant places of Decimal objects
When using Decimal objects, one expects that the precision in which one specified the original value is retained. However, when converting between units, I've noticed that precision is added:

```
import decimal

import pint

print(f"Pint version: {pint.__version__}")
print(f"Decimal version: {decimal.__version__}")

ureg = pint.UnitRegistry(non_int_type=decimal.Decimal)

x = ureg.parse_expression("2 microliter milligram/liter")
print(x.m)
print(x.to_compact().m)
print(x.to_base_units().m)
print(x.to("ng").m)

>>> Pint version: 0.19.2
>>> Decimal version: 1.70
>>> 2
>>> 2.000
>>> 2.000E-12
>>> 2.000
```

The expected behaviour should be that changing the units does not change the precision.  What I suspect might be happening is somewhere in the code the value is multiplied by another decimal object that has a higher precision, like in this example:

```
decimal.Decimal("2")*decimal.Decimal("1.000")
>>> Decimal('2.000')
```

I don't know enough about the codebase, but from a quick look maybe this behaviour might be caused from Line 1044 in `registry.py` in the `_convert` function. The comments for that bit of code seem to indicate that a factor that the value is being multiplied by is being converted to a Decimal, potentially with the level of precision that is a problem here:

```
# factor is type float and if our magnitude is type Decimal then
# must first convert to Decimal before we can '*' the values
    factor = Decimal(str(factor))
elif isinstance(value, Fraction):
    factor = Fraction(str(factor))

if inplace:
    value *= factor
else:
    value = value * factor
```

With some pointers I might be able to dig deeper and try to submit a pull request, but I'm a bit out of my depth right now. Any help appreciated!

noahsprent:
I am definitely not very familiar with pint, and haven't contributed much to open-source projects before, so please excuse any clumsiness in execution here! In #1527 @hgrecco suggested that `decimal.Decimal.Normalize` would work to ensure that additional significant places aren't introduced when converting quantities. 

This worked to fix the issue that I was having, so I thought I'd make it into a pull request, but I suspect that I might have missed something, as he also said "But I cannot think of a way of putting this in pint in simple way". Sorry if this is the case, and I'm creating more work then I'm helping with...

- [X] Closes #1527
- [X] Executed ``pre-commit run --all-files`` with no errors
- [ ] The change is fully covered by automated unit tests (Not very experienced with this, and I suspect that I have put them in the wrong place, sorry! But I tried.
- [X] Documented in docs/ as appropriate (Don't think that this needs an addition in the docs?)
- [X ] Added an entry to the CHANGES file (Again, don't think that I did this correctly but figured I'd do something instead of nothing)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@ Pint Changelog
 - Reorganized code into facets.
   Each facet encapsulate a Pint functionality.
   (See #1466, #1479)
+- Fix changing number of significant places when converting units with Decimal magnitudes (See #1527)
 
 
 0.19.2 (2022-04-23)
--- a/pint/facets/plain/registry.py
+++ b/pint/facets/plain/registry.py
@@ -1047,6 +1047,10 @@ class PlainRegistry(metaclass=RegistryMeta):
         # must first convert to Decimal before we can '*' the values
         if isinstance(value, Decimal):
             factor = Decimal(str(factor))
+
+            # we need to remove trailing zeros that would affect the number
+            # of significant places returned
+            factor = factor.normalize()
         elif isinstance(value, Fraction):
             factor = Fraction(str(factor))
 
--- a/pint/testsuite/test_converters.py
+++ b/pint/testsuite/test_converters.py
@@ -1,5 +1,9 @@
 import itertools
+from decimal import Decimal
 
+import pytest
+
+from pint import UnitRegistry
 from pint.compat import np
 from pint.converters import Converter
 from pint.facets.nonmultiplicative.definitions import (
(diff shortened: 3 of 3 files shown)


tovrstra:
I only notice now that you probably ran into the same problem as I did in #1621. I solved this in another way, taking away the cause rather than fixing it up later, see #1625, so I'm not sure if we're talking about the same problem. 

Can you try if my fix also solves your problem?

noahsprent:
Better fix in #1625

PR #1539, opened 2022-06-29
avoid the `RecursionError` raised by `np.where`

### LINKED ISSUE #1510: `RecursionError` with `numpy.where`
From xarray-contrib/pint-xarray#136: running
```python
import pint
import numpy as np

q = pint.Quantity([2, 3, 6, 5], "kelvin")
np.where(q.max(), q)
```
results in a `RecursionError`. I'm not sure this operation actually makes sense (the dimensions mismatch, for example), but we should definitely make sure this does not trigger a `RecursionError`.

keewis:
The easiest fix is to just strip the units of the condition. This might fail with offset units / give unexpected results, so instead we could also forbid passing a quantity (i.e. explicitly raise an error).

- [x] Closes #1510


jules-ch:
CI failing cause of https://github.com/dask/dask/issues/9137

I'd be in favor of enforcing condition as something that is not a Quantity. It will be both safer for us to debug & avoid unexpected results.
Thoughts @hgrecco

keewis:
I now think that while raising would be the cleanest way, it would be inconsistent with the remaining `numpy` functions dealing with truthiness (for `np.all` and `np.any` see #965). To stay consistent, we should probably raise only if the unit of the condition is a offset unit.

jules-ch:
Agree with you raising is the cleanest but for consistency with other boolean / filter related function we can push this change.

PR #1544, opened 2022-07-10
Added polynomial class and its tests

jsmitty-here:
- [N/A] Closes # (Not related to any known open issue)
- [] Documented in docs/ as appropriate

# Explanation 
I've been using pint for a project I am working on for fluid calculations. I needed a polynomial to model certain equations where the input and outputs of the polynomial have their own units; such as a pump curve which takes a flowrate unit and outputs a pressure unit, or vice versa. 

I built a Polynomial class inheriting from the numpy Polynomial class that incorporates x & y unit variables for my own project and thought this could be a helpful class for the pint library as well. I have implemented unittests for this class in the test_polynomial.py file.

This is my first commit to this repo, so please let me know if I need to make any changes before this PR can be merged.

# Example Uses
### 1. Example modeling a pump curve
```
flow_unit, pressure_unit = ureg.gallon / ureg.minute,  ureg.psi

pump_curve = Polynomial([46.029, -0.1546, -0.0053], flow_unit,  pressure_unit)

pump_curve.solve(0 * flow_unit)                                                # Returns: 46.029 pound_force_per_square_inch
pump_curve.y_intercept                                                             # Returns: 46.029 pound_force_per_square_inch
pump_curve.x_intercept                                                             # Returns: [-108.91112924   79.74131792] gallon / minute
pump_curve.solve_for_x(10 *pressure_unit , min_value=0))      # Returns: 69.144682905577 pound_force_per_square_inch
pump_curve.derivative_at(30 * flow_unit)                                 # Returns: -0.4726 minute * pound_force_per_square_inch / gallon

```

### 2. Example modeling a distance vs time polynomial
```
distance_unit, time_unit = ureg.meter,  ureg.second

distance_curve = Polynomial([0, 1, 2], time_unit ,  distance_unit)

distance_curve .solve(0 * distance_unit)                                          # Returns: 0.0 meter
distance_curve .y_intercept                                                             # Returns: 0.0 meter
distance_curve .x_intercept                                                             # Returns: -0.5 second
distance_curve .solve_for_x(10 * distance_unit, min_value=0))       # Returns: 1.9999999999999996 second
distance_curve .derivative_at(2 * distance_unit)                              # Returns: 9.0 meter / second

velocity_curve = distance_curve .derivative                             # Returns: Polynomial that is the derivative of the original
velocity_curve.solve_for_x(5 * distance_unit / time_unit )        # Returns: 1.0 second

acceleration_curve = velocity_curve.derivative                        # Returns: Polynomial that is the derivative of the original
acceleration_curve.solve(2 * time_unit)                                    # Returns: 4.0 meter / second ** 2

```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -33,6 +33,7 @@ Other contributors, listed alphabetically, are:
 * John David Reaver <jdreaver@adlerhorst.com>
 * Jonas Olson <jolson@kth.se>
 * Jules Chéron <me@julescheron.com>
+* Justin Smits <smits.justin98@gmail.com>
 * Kaido Kert <kaidokert@gmail.com>
 * Kenneth D. Mankoff <mankoff@gmail.com>
 * Kevin Davies <kdavies4@gmail.com>
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,12 @@
 Pint Changelog
 ==============
 
+0.20.1 (unreleased)
+------------------
+
+- Added Polynomial class that inherits from numpy's polynomial
+  class and incorporates unit for the x and y variables.
+
 0.20 (unreleased)
 ------------------
 
--- a/pint/polynomial.py
+++ b/pint/polynomial.py
@@ -1,0 +1,201 @@
+"""
+    pint.polynomial
+    ~~~~~~~~~~~~~~
+
+    A polynomial class inheriting from numpy.polynomial.polynomial.Polynomial incorporating the pint Quantity unit values.
+
+    :copyright: 2016 by Pint Authors, see AUTHORS for more details.
+    :license: BSD, see LICENSE for more details.
+"""
+
+from __future__ import annotations
+
+from math import inf
+from typing import Optional, Union
+
+from numpy.polynomial import polynomial as p
+
+from . import _DEFAULT_REGISTRY, Quantity, Unit
+
+
+class Polynomial(p.Polynomial):
+    def __init__(
+        self,
+        coef: list[float],
+        x_unit: Unit = _DEFAULT_REGISTRY.dimensionless,
+        y_unit: Unit = _DEFAULT_REGISTRY.dimensionless,
+    ):
+        super(Polynomial, self).__init__(coef)
+        self.x_unit = x_unit
+        self.y_unit = y_unit
+
+    @property
+    def y_intercept(self) -> Quantity:
+        return self.coef[0] * self.y_unit
+
+    @staticmethod
+    def _get_roots_of_polynomial(
+        poly: p.Polynomial,
+        x_min: float = -inf,
+        x_max: float = inf,
+        real_only: bool = True,
+    ):
+        roots: set = set(poly.roots())
+        if real_only:
+            roots = {float(root.real) for root in roots if root == root.real}
+        roots: list = list({root for root in roots if x_min <= float(root) <= x_max})
+        return roots
+
+    def _x_at_y(
+        self,
+        y_value: float,
+        x_min: float = -inf,
+        x_max: float = inf,
+        real_only: bool = True,
+    ) -> list[Union[float, complex]]:
+        solutions = self._get_roots_of_polynomial(
+            (self - y_value), x_min, x_max, real_only
+        )
+        if len(solutions) == 1:
+            return solutions[0]
+        return solutions
+
+    @property
+    def real_roots(self) -> list[float]:
+        return self._get_roots_of_polynomial(self, real_only=True)
+
+    @property
+    def positive_real_roots(self) -> list[float]:
+        return self._get_roots_of_polynomial(self, x_min=0, real_only=True)
+
+    @property
+    def x_intercept(self) -> Optional[Quantity]:
+        roots = self.real_roots
+        if len(roots) == 0:
+            return None
+        else:
+            root = min(roots)
+        return root * self.x_unit
+
+    def solve(self, value: Quantity, min_value: float = -inf) -> Quantity:
+        x = value.m_as(self.x_unit)
+        return max(self(x), min_value) * self.y_unit
+
+    def solve_for_x(self, y_value: Quantity, min_value: float = -inf) -> Quantity:
+        y = y_value.m_as(self.y_unit)
+        try:
+            return self._x_at_y(y, x_min=min_value, real_only=True) * self.x_unit
+        except TypeError:
+            raise ValueError(
+                "There are no values of {} that output {}!".format(self, y_value)
+            )
+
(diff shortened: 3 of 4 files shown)


hgrecco:
While I do agree that it is desirable to improve the numpy/scipy support, I would like to discuss a few things:
1. Is there a way to do it in a more direct manner (e.g. using wraps to avoid rewriting the internal code).
2. How should we organize it? Because it seems that this might explode easily. Maybe it is better to put it in a numpy support module?

jsmitty-here:
Thanks for the response. Regarding your discussion points:

1. Do you have a vision on how this would be done? I may be misunderstanding what you mean. By a 'wrap to avoid rewriting the internal code' I assume you are referring to the dunder methods? If so I will certainly need to investigate how I could change this. Maybe something like this:

```
from numpy.polynomial import polynomial as p

def numpy_polynomial(cls):
      
    class Wrapper:
         """ Handles Integration with numpy Polynomial base class """

    return Wrapper
  
@numpy_polynomial
class Polynomial(p.Polynomial):
    """ Handles Initialization and any new polynomial functions. """

    def __init__(self, coef: list[float], x_unit: Unit, y_unit: Unit):
        super().__init__(coef)
        self.x_unit= x_unit
        self.y_unit= y_unit
```
   Please let me know if that is not what you had in mind. 

2. Agreed, I could see mathematical implementations incorporating pint growing beyond polynomials. Are you suggesting the 'polynomial' module (and modules like it) should be organized into a numpy support directory within pint or be segregated from the main pint library? 

This is my first attempt at contributing to an open source library so I apologize for any informalities. Additionally, If the polynomial feature does not align with the overall goal of the pint library just let me know and I will take no offense. Like I said, I had already wrote this class and it's tests for a personal project I am working on.

jsmitty-here:
I may need some assistance with using wraps of this type. My initial attempt was to add the wraps into the class methods:
Converting a function like this:
```python
class Polynomial(p.Polynomial)
    ...
    def solve(self, value: Quantity, min_value: float = -inf) -> Quantity:
        return max(self(value), min_value) * self.y_unit
```
to this:
```python
class Polynomial(p.Polynomial)
    ...
    @staticmethod
    @ureg.wraps(self.y_unit, (self.x_unit, self.x_unit))
    def solve(value: float, min_value: float = -inf) -> float:
        return max(self(value), min_value)
```
however, I can not access the x_unit, and y_unit instance variables from self for the ureg.wraps nor can I use self inside the functions as the method now needs to be a static method to be compatible with the ureg.wraps decorator function. 

Additionally, I do not see how these ureg unit conversion wraps could be used on any of the functions or dunder methods which return 'Polynomial' as they return a new instance of the polynomial class, not a float or Quantity.

I do agree that there is a lot of repeating code considering the class primarily contains just a numpy polynomial, and it's unit values. I just have not discovered the best way to implement that yet.

andrewgsavage:
I've written something similar for water pump calculations too. My approach was to use pint-pandas to convert units to metric (any coherent unit system would work) and then drop units so I can work with floats. This allows me to use code from examples and documentation much easier than would be possible with the unit aware Polynomial class, which needs each method to be added and tested. I can convert units at the end if needed, although I've not needed to do this as metric has been sufficient.

Is this approach (using pint or pint-pandas to convert values to a coherent system) something you considered? I think curve fitting is a common use case for pint, and I don't think the pint documentation points people to convert then drop units. I wonder if it's something worth adding - this is how every FEA and MBD simulation package I've used deals with units. 

That being said, the Polynomial class is a nice idea and will be great to use when it's in a mature state. I think a really cool example would be to have two pumps curves with different units. Create a Polynomial object for each, and then can add the two Polynomial objects to get a pump curve for the two pumps in series.

jsmitty-here:
@andrewgsavage 
I have not yet investigated pint-pandas and maybe this could solve some of my issues. 

In the application I am building, I am hoping to not drop units at any point. Hence the creation of a polynomial class with the units caked in. 
My main driver for making this polynomial class was for pump curves and generating system curves (fitting a polynomial to the pressure drop through a piping system at a range of flowrates), however, I see how it could be used for any polynomial physics equations where units matter (see my first comment where the derivative of the 'distance_curve', with units 'second' vs 'meter', generated a new polynomial with units 'second' vs 'meter / second') . 

Additionally, I envision the units of the pump curve not actually being important thanks to Pint. The user could input any Quantity (of comparable units) and convert the returned Quantity as needed. For example:
```python
# Define units for simplicity
GPH = ureg.gallon / ureg.hour
PSI = ureg.psi
LPM = ureg.liter/ ureg.minute
PASCAL = ureg.pascal

curve_1 = Polynomial([20, 0, -0.05], GPH, PSI)      # f( x GPH) = y PSI = 20.0 - 0.05 * x ** 2
curve_2 = Polynomial([30, -0.1, -0.02], LPM, PASCAL) # f( x LPM) = y PASCAL= 30.0 + -0.1* x - 0.02 * x ** 2

# The class handles the unit conversion for the input argument internally
curve_1.solve(1 * GPH)    # 19.95 pound_force_per_square_inch
curve_1.solve(1 * GPH).to(PASCAL)    # 137550.41 pascal
curve_1.solve(1 * LPM)    # 7.44 pound_force_per_square_inch
curve_1.solve(1 * LPM).to(PASCAL)    # 51285.71 pascal

curve_2.solve(1 * GPH)    # 29.99 pascal
curve_2.solve(1 * GPH).to(PSI)    # 0.0044 pound_force_per_square_inch
curve_2.solve(1 * LPM)    # 29.88 pascal
curve_2.solve(1 * LPM).to(PSI)    # 0.0043 pound_force_per_square_inch
```

Now if the user actually wants to convert the entire polynomial to a new unit system, I have a method like such:
```python
curve_1 = Polynomial([20, 0, -0.05], GPH, PSI)      # f( x GPH) = y PSI = 20.0 - 0.05 * x ** 2
# May rename this method from "to_units()" to "to()" to match the Quantity's "to()" unit conversion method.
curve_1.to_units(x_unit=LPM, y_unit=PASCAL)      # f(x LPM) = y PASCAL = 137895.14586336722 − 86609.4395918349 * x ** 2
```

In the case of adding two (or more) pump curves:
```python
curve_1 = Polynomial([20, 0, -0.05], GPH, PSI)      # f( x GPH) = y PSI = 20.0 - 0.05 * x ** 2
curve_2 = Polynomial([30, -0.1, -0.02], LPM, PASCAL) # f( x LPM) = y PASCAL= 30.0 + -0.1* x - 0.02 * x ** 2

combined_curve = curve_1 + curve_2    # + ... + curve_N
print(combined_curve)    # f( x GPH) = y PSI = 20.004351132131905 - 9.150459358810584e-07 * x - 0.05000001154608556 * x ** 2
```

PR #1569, opened 2022-08-19
Update contributing.rst

edouardberthe:
It seems that the [`test`](https://github.com/hgrecco/pint/blob/master/setup.cfg#L36-L40) extra requirements section is needed to run test locally, when willing to contributing.

(I still have 3 tests failing even and 13 xfailed but that's probably another pb)


jules-ch:
Thanks ! :)

PR #1595, opened 2022-09-30
Final step to split the registry from the parser

hgrecco:
Overview:
- All the code in facets is now independent of the definition textual format. In particular, defintions such as UnitDefinition, ContextDefinition an so on cannot be built directly from a string. (some functions are kept only temporarily to simplify but transition) Building Definition objects from string requires a parser that emits them.
- The standart pint format is implemented in delegates/txt_parser using flexparser. Briefly each single line statement is mapped to a ParsedStatement class and each larger construct to a Block class.
- The registry then has an adder function that takes a definition an incorporate it into the registry.

A few nice features of this approach:
1. The Definition objects are standalone public objects, you can now build them in a programatic way and incorporate them to the registry using the define function that will dispatch to the correct adder:

      >>> new_unit = UnitDefintion( ....)
      >>> ureg.define(new_unit)  # might be called add in the future

   No more being forced to use string definitions (but you can still use them if you want)

2. Composition over inheritance. The Registry does not know how to parse a definition, but it delegates this to another class which can be changed. This makes it very easy to write another parser (faster, simpler) o try out a completely different file format.

3. Error messages can be more meaningful.

Backwards incompatible changes
- is_base parameter Definitions is not needed any more. It is now computed automatically leading to a leaner experience and also avoiding incompatible states
- alias for dimensionality has been removed (for now at least) The only one defined was speed as an alias of velocity.
- (Context|Group|System).from_lines and Definition.from string have been rewritten in terms of the new parser. But will be likely removed in the future
- Changing non_int_type is not possible after registry has been created
- load_definition raises FileNotFoundError instead of a generic exception if the file was not found
- the string representation of several definitions is now not so user friendly terms of the new parser. But will be likely removed in the future
- Changing non_int_type is not possible after registry has been created
- load_definition raises FileNotFoundError instead of a generic exception if the file was not found
- the string representation of several definitions is now not so user friendly. terms o


hgrecco:
As far as I can tell, the only error is due to Exceptions in dask not playing nice with being frozen instance classes. But we can get rid of that or try to fix it withe pickle version or using pickle related magic methods.

hgrecco:
> As far as I can tell, the only error is due to Exceptions in dask not playing nice with being frozen instance classes. But we can get rid of that or try to fix it withe pickle version or using pickle related magic methods.

confirmed, all test are passe using `frozen=False` in exceptions.

hgrecco:
A colleague suggested changing `txt_parser` to `txt_defparser`, `Parser` to `DefParser` and so on. Parser will be kept only to parse the mathematical expressions (i.e. ParserHelper) in which there is no other format possible (nobody is going to use reverse polish notation I guess!)

hgrecco:
@jules-ch What do you think about merging this.

hgrecco:
I would like to merge this by the end of the week, so any other change does not diverge too much. Any opposition?

PR #1603, opened 2022-10-09
Update test_non_int.py for pytest.

cgevans:
Tests in test_non_int.py were not being discovered by pytest because the class names did not start with Test; they also had a few incompatibilities with pytest, which were hidden as a result of the tests not being run.  This commit both enables the tests to be discovered and updates them for pytest:

- Class names are changed.
- Dead code (FakeWrapper) is removed.
- Non-float assert_quantity_almost_equal is fixed.
- Parametrization, capture_log, subTests are updated for pytest.
- kwargs/constants are updated for pytest.
- caplog in test_quantity_creation tests for specific message.
- Class inheritance is changed to properly run setup_class and special assert_quantity_almost_equal.

- [ ] Closes # (insert issue number) (N/A, but see #1600.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,7 @@ Pint Changelog
 - Move ASV benchmarks to dedicated folder. (Issue #1542)
 - Fix a recursion error that would be raised when passing quantities to `cond` and `x`.
   (Issue #1510, #1530)
+- Update test_non_int tests for pytest.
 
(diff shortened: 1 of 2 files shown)


jules-ch:
Thanks, it was a side effect of transitioning to pytest. Discovery of those test disappeared.
I'll do a quick review, but this looks good to me so far.

PR #1608, opened 2022-10-11
Fix setitem with a masked array with multiple items (Fixes #1584)

### LINKED ISSUE #1584: `Quantity.__setitem__` skips unit check when given a mask array
This works:
```python
import numpy as np
import pint

masked = np.ma.array([1,2,], mask=[True, False])
np.ma.is_masked(masked)

q = pint.Quantity(np.ones(shape=(2, )), 'm')
q[:] = masked  # no unit, but accepted

masked_q = pint.Quantity(masked, 'mm')
q[:] = masked_q  # wrong unit, but not checked
```
Offending code is here: https://github.com/hgrecco/pint/blob/208d36a5bb051465296dad846296bb19628e2307/pint/facets/numpy/quantity.py#L246-L249

Not sure how this `np.ma.is_masked` was intended?

dopplershift:
This was incorrectly passing through some non-masked values.

- [x] Closes #1584


pint/facets/numpy/quantity.py
```diff
@@ -245,7 +245,10 @@ def __getitem__(self, key):
 
     def __setitem__(self, key, value):
         try:
-            if np.ma.is_masked(value) or math.isnan(value):
+            # If we're dealing with a masked single value or a nan, set it
+            if (
+                getattr(value, "size", 0) == 1 and np.ma.is_masked(value)
+            ) or math.isnan(value):
```
SimonHeybrock, line 251: I think this is fixing one of the issues, but it will still replace the current (often not masked) magnitude with the magnitude of an arbitrary masked quantity. I think this would at the very least also have to check if `self[key]` is masked, but even then I would argue that ignoring the unit is not conceptually correct (it is the values that are masked not the units).

dopplershift: And again I argue that you can convert a masked value to whatever you want--in the end the result is still masked. 

IMO, Pint's primary purpose is *tracking* and *properly converting*. Catching gross errors in user code with the `DimensionalityErrors` that result is a byproduct of that, not a primary feature to me.

What I've done here keeps things in line with the way `NaN` is treated, which is a design decision that pre-dates me (but I happen to agree with).

If you would like a different approach, I would suggest putting up your own PR that makes Pint behave how you'd like. At that point the merits of the new behavior (and any possible breakage/change to existing tests) can be considered in detail there.

dopplershift:
Part of this is motivated by the way masked arrays are implemented. I want this to work without error and warning:
```python
import numpy as np
import pint
a = pint.Quantity(np.ma.array([1, 2, 3], mask=False), 'm')
a[1] = np.ma.masked
```

That's a canonical way to set the mask of an item. There's no need for units to be involved there, because you're not even setting a value; it's using a sentinel value as a way to set the mask on array, which has no units.

pint/facets/numpy/quantity.py
SimonHeybrock: Does not looked masked to me:
```
import pint
import numpy

q1 = pint.Quantity(np.array([1,2]), 'm')
q2 = pint.Quantity(np.ma.array([3,4], mask=[True, True]), 'mm')
q1[0:1] = q2[0:1]
print(q1)  # [3 2] meter
```
We have now converted "masked 3 millimeter" to "3 meter"

dopplershift: That's correct and not good, but it's also exploit not-great behavior by numpy, by which you can slice a masked array, stuff it in a regular array, and the mask is silently dropped. 😬 I'll add a check to make sure our value can actually handle masking.

dopplershift:
So I've corrected my patch and the code that matches the `isnan` handling to check for `_magnitude` being an underlying `MaskedArray`.

The broader problem identified by @SimonHeybrock above is complicated by the fact that while I think Pint is doing the correct thing by converting units, the problem is that `MaskedArray` is causing the problem by not performing the mathematics needed for the conversion on the masked value. I'm not sure what exactly to do about that, and I think is outside the scope of this PR.

pint/facets/numpy/quantity.py
```diff
@@ -245,7 +245,12 @@ def __getitem__(self, key):
 
     def __setitem__(self, key, value):
         try:
-            if np.ma.is_masked(value) or math.isnan(value):
+            # If we're dealing with a masked single value or a nan, set it
+            if (
+                isinstance(self._magnitude, np.ma.MaskedArray)
+                and np.ma.is_masked(value)
+                and getattr(value, "size", 0) == 1
+            ) or math.isnan(value):
```
SimonHeybrock, line 253: Have you considered using
```suggestion
            if (value is np.ma.masked) or math.isnan(value):
```
Then your requirement that `a[1] = np.ma.masked` works should be fulfilled, but you may avoid the other problems.

dopplershift: I did contemplate that and ended up looking back when I contributed the original change. That specifically mentioned masked values. I think I can live with this version too (emphasis on *think*), but I'd want to hear thoughts from @hgrecco or any of the other core committers.

And for absolutely clarity: changing to that won't change the fact that setting using a masked array (with masked values) can completely subvert the units.

jules-ch:
First, I think operations with mixing masked & regular array is out of scope.
This lead to undefined behaviour IMO, even with numpy I think it returns inconsistent values, see:

```python
>>> import numpy as np
>>> q = np.ones(shape=(2,))
>>> q
array([1., 1.])
>>> q[:] = np.ma.masked
>>> q
array([0., 0.])
```

or 

```python

>>> import numpy as np
>>> masked = np.ma.array([1, 2], mask=[True, False])
>>> masked
masked_array(data=[--, 2],
             mask=[ True, False],
       fill_value=999999)
>>>q = np.ones(shape=(2,))
>>>q
array([1., 1.])
>>>q[:] = masked
>>>q
array([1., 2.])
```

Operations with mask should only be used on masked array.
If you check the numpy documentation, event the example shown in the issue seems out of scope.
Masked array should be initialized with a mask which can be modified. But stting slice with another masked array is never shown anywhere in the docs.

The use case we need to address is modifying the mask & fill_value accordingly with units.

Setting masked values or nan without unit checking.

```python
>>> import numpy as np
>>> q = np.ma.array([1, 2, 3, -1, 5], mask=False)
>>> q
masked_array(data=[1, 2, 3, -1, 5],
             mask=[False, False, False, False, False],
       fill_value=999999)
>>> q[0:2] = np.ma.masked
>>> q
>>> q
masked_array(data=[--, --, 3, -1, 5],
             mask=[ True,  True, False, False, False],
       fill_value=999999)
```

```python
>>> import numpy as np
>>> q = np.array([1.0, 2.0, 3.0, -1.0, 5.0], dtype=float)
>>> q
array([ 1.,  2.,  3., -1.,  5.])
>>> q[0:2] = np.nan
>>> q
array([nan, nan,  3., -1.,  5.])
```

jules-ch:
What do you think @dopplershift after giving my thoughts on this. I'd be willing to close this issue.
Maybe add some examples & a warning when using MaskedArray.

dopplershift:
@jules-ch  (I missed when you replied a month ago apparently) I agree with your overall plan, but want to point out that this PR fixes the `if` at the top of the block to only pass when given scalars, which is the intent. When I made the original patch, it was never supposed to pass through an array.

jules-ch (approved):

PR #1619, opened 2022-10-21
Adding compatability code for changes in importlib.resources. Closes #1618

### LINKED ISSUE #1618: importlib.resource.path (pint/parser.py:264) is deprecated in Python 3.11, use files() instead
```
../lib/python3.11/site-packages/pint/parser.py:264: DeprecationWarning: path is deprecated. 
Use files() instead. Refer to https://importlib-resources.readthedocs.io/en/latest/using.html#migrating-from-legacy for migration advice.
```

hwalinga:
- [x] Closes #1618
- [x] The change is fully covered by automated unit tests (I guess you will add 3.11 to tox setup.)
- [ ] Documented in docs/ as appropriate (not needed)
- [ ] Added an entry to the CHANGES file (just too minor)

### DIFF
--- a/pint/parser.py
+++ b/pint/parser.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 
 import pathlib
 import re
+import sys
 from dataclasses import dataclass, field
 from functools import cached_property
 from importlib import resources
@@ -261,8 +262,12 @@ class Parser:
         it will use python importlib.resources.read_binary
         """
 
-        with resources.path(__package__, resource_name) as p:
-            filepath = p.resolve()
+        if sys.version_info[1] == 8:  # Remove if dropping 3.8
+            with resources.path(__package__, resource_name) as p:
+                filepath = p.resolve()
+        else:  # 3.9 and above
+            with resources.as_file(resources.files(__package__)) as p:
+                filepath = (p / resource_name).resolve()
 
         if filepath.exists():
             if self._diskcache is None:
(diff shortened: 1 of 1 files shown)


hwalinga:
Alternatively you can use https://pypi.org/project/importlib-resources/, but that would add a new dependency to the project and there isn't that much use of it.

hgrecco:
Thanks for submitting this patch. I will merge the same concept into the new parser.

hgrecco:
Merged this concept into https://github.com/hgrecco/flexparser

Now flexparser is ready for Python 3.11 (and therefore pint will be able to parse resources without warnings)
https://github.com/hgrecco/flexparser/actions/runs/3301572697

As soon as Python 3.11 is released, I will release a new version of flexparser against it and incorporate the new version to `pint/_vendor`

Thanks!

hwalinga:
>As soon as Python 3.11 is released,

Just so you know, that is this Monday  ; )

Thanks for the fix!

hgrecco:
Thank you. I just pushed that we test the whole code base against 3.11-dev in #1622

hgrecco:
All tests are passing on python 3.11 but I still need to check the logs.

I think we should update the current tests to fail on DeprecationWarning and PendingDeprecationWarning. Anyone knows a global way to do it in pytests?

hgrecco:
and maybe even fail on all warnings (except those that are explicitly caught).

PR #1625, opened 2022-10-23
Fix creation of trailing zeros in Decimal objects when converting units

### LINKED ISSUE #1621: Unit conversion with `Decimal` sometimes introduces trailing zeros
The following example shows somewhat unexpected behavior:

```python
import pint
from decimal import Decimal
a = ureg.Quantity("5.0 mV/m")
b = a.to_base_units()
print(a)
print(b)
print(a.magnitude.as_tuple())
print(b.magnitude.as_tuple())
```

This prints out

```
5.0 millivolt / meter
0.005000 kilogram * meter / ampere / second ** 3.000
DecimalTuple(sign=0, digits=(5, 0), exponent=-1)
DecimalTuple(sign=0, digits=(5, 0, 0, 0), exponent=-6)
```

The conversion to base units, in some cases, introduces extra trailing zeros.
This can be confusing as it gives the impression there are suddenly more significant digits.
When using `pint` in an educational setting, this is not ideal.
(The issue cannot be reproduced with all units, e.g. no problem when repeating this example with `"5.0 mV"`.)

If was able to fix this by changing `1.0` to `1` on the following line:

https://github.com/hgrecco/pint/blob/c07fc9b218ea39965d8d4492e52168fe8f8a644b/pint/util.py#L568

I'm not sure if that could break other things?

(Note that this is just a minor snag. So far `pint` has been super useful! Thanks!!)

tovrstra:
Closes #1621

I've also modified conftest, to facilitate testing with customized unit registries. This is to be discussed, obviously.

- [x] Documented in docs/ as appropriate **<-- probably not needed for a minor bug fix?**
- [x] Discuss changes in `conftest.py`


noahsprent:
This worked to fix my issue in #1527

tovrstra:
@noahsprent Thanks for the feedback. I'll add a few more tests and make sure all tests pass.

tovrstra:
@jules-ch Would you be available for review? I believe the PR should be ready now. (I've added a test inspired by #1527.) Thanks!

hgrecco:
Thanks for the PR 👏

noahsprent:
Thanks very much @hgrecco!

tovrstra:
Also thanks from my side. Very much appreciated!

PR #1629, opened 2022-10-25
Fix cryptic error when re-registering a formatter

Yoshanuikabundi:
Hiya! I've been getting a very cryptic error with the [following code](https://github.com/openforcefield/openff-toolkit/pull/1442#issuecomment-1290535524):

```python
@pint.register_unit_format("simple")
def format_unit_simple(unit, registry, **options):
    return " * ".join(f"{u} ** {p}" for u, p in unit.items())
```

The error is:

```
openff/toolkit/utils/utils.py:118: in <module>
    def format_unit_simple(unit, registry, **options):
../../../micromamba-root/envs/test/lib/python3.8/site-packages/pint/formatting.py:158: in wrapper
    raise ValueError(f"format {name:!r} already exists")  # or warn instead
E   ValueError: Invalid format specifier
```

I believe this is because the colon shouldn't be there. `f"{1:!r}"` fails on Python 3.8, 3.9, and 3.10, whereas `f"{1!r}"` succeeds.

So here is the smallest ever PR!

### DIFF
--- a/pint/formatting.py
+++ b/pint/formatting.py
@@ -157,7 +157,7 @@ def register_unit_format(name):
 
     def wrapper(func):
         if name in _FORMATTERS:
-            raise ValueError(f"format {name:!r} already exists")  # or warn instead
+            raise ValueError(f"format {name!r} already exists")  # or warn instead
         _FORMATTERS[name] = func
 
     return wrapper
--- a/pint/testsuite/test_formatting.py
+++ b/pint/testsuite/test_formatting.py
@@ -65,5 +65,5 @@ def test_register_unit_format(func_registry):
     with pytest.raises(ValueError, match="format 'custom' already exists"):
 
         @fmt.register_unit_format("custom")
-        def format_custom(unit, registry, **options):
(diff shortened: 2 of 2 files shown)


jules-ch:
Thanks I think we can even get rid of the `!r` tranformer.
Can you add a test to cover this case ?

Yoshanuikabundi:
Test added!

I like the `!r`, I think it makes the message clearer - it's clearer that `'custom'` is a string than `custom`. But I'll take it out if you like.

jules-ch:
Can you fix the linter , you just need to add a comment `# noqa` in the line where you define the second formatter

keewis:
the names of the formatters are actually never used, so it would be totally fine to just rename the functions. For example, into `format_custom` and `format_custom_redefined` or `format_custom1` and `format_custom2`.

jules-ch:
@Yoshanuikabundi, I took the liberty to make the change to fix the CI on your behalf thank you for this fix ! :heart:

Yoshanuikabundi:
Thanks!! Sorry this slipped my mind!

PR #1633, opened 2022-10-25
Simplify registry subclassing

### LINKED ISSUE #1631: Regression: Units from derived unit registry no longer instance of pint.Unit
Hi,

when using Pint 0.20, a unit from a subclass of pint.UnitRegistry is no longer considered an instance of pint.Unit. Hard to explain, this code shows it:

```python
import pint

class MyUnitRegistry(pint.UnitRegistry):
    pass

print(isinstance(MyUnitRegistry().Unit("m"), pint.Unit))
```
With Pint 0.19.2, this prints `True`, with Pint 0.20, it prints `False`.

This leads to weird behaviour in e.g. pint_xarray when trying to check if something is a unit (done with `isinstance(pint.Unit)`) when the unit is derived from a subclass derived from pint.UnitRegistry. In my specific use case, I'm combining [pint_xarray](https://pint-xarray.readthedocs.io/en/stable/) and [openscm-units](https://openscm-units.readthedocs.io/en/latest/), and the combination breaks if Pint 0.20 is used.

I'm not quite sure if this can be solved / worked around downstream. Wanting to check if something is a unit via `isinstance` sounds reasonable, and subclassing also doesn't sound super wrong.

hgrecco:
- [x] Closes #1631

### DIFF
--- a/docs/user/custom-registry-class.rst
+++ b/docs/user/custom-registry-class.rst
@@ -21,7 +21,7 @@ and create your own.
 - ContextRegistry: the capability to contexts: predefined conversions
   between incompatible dimensions.
 - NonMultiplicativeRegistry: adds the capability to handle nonmultiplicative units (offset, logarithmic).
-- PlainRegistry: bas implementation for registry, units and quantities.
+- PlainRegistry: base implementation for registry, units and quantities.
 
 The only required one is `PlainRegistry`, the rest are completely
 optional.
--- a/docs/user/custom-registry-class.rst
+++ b/docs/user/custom-registry-class.rst
@@ -1,0 +1,83 @@
+.. _custom_registry_class:
+
+Custom registry class
+=====================
+
+Pay as you go
+-------------
+
+Pint registry functionality is divided into facets. The default
+UnitRegistry inherits from all of them, providing a full fledged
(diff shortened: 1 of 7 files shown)


hgrecco:
I have also added some docs in advanced / custom-registry-class

ping @jules-ch

docs/user/custom-registry-class.rst
```diff
@@ -0,0 +1,83 @@
...
+- FormattingRegistry: adds the capability to format quantities and units into string.
+- SystemRegistry: adds the capability to work with system of units.
+- GroupRegistry: adds the capability to group units.
+- MeasurementRegistry: adds the capability to handle measurements (quantities with uncertainties).
+- NumpyRegistry: adds the capability to interoperate with NumPy.
+- DaskRegistry: adds the capability to interoperate with Dask.
+- ContextRegistry: the capability to contexts: predefined conversions
+  between incompatible dimensions.
+- NonMultiplicativeRegistry: adds the capability to handle nonmultiplicative units (offset, logarithmic).
+- PlainRegistry: bas implementation for registry, units and quantities.
```
mikapfl, line 24:
```suggestion
- PlainRegistry: base implementation for registry, units and quantities.
```

PR #1639, opened 2022-10-27
fix grid docs urls

jules-ch:
Fix docs grid urls


hgrecco:
Thanks for the quick fix!

PR #1660, opened 2022-11-23
Fix unit check with `atol` using np.allclose.

### LINKED ISSUE #1658: numpy.allclose does not correctly handle the optional atol argument
While vanilla invocations of numpy.allclose work with pint, as do those supplying a dimensionless "rtol" parameter, invocations supplying "atol" fail with a DimensionalityError if the "atol" parameter is anything other than dimensionless, where as it should be expected have the dimensionality of the two primary arguments.

```python
import pint
import numpy as np

ureg = pint.UnitRegistry()
x = np.arange(10) * ureg.m
y = x.to(ureg.cm)
np.allclose(x, y, atol=1 * ureg.micron)
```

jules-ch:
- [x] Closes #1658


hgrecco:
🎉

keewis:
this unfortunately appears to have broken `isclose` / `allclose` with the default value for `atol`, which remain dimensionless / non-quantities. I guess we'd have to use sentinel values and pass through the defaults unchanged, if we decide to keep the change.

I'm undecided on whether or not to keep checking the units, though, as this is a numerical parameter that we could just as well ignore the unit of (I think it's only important to check that the `a` and `b` have consistent units).

@jthielen, any ideas / comments?

In any case, I think we need to resolve this one way or another before issuing a release.

jules-ch:
Is it possible to keep both behaviour, defaulting to unit of `a` ?
Idk if we shoudl

jules-ch:
I've just tested it, it works with default atol argument, I'm going to add some tests.

keewis:
oh, sorry, I totally misunderstood the implications of this change: the defaults work fine since they're not directly specified, but `xarray` has its own defaults that I can't really change, and *those* fail now with `pint`. As I said, I'd argue that these are numerical arguments that don't actually *need* a unit, but if it has one it should be consistent.

So my suggestion would be to allow either units compatible with the ones of `a` and `b`, or non-quantities/maybe `dimensionless`. If I understand the implementation correctly, `isclose` / `allclose` convert to the units of `a`, so we *could* fix this by attaching the units of `a` to `atol` (again, not sure if `dimensionless` should be an error or not).

dopplershift:
This has also caused failures on MetPy's test suite if we try to use Pint 0.21. For us, all we need is that if `atol` doesn't have units, they're assumed to match a. (We generally check proper units before handing off to other test functions.)

PR #1669, opened 2022-11-28
creation functions according to NEP 35 and updated `*_like` behavior

### LINKED ISSUE #882: Expected behavior of `np.*_like` functions

### LINKED ISSUE #1037: Implement np.array/np.asarray
Let's say I have a function like the following where doing the calculation involves doing a list comprehension which generates a list, and I want to turn it back into a numpy array before returning. The units of each element of the list comprehension will have the same units as some input `x`: if `x` is a normal Python number/numpy array, then I can turn the list into an array with a simple `np.array(lst)` or `np.asarray(lst)`. However, if `x` has units, this fails. My goal is to avoid introducing any pint-dependent code in `do_calculation` (e.g., let's say I have a big library of Python functions that calculate various physical formulae, part of the beauty of pint is if I pass in Quantities as inputs to these, the result is automatically carries units, but the library itself has no knowledge or dependency on pint; this list comprehension issue is the only place hangup).

Example:
```
import numpy as np
import pint
u = pint._DEFAULT_REGISTRY
def do_calculation(x, N):
    lst = [x * i for i in range(N)]
    ary = np.array(lst)
    ary2 = ary * 2
    return ary2
do_calculation(3, 10) # works
do_calculation(3 * u.meter, 10) # fails with ValueError: setting an array element with a sequence.
```

Is this something that will eventually be addressed by improvements in NEP18 support? Is there another best practice I should be following?

### LINKED ISSUE #1406: Handle array creation functions such as np.asanyarray
It's common for downstream packages to force inputs to numpy-likes using `np.asarray`, `np.asanyarray` etc. These have a `like` argument to handle duckarrays, it would be nice if pint handled these methods. Dask supports this if you want ideas how to implement it.

```python
import numpy as np 
import pint
import dask.array as da

a = np.array([1, 2, 3])
aa = pint.Quantity(a, "m")
np.asanyarray(a, like=aa)

  File "<ipython-input-7-59dc788bb8ad>", line 1, in <module>
    np.asanyarray(a, like=aa)

TypeError: no implementation found for 'numpy.asanyarray' on types that implement __array_function__: [<class 'pint.quantity.Quantity'>]

# dask handles this just fine:
np.asanyarray(a, like=da.array(a))
dask.array<array, shape=(3,), dtype=int32, chunksize=(3,), chunktype=numpy.ndarray>
```

keewis:
- [x] Closes #1406, closes #1037, closes #882

For a more consistent experience with the `*_like` functions and the functions from NEP 35 (introduced in `numpy=1.20`, minimum version is `1.19.5` but could be `1.20` according to NEP 29), this changes the semantics to:
- have `full_like` return the units of `fill_value` (a non-quantity `fill_value` results an array of the type of the magnitude):
```python
full_like(arr, fill_value=q) == Quantity(full_like(arr, fill_value=q.magnitude), q.units)
```
- have `empty_like` use the units of the argument:
```python
empty_like(q) == Quantity(empty_like(q.magnitude), q.units)
```
- have `ones_like` and `zero_like` a quantity with the units of the argument:
```python
ones_like(q) == full_like(q, fill_value=Quantity(1, q.units))
zeros_like(q) == full_like(q, fill_value=Quantity(0, q.units))
```
- support the NEP 35 keyword `like` for creation functions:
```python
np.array(arr, like=q) == Quantity(np.array(arr), q.units)
np.arange(10, like=q) == Quantity(np.arange(10, like=q.magnitude), q.units)
...
```

I'm not sure I got all of the functions that were changed by NEP 35, I couldn't find a comprehensive list. The ones I found are:
- `array`
- `asarray`
- `asanyarray`
- `arange`
- `ones`
- `zeros`
- `empty`
- `full`
- `identity`
- `eye`

but I'm almost certain I missed some. The definition of all of them except `full` is pretty simple: just use the units of the `like` argument. For `full`, I choose to go with the unit of the `fill_value` or `dimensionless` if the fill value happens not to be a quantity (because by passing `like` we expect a quantity). I'm not really sure about that, though.

cc @jthielen

### DIFF
--- a/pint/facets/numpy/numpy_func.py
+++ b/pint/facets/numpy/numpy_func.py
@@ -950,9 +950,11 @@ def implement_nep35_func(func_str):
 nep35_function_names = {
     "array",
     "asarray",
+    "asanyarray",
     "arange",
     "linspace",
     "identity",
+    "eye",
 }
 for func_str in nep35_function_names:
     implement_nep35_func(func_str)
--- a/pint/facets/numpy/numpy_func.py
+++ b/pint/facets/numpy/numpy_func.py
@@ -934,6 +934,30 @@ for func_str in ["var", "nanvar"]:
     implement_func("function", func_str, input_units=None, output_unit="variance")
 
 
+def implement_nep35_func(func_str):
+    # If NumPy is not available, do not attempt implement that which does not exist
+    if np is None:
+        return
+
+    func = getattr(np, func_str)
+
+    @implements(func_str, "function")
+    def implementation(*args, like, **kwargs):
+        result = func(*args, **kwargs)
+        return like._REGISTRY.Quantity(result, like.units)
+
+
+nep35_function_names = {
+    "array",
+    "asarray",
+    "arange",
+    "linspace",
+    "identity",
+}
+for func_str in nep35_function_names:
+    implement_nep35_func(func_str)
+
+
 def numpy_wrap(func_type, func, args, kwargs, types):
     """Return the result from a NumPy function/ufunc as wrapped by Pint."""
 
--- a/pint/facets/numpy/numpy_func.py
+++ b/pint/facets/numpy/numpy_func.py
@@ -530,18 +530,19 @@ def _full_like(a, fill_value, dtype=None, order="K", subok=True, shape=None):
     # Make full_like by multiplying with array from ones_like in a
     # non-multiplicative-unit-safe way
     if hasattr(fill_value, "_REGISTRY"):
-        return fill_value._REGISTRY.Quantity(
-            (
-                np.ones_like(a, dtype=dtype, order=order, subok=subok, shape=shape)
-                * fill_value.m
-            ),
-            fill_value.units,
-        )
+        units = fill_value.units
+        fill_value_ = fill_value.m
     else:
-        return (
-            np.ones_like(a, dtype=dtype, order=order, subok=subok, shape=shape)
-            * fill_value
-        )
+        units = None
+        fill_value_ = fill_value
+
+    magnitude = np.full_like(
+        a.m, dtype=dtype, order=order, subok=subok, shape=shape, fill_value=fill_value_
+    )
+    if units is not None:
+        return fill_value._REGISTRY.Quantity(magnitude, units)
+    else:
+        return magnitude
 
 
 @implements("interp", "function")
--- a/pint/facets/numpy/numpy_func.py
+++ b/pint/facets/numpy/numpy_func.py
@@ -964,6 +964,27 @@ for func_str in nep35_function_names:
     implement_nep35_func(func_str)
 
 
+@implements("full", "function")
+def _full(shape, fill_value, dtype=None, order="C", *, like):
+    # Make full_like by multiplying with array from ones_like in a
+    # non-multiplicative-unit-safe way
+    if hasattr(fill_value, "_REGISTRY"):
+        units = fill_value.units
+        fill_value_ = fill_value.m
+    else:
+        units = None
+        fill_value_ = fill_value
+
+    magnitude = np.full(shape=shape, fill_value=fill_value_, dtype=dtype, order=order)
+    if units is not None:
+        return fill_value._REGISTRY.Quantity(magnitude, units)
+    else:
+        return like._REGISTRY.Quantity(magnitude, units)
+
+
+nep35_function_names.add("full")
+
+
 def numpy_wrap(func_type, func, args, kwargs, types):
     """Return the result from a NumPy function/ufunc as wrapped by Pint."""
 
--- a/pint/facets/numpy/numpy_func.py
+++ b/pint/facets/numpy/numpy_func.py
@@ -903,9 +903,6 @@ for func_str in [
     "isreal",
     "iscomplex",
     "shape",
-    "ones_like",
-    "zeros_like",
-    "empty_like",
     "argsort",
     "argmin",
     "argmax",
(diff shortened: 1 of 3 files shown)


jthielen:
The changes here look good to me, and as far as I'm aware, that's the same list of functions that I'm aware of that'd be covered by NEP 35. So I'd say go with it, and if something a user needs happens to have been missed, it could be added later?

Speaking of adding later, though, would you be able to add code comments (and modify the existing ones that begin with `# Make full_like by multiplying`) clarifying the approaches taken here, so that it is easier for future contributors to these sections of code? For example, the code pattern in https://github.com/hgrecco/pint/pull/1669/files#diff-3902d8460bf42f38121b360a0cf397597567077c08aa6d0a565a65332290395cR953-R967 seems unnatural on first glance, and so could use comments to clarify that first you're doing the "automatic" implementations, then the custom ones. Also, while it is still an unchecked checklist item (at time of writing comment), I think these nuances of array creation could benefit from some direct documentation in the NumPy page of the docs.

jules-ch:
Seems like some tests with xarray are making CI fail.
Do we know what's triggering this ?

keewis:
thanks, @jthielen, those are some good points. I've tried to address them in the most recent commits, though I suspect to be really useful we'd need to rewrite / restructure the `numpy` guide: right now, the implementation details are condensed into a single paragraph, which doesn't make it easy to search for a specific type of function, and we're also missing clear examples. Instead, I suggest having one section per function category. If nobody objects, I'll try to do that in a new PR (not sure when I get to that, though).

@jules-ch: no idea. Might be because of changes in this PR, I'll have to investigate.

keewis:
> Seems like some tests with xarray are making CI fail. Do we know what's triggering this ?

It seems `xarray` is using `zeros_like` in `isnull` to construct a all-`False` array for dtypes that are not nullable, and because we changed `zeros_like` to return a quantity this has stopped working. To be fair, this won't require 

I'd argue that this is implicit behavior in `xarray` (`False` just so happens to be represented as `0` in `numpy`), and that it would be much better to use `full_like` in this case:
```python
zeros_like(data, dtype=bool)  # old
full_like(data, dtype=bool, fill_value=False)  # new
```

However, even though we can fix this `xarray` very quickly we'd basically have to require the newest version of `xarray` for use with a new release of `pint`. So I'm not sure how to best handle this: should we update `xarray`, and wait 6 months before merging this PR? Or is it fine to just merge and require people to update `xarray`?

keewis:
another option would be to avoid wrapping `bool` dtypes, but that might make the code much more complicated

keewis:
this has become tricky... see pydata/xarray#6752 for more details.

The most recent commit fixes a bug in `full_like` when called on a `pint` quantity wrapping a `dask` array (`dask` chose not to implement `subok`), which prevents `xarray` from merging my PR and releasing a new version compatible with the new behavior of `*_like`. However, we need that version to be able to change the behavior of `*_like` without breaking `xarray`'s implementation of `isnull` on non-nullable dtypes (all except `object`, `float`, and `complex`).

To get out of this dead-lock, I think we need to somehow introduce a way to switch between old and new behavior of `*_like` (a new registry option?). This would allow us to properly deprecate the old behavior: the next version would ship with deprecation warnings for the old behavior of `*_like`, the version after that would switch the behavior but allow switching back to the old behavior, and some versions later we'd remove the switch and the old code. In general, though, I think we can keep the deprecation cycle short since we warned about the behavior of `*_like` being experimental.

(and of course, if someone has a better idea I'd be open to that as well)

andrewgsavage:
>t he next version would ship with deprecation warnings for the old behavior of *_like

Do you want to add a depreciation warning in another PR?

PR #1670, opened 2022-11-28
use `packaging.version` instead of the deprecated `distutils.version`

keewis:
- [x] Executed ``pre-commit run --all-files`` with no errors
- [ ] The change is fully covered by automated unit tests
- [ ] Documented in docs/ as appropriate
- [ ] Added an entry to the CHANGES file

PR #1671, opened 2022-11-29
explicitly install `packaging` for testing

keewis:
Not really necessary since some of the dependencies pull it in, anyways, but it's good to explicitly install it to avoid breaking CI in case that changes.

- [x] follow-up to #1670


jules-ch:
Yes we add some issues regarding packaging when it was necessary to have it when we had features only available for certain versions of numpy.

It should be available for CI, but better safe than sorry

jules-ch (approved):

PR #1677, opened 2022-12-11
Array ufunc multiplication

### LINKED ISSUE #1674: Operations between arrays of quantity scalars and quantity holding array resulting in incorrect units
When implementing array_ufunc in pint-pandas I've hit some odd results. 
```python3
import pint
ureg =pint.UnitRegistry()

Q_ = ureg.Quantity
q_arr = Q_(np.array([1,2]),"m")
arr_of_q = np.array([Q_(1,"m"), Q_(2,"m")], dtype="object")
C:\Users\A\AppData\Local\Temp\ipykernel_16288\106945721.py:3: UnitStrippedWarning: The unit of the quantity is stripped when downcasting to ndarray.

print(arr_of_q)
[<Quantity(1, 'meter')> <Quantity(2, 'meter')>]


print(q_arr*arr_of_q)
[<Quantity(1, 'meter')> <Quantity(4, 'meter')>] meter # odd to have meters inside and outside array


print(arr_of_q*q_arr)
[1 4] meter # Should be m^2

```

andrewgsavage:
- [x] Closes #1674 


pint/testsuite/test_issues.py
```diff
@@ -886,6 +886,15 @@ def test_issue_1300(self):
...
         assert m.default_format == "~P"
 
+    @helpers.requires_numpy()
+    def test_issue1674(self, module_registry):
+        Q_ = module_registry.Quantity
+        arr_of_q = np.array([Q_(2, "m"), Q_(4, "m")], dtype="object")
+        q_arr = Q_(np.array([1, 2]), "m")
+
+        helpers.assert_quantity_almost_equal(arr_of_q * q_arr, Q_([2, 8], "m^2"))
+        helpers.assert_quantity_almost_equal(arr_of_q / q_arr, Q_([2, 2], ""))
```
burnpanck, line 896: This doesn't look quite right to me: You start from an `dtype=object` array of quantity scalars, where each element could potentially be of a different unit. The result is now a "normal" quantity array, where all elements share a single unit.

Please add a test case for `arr_of_q = np.array([Q_(2, "m"), Q_(4, "s")], dtype="object")`.

Obviously, this will not work with a quantity array, and the expected result of `arr_of_q * q_arr` would be `np.array([Q_(2, "m^2"), Q_(4, "m s")], dtype="object")`. For consistency, I would have expected to receive a `dtype=object` array in all cases.

burnpanck (changes requested):

hgrecco:
@andrewgsavage can you comment on @burnpanck objection?

andrewgsavage:
yea hold merging off till it's addressed

andrewgsavage:
numpy tests are working on numpy 1.23 but failing on 1.25

I was getting this message for the failing test on 1.23
  F:\A\repos\pint\pint\compat.py:306: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
    out = lhs == rhs

andrewgsavage:
this is ready for review

andrewgsavage:
Can this be reviewed? I'd like to make pint-pandas changes that rely on this.

pint/compat.py
```diff
@@ -303,7 +303,10 @@ def eq(lhs: Any, rhs: Any, check_all: bool) -> Union[bool, Iterable[bool]]:
     -------
     bool or array_like of bool
     """
-    out = lhs == rhs
+    try:
+        out = lhs == rhs
+    except ValueError:
+        return False
```
burnpanck, line 309: What is the reason for this change? Two things being incomparable is not necessarily the same as not being equal. This is a behavioural change that I don't see how it is related to the bug to be fixed (except maybe to simplify the test code?). This may potentially silently cause wrong behaviour. That said, the contract for `eq` isn't really clear and I don't know how/where it is used. It is well possible that the behaviour before the change was wrong.

burnpanck (changes requested):
Sorry, I didn't realise that I was listed as a reviewer (probably added automatically by my previous comment). I feel honoured that you are interested in my review, but I'm by no means a maintainer of pint or any related project, so I'm no authority here. I'm just an occasional contributor who likes failures to be loud instead of silent (and thus a big advocate of units libraries in any language).

pint/facets/numpy/numpy_func.py
```diff
@@ -284,6 +284,18 @@ def implement_func(func_type, func_str, input_units=None, output_unit=None):
 
     @implements(func_str, func_type)
     def implementation(*args, **kwargs):
+        if func_str in ["multiply", "true_divide", "divide", "floor_divide"] and any(
+            [
+                _is_sequence_with_quantity_elements(arg) and not _is_quantity(arg)
```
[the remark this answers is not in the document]
andrewgsavage: done

pint/compat.py
andrewgsavage: my mistake, this was to get something working in pint-pandas. reverted this

andrewgsavage:
this is ready for another look

burnpanck (approved):
LGTM

pint/facets/numpy/numpy_func.py
```diff
@@ -284,6 +284,18 @@ def implement_func(func_type, func_str, input_units=None, output_unit=None):
...
+            [
+                not _is_quantity(arg) and _is_sequence_with_quantity_elements(arg)
+                for arg in args
+            ]
+        ):
+            # the sequence may contain different units, so fall back to element-wise
+            return np.array(
+                [func(args[0][i], args[1][i]) for i in range(len(args[0]))],
+                dtype=object,
+            )
```
keewis, line 297: I wonder if it would be better to use `zip` instead? We're loading everything into memory, anyways:
```suggestion
            return np.array([func(*func_args) for func_args in zip(*args)], dtype=object)
```

PR #1685, opened 2022-12-19
run `black` on notebooks as well

keewis:
A while ago, `black` introduced the `black-jupyter` hook, which blackens code in notebook cells.

Being consistent within a project is important, so in a new PR we should try to blacken the code in the narrative documentation / doctests as well.


hgrecco:
👍

Issue #1679, opened 2022-12-14
Pint typing (again?)


jldiaz:
According to this comment, https://github.com/hgrecco/pint/pull/1259#issuecomment-848086093 typing integration is done, and working with mypy. This is not my experience. I'm using `pint 0.20.1` and `mypy 0.991` and any `Quantity` is inferred as type `Any` by mypy.

I tried this example:

```
from pint import UnitRegistry

ureg = UnitRegistry()

float_quantity = ureg.Quantity(150.0, "m**2")
int_quantity = ureg.Quantity(150, "m**2")
list_quantity = ureg.Quantity((1, 2, 3, 4), "m**2")
str_quantity = ureg.Quantity("50*2 m")
sequence = [ureg.Quantity(2, "m"), ureg.Quantity(3, "m")]

reveal_type(float_quantity)
reveal_type(int_quantity)
reveal_type(float_quantity.m)
reveal_type(float_quantity.to("km**2"))
reveal_type(str_quantity)
reveal_type(list_quantity)
```

And ran it through `mypy`, but the output was puzzling:

```
typetest.py:11: note: Revealed type is "Any"
typetest.py:12: note: Revealed type is "Any"
typetest.py:13: note: Revealed type is "Any"
typetest.py:14: note: Revealed type is "Any"
typetest.py:15: note: Revealed type is "Any"
typetest.py:16: note: Revealed type is "Any"
Success: no issues found in 1 source file
```

Why is this? Something was changed in pint or in mypy? Did I miss something?

**Note** Also Pylance (Microsoft's python type checker powered by Pyright, used by VSCode) infers the type `Any` for all these variables. Pylance version is `v2022.12.20`.

austinorr:
@jldiaz Hi there, I just had this same issue and dove in to the type checker code. This PR ( #1686 ) fixed this regression for me, can you give it a try and let me know if it would close this issue for you as well? I'd rather not open extra issues here if I can avoid it!

jldiaz:
@austinorr Thank you for your answer.

Unfortunately  PR #1686 does not solve the problem of `Quantity` being considered of type `Any`, and in addition it introduces new problems at every line which uses `ureg.Quantity()` to build a quantity. Mypy output is now:

```
typetest.py:5: error: "PlainQuantity[Any]" not callable  [operator]
typetest.py:6: error: "PlainQuantity[Any]" not callable  [operator]
typetest.py:7: error: "PlainQuantity[Any]" not callable  [operator]
[...]

Issue #1662, opened 2022-11-24
Release 0.20.x breaks generics on Quantity


gerbenoostra:
Since 0.18, and up to 0.19.*, one could use generics to ensure the magnitude is the right type, and could also including numpy's typing mechanism.

For example:
```
from pint import Quantity
import numpy as np
import numpy.typing as npt

def method(measure1: Quantity[int], measure2: Quantity[npt.NDArray[np.float_]]):
    pass
```
But with 0.20.1 one gets a `TypeError: <class 'pint.util.Quantity'> is not a generic class`

In 0.20.1 I see the following facets:
```
from pint.facets.plain.quantity import PlainQuantity
from pint.facets.numpy.quantity import NumpyQuantity
```
But only `PlainQuantity` is generic, so I don't know how to migrate the `numpy` quantities while preserving the `numpy` arrray type annotations.

Probably related to https://github.com/hgrecco/pint/issues/1641

jules-ch:
I think inheriting NumpyQuantities from PlainQuantity will fix this issue

gerbenoostra:
How is the type hierarchy designed? Should `Quantity` be a descendent of every facet, merging their functionality. Or should each facet inherit from a base `Quantity` that provides the common interface.
Either way, providing a generic version for each to ensure the right measurement type would help in type safety.

hgrecco:
The default `Quantity` inherits from every facet, but is constructed dynamically according to the registries. Maybe we should provide a more static typing friendly version.

hgrecco:
Feel free to reopen if needed

Issue #1656, opened 2022-11-22
Quantity module import not found in 0.20


James-Wickenden:
Hi,

In Pint 0.20 and pint-pandas 0.3 the pint.quantity module has been removed, causing our tests to fail.
I'm working around it for the moment by importing pint==0.19.2.

Is there an intended alternative in the pint API going forward? I understand this issue is similar in scope to #1641, so apologies if I'm asking something thats already been asked.

Thanks!

hgrecco:
Feel free to reopen if another problem is found

Issue #1641, opened 2022-10-31
Breaking change not mentioned in changes, is there an easy way to support both Pint0.19 and 0.20?


MauriceHendrix:
The latest release  (0.20) caused a breaking change Is there an easy way for our project to support both pint 0.19 and 0.20?
It seems to do with a few things not mentioned in the change log. 

Errors we have seen:
`from pint.converters import ScaleConverter` class not found, it seems this was relocated to `pint.facets.plain.definitions.ScaleConverter`?
`from pint.definitions import UnitDefinition` class not found, it seems this was relocated to `from pint.facets.plain.definitions.UnitDefinition`?

'''
definition = UnitDefinition(qname, '', (), ScaleConverter(quantity.to(self._registry.dimensionless)))
E           TypeError: __init__() missing 1 required positional argument: 'reference'
'''
Reference used to be optional, before I beleive?

hgrecco:
The API to programmatically define units, dimensions, etc has not been really stabilized yet. But now that the parser has been split, I think we are ready to do this.

My idea would be:
1. import all relevant things into a module or modules (i.e. converters and definitions, or just definitions)
2. point people to call `ureg.define(obj)` where obj is an instance of any of the definition classes.

For the time being you can just do:
 definition = UnitDefinition(qname, '', (), ScaleConverter(quantity.to(self._registry.dimensionless)))

MauriceHendrix:
@hgrecco yes that's what I figured, but what imports would I use? Is there a correct set of imports that will work for both pint 0.19 and 0.20?

hgrecco:
I was sure that the facets change was going to bring some trouble, that is why we stalled the release. But it was worth it as we can now hack the code much easier. How about trying to release a 0.20.2 that brings this back?

MauriceHendrix:
It would be really nice tbh not to have to do the following if possible
```
try:
    from pint.converters import ScaleConverter
except ImportError:
    from pint.facets.plain.definitions import ScaleConverter
```

I would just move to requiring Ping 0.20+ but unfortunately the powers that be require us to still support Python 3.6 & 3.7 
https://chaste.cs.ox.ac.uk/trac/wiki/InstallGuides/DependencyVersions

jules-ch:
@hgrecco do you want to bring back definitions & converters back with all of the dataclasses with it to fix this issue ?

hgrecco:
Feel free to reopen if another problem is found

PR #1686, opened 2022-12-20
patch imports during type check

austinorr:
During the 0.20x update several TYPE_CHECKING related imports were moved and renamed during the facets refactoring. For example, here in the [ `pint._typing` ](https://github.com/hgrecco/pint/blob/521dc80fabdacae20d57d9e3dce95e04633894ca/pint/_typing.py#L5) module:
```
if TYPE_CHECKING:
    from .facets.plain import Quantity, Unit, UnitsContainer
```
`facets.plain` does not have any of these objects after they were renamed to `PlainUnit` etc. This causes downstream project type checkers to fail with many strange errors.  This PR attempts to fix all of the places that TYPE_CHECKING imports are misdirected. To find them I used mypy and pylance in VSCode.

Potentially related issues: #1656, #1662, #1641, #1679


jules-ch:
I'll try to review that soon

hgrecco:
Thanks for the fix

PR #1721, opened 2023-03-07
add NEP18 overrides for `max`, `min`, and `round`

keewis:
`numpy` recently changed the names of `amin` and `amax` to `min` and `max`, with the old names being kept around as aliases. However, that means that we need to override *both* until `numpy<1.25` drops out of support (which will be a while, considering that `numpy=1.25` has not been released yet).

Additionally, `round_` has been deprecated in favor of `round`.

The test coverage has already been sufficient (we didn't get errors in CI because we don't test `numpy` dev), but I changed the `round` tests to not use the deprecated name.

- [x] ~The change is fully covered by automated unit tests~
- [x] ~Documented in docs/ as appropriate~


jules-ch:
Thanks

PR #1722, opened 2023-03-08
Fix __dask_postcompute__() to better preserve type

dopplershift:
In Unidata/MetPy#2945, a call to dask's .compute() was causing the resulting type to be a different Quantity() variant (from pint.util rather than the parent registry), which resulted in isinstance() failing. This changes things to use the appropriate type from `self` rather than hard-coded class names.


jules-ch:
Thanks @dopplershift

PR #1727, opened 2023-03-20
Update error URLs for log / offset units.

cgevans:
Log / offset unit documentation changed location, making the URLs shown in  error message no longer valid.  In addition, the two are on separate pages in the most recent documentation.  This commit updates the URLs. It also, as it seems most likely to be stably useful for users, points to the latest stable documentation, rather than the latest dev version.

I didn't bother to make an issue for this, because it is rather minimal.

- [N/A] Closes # (insert issue number)
- [N/A] The change is fully covered by automated unit tests
- [N/A] Documented in docs/ as appropriate


hgrecco:
Thanks 🎉

PR #1729, opened 2023-03-20
Parse '°' along with previous text, not adding a preceding space

### LINKED ISSUE #1728: Δ°C is not recognized as a unit, despite being the defined symbol for delta_degree_Celsius.
Tested with 0.20.1 and 0.20.2.dev37+g5314065.

```
import pint

ureg = pint.UnitRegistry()

ureg("1 delta_degree_Celsius") # name, works

ureg("1 ΔdegC") # alias, works

ureg("1 Δ°C") # defined_symbol, does not work.

ureg.Quantity("1", "Δ°C") # also doesn't work; see traceback below.
```

This means the unit won't be recognized from its own output, eg, `ureg(f"{ureg('1 delta_degree_C'):~}")` will not work.

The same error does not occur for °C itself.  It appears that the unit parser is forcing  split between Δ and °, and then trying to interpret Δ as its own unit.  It does not do this for ΔdegC, however.

```
UndefinedUnitError                        Traceback (most recent call last)
Cell In[10], line 1
----> 1 ur.Quantity(1, 'Δ°C')

File ~/devel/default-venv/lib64/python3.11/site-packages/pint/facets/plain/quantity.py:222, in PlainQuantity.__new__(cls, value, units)
    220     units = units
    221 elif isinstance(units, str):
--> 222     units = inst._REGISTRY.parse_units(units)._units
    223 elif isinstance(units, SharedRegistryObject):
    224     if isinstance(units, PlainQuantity) and units.magnitude != 1:

File ~/devel/default-venv/lib64/python3.11/site-packages/pint/facets/plain/registry.py:1071, in PlainRegistry.parse_units(self, input_string, as_delta, case_sensitive)
   1069 for p in self.preprocessors:
   1070     input_string = p(input_string)
-> 1071 units = self._parse_units(input_string, as_delta, case_sensitive)
   1072 return self.Unit(units)

File ~/devel/default-venv/lib64/python3.11/site-packages/pint/facets/nonmultiplicative/registry.py:66, in NonMultiplicativeRegistry._parse_units(self, input_string, as_delta, case_sensitive)
     63 if as_delta is None:
     64     as_delta = self.default_as_delta
---> 66 return super()._parse_units(input_string, as_delta, case_sensitive)

File ~/devel/default-venv/lib64/python3.11/site-packages/pint/facets/plain/registry.py:1104, in PlainRegistry._parse_units(self, input_string, as_delta, case_sensitive)
   1102 many = len(units) > 1
   1103 for name in units:
-> 1104     cname = self.get_name(name, case_sensitive=case_sensitive)
   1105     value = units[name]
   1106     if not cname:

File ~/devel/default-venv/lib64/python3.11/site-packages/pint/facets/plain/registry.py:590, in PlainRegistry.get_name(self, name_or_alias, case_sensitive)
    588 candidates = self.parse_unit_name(name_or_alias, case_sensitive)
    589 if not candidates:
--> 590     raise UndefinedUnitError(name_or_alias)
    591 elif len(candidates) == 1:
    592     prefix, unit_name, _ = candidates[0]

UndefinedUnitError: 'Δ' is not defined in the unit registry
```

cgevans:
Currently, as tokenize.tokenize does not parse ° as a NAME, it is translated to " degree".  However, the preceding space means that units with a ° in the middle, for example, Δ°C, cannot be parsed.  Removing the preceding space fixes this, and does not appear to break any tests.

I have not added tests for this yet, or documentation.  I am creating this draft PR to check tests here, as there are test failures with 3.11.

- [x] Closes #1728


cgevans:
I've added some tests for various arrangements of ° (coverage won't seem to change, because from a line-by-line perspective, this line is already used in other tests).

The primary problem I can see with this change is that something like "mol°K" does not work, though "mol °K" does.  This is not currently tested for, and I'm not sure if the first was intended behaviour.

cgevans:
It looks like the PR got updated to the current master, but then closed without merging; was this intentional?

hgrecco:
No! Let me check

PR #1741, opened 2023-04-14
use `**kwargs` to pass along additional arguments to `ones_like`

keewis:
This is a fix to be included in a release before merging #1669, as the changes there would break `xarray` with no possible work-around. Not sure if we need tests, as those would probably be tricky to write and maintain.

@hgrecco, @jules-ch, I'd like to have a bugfix release some time soon, and include #1669 in the next release. Would that be possible? (Also, I'd be happy to help make that happen)


hgrecco:
🎉

PR #1747, opened 2023-04-19
Use already existing cache if found during build

### LINKED ISSUE #1746: Dimensional equivalents missing when loading from cache
When instantiating a `UnitRegistry` from cached files, the dimensional equivalents are instantiated as an empty dictionary. This results in all calls to `get_compatible_units` to fail with a key error.

```python
import pint
ureg = pint.UnitRegistry(cache_folder=tmp_path)
t = ureg.get_compatible_units("L")
assert len(t) > 0

ureg = pint.UnitRegistry(cache_folder=tmp_path)
t = ureg.get_compatible_units("L")  # raises an error
```

The error in this particular stack trace is

```
>       t = ureg.get_compatible_units('L')

pint/testsuite/test_diskcache.py:106:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pint/facets/plain/registry.py:840: in get_compatible_units
    equiv = self._get_compatible_units(input_units, group_or_system)
pint/facets/system/registry.py:237: in _get_compatible_units
    return frozenset(members & super()._get_compatible_units(input_units, None))
pint/facets/group/registry.py:123: in _get_compatible_units
    ret = super()._get_compatible_units(input_units, group)
pint/facets/context/registry.py:398: in _get_compatible_units
    ret = super()._get_compatible_units(input_units, group_or_system)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <pint.registry.UnitRegistry object at 0x102fa6740>
input_units = <ParserHelper(1, {'L': 1})>, group_or_system = None

    def _get_compatible_units(self, input_units, group_or_system):
        """ """
        if not input_units:
            return frozenset()

        src_dim = self._get_dimensionality(input_units)
>       return self._cache.dimensional_equivalents[src_dim]
E       KeyError: <UnitsContainer({'[length]': 3})>

pint/facets/plain/registry.py:850: KeyError
```

camuthig:
- [x] Closes #1746

This pull request attempts to update the `PlainRegistry` to set `self._cache` to the data retrieve from the cached data. Tests are passing, so I believe this change resolves the issue linked, but I'm not confident if it is the correct resolution or not.

### DIFF
--- a/pint/facets/plain/registry.py
+++ b/pint/facets/plain/registry.py
@@ -534,7 +534,9 @@ class PlainRegistry(metaclass=RegistryMeta):
             if cache is None:
                 self._build_cache()
                 diskcache.save(self._cache, loaded_files, "build_cache")
-            return
+            else:
+                self._cache = cache
+                return
 
         self._cache = RegistryCache()
 
--- a/pint/testsuite/test_diskcache.py
+++ b/pint/testsuite/test_diskcache.py
@@ -95,3 +95,13 @@ def test_change_file(tmp_path):
     assert ureg.x == 1235
     files = tuple(ureg._diskcache.cache_folder.glob("*.pickle"))
     assert len(files) == 4
+
+
+def test_dimensional_equivalent_are_cached(tmp_path):
+    ureg = pint.UnitRegistry(cache_folder=tmp_path)
+    t = ureg.get_compatible_units("L")
+    assert len(t) > 0
+
+    ureg = pint.UnitRegistry(cache_folder=tmp_path)
+    t = ureg.get_compatible_units("L")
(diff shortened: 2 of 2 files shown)


hgrecco:
I don't think this is a good idea without further thoughts. Basically, I am worried that we might be caching objects which are specific to this registry. Remember that certain objects (Quantity/Unit/Context/Measurement) are registry instance specific and should not be cached. That part of the code is not in use to investigate if this was a good idea or not.

Issue #1128, opened 2020-07-01
Shall we create a QuantityArray class?


hgrecco:
I have just released `Pint-Pandas 0.1` So first, let me publicly thank @andrewgsavage, @znicholls and @sigvaldm for the awesome work you have done to code, discuss and push this package. Congratulations!

I have also released `Pint 0.14` to support a few last minute changes in `Pint-Pandas`.

`Pint-Pandas` has exposed (one more time) an old discussion: Do we need a `QuantityScalar` and `QuantityArray` classes? This was something on the table at the beginning of Pint. The conclusion was to avoid it for simplicity and we found a very nice way to do it without sacrifizing functionality. From very early releases we were able to support a large number of NumPy functions without requiering numpy and to have a nice way to write array quantities (`[1, 2, 3] * ureg.meter`). We also made the choice to support but doe not require numpy. I still think this was the right call and we should try to stay on that path.

Pint-Pandas has exposed a some cases in which having a `QuantityArray` seems to be unavoidable to fully support Pandas functionality with Pint. The question is ... shall we do it? If yes, how to do it in a way which is transparent to the normal user?

PS.- Thanks again to the people behind Pint-Pandas.

znicholls:
I read through #875, #764 and #753 yesterday and found these super helpful background resources. Unfortunately they made things seem very complicated and I couldn't see an obvious way forward. 

The big conflict I saw was that, on the one hand, numpy's separation of scalar and array classes is seen as a ['design mistake'](https://github.com/hgrecco/pint/issues/753#issuecomment-536240372), whereas on the other hand, not having the separation is causing headaches with Pandas (particularly with [how Pandas decides whether something is iterable or not](https://github.com/hgrecco/pint-pandas/issues/33#issuecomment-647225205)). At risk of poking the hornet's nest, perhaps the simplest solution is to enquire in Pandas about whether https://github.com/pandas-dev/pandas/blob/4743cc91719146167eefb43865413fc0b0aef1d4/pandas/_libs/lib.pyx#L992 could be updated so that it checks `ndim` of any classes which have such an attribute, rather than only checking `ndim` of numpy arrays? (Only checking if the object is actually a numpy array seems an odd choice to me which restricts the ability of Pandas to be extended...)

hgrecco:
@znicholls Those discussions are indeed useful. I am pinging the people that were involved. @jthielen @crusaderky @shoyer @keewis

hgrecco:
[...]

PR #1778, opened 2023-05-13
Remove __array__ interface

coroa:
- [ ] Closes https://github.com/hgrecco/pint-pandas/issues/174
- [ ] ~Documented in docs/ as appropriate~

In hunting down options to address https://github.com/hgrecco/pint-pandas/issues/174 , the only solution i found so far to reliably work was to get rid of the `__array__` interface entirely, as a more extreme version of the discussion in https://github.com/hgrecco/pint/issues/1128 .

I was quite surprised to learn that nothing in pint directly depends on the interface, but i guess it would be too much of a breaking change. Nevertheless, I like the idea of pint not ever implictly stripping units.

More realistically this means a proper implementation of the `__array_ufunc__` alternative might solve at least the linked issue.

I guess you all are exhausted by these discussion already, sorry that i am pbb missing all the basics.

### DIFF
--- a/pint/facets/numpy/quantity.py
+++ b/pint/facets/numpy/quantity.py
@@ -129,10 +129,10 @@ class NumpyQuantity(Generic[MagnitudeT], PlainQuantity[MagnitudeT]):
         if isinstance(values, self.__class__):
             values = values.to(self).magnitude
         elif self.dimensionless:
-            values = self.__class__(values, "").to(self)
+            values = self.__class__(values, "").to(self).magnitude
         else:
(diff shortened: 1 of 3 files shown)


andrewgsavage:
This is interesting, the main issues we've had previously was due to a pint scalar Q(1,"m") being treated as iterable since it has a __iter__ method, which pandas looked for to determine whether an object was array like. The __iter__ would fail, but pandas was checking whether it exsists not whether it fails.

This is interesting as you've removed __array__ not __iter__.

andrewgsavage:
can close this since reductions now work for pint-pandas

PR #1783, opened 2023-05-17
allow non-quantity `atol` parameters for `isclose` and `allclose`

keewis:
As suggested in https://github.com/hgrecco/pint/pull/1660#issuecomment-1536572859, this uses the unit of `a` if `atol` is a non-quantity (i.e. a non-`pint` type).

@dopplershift, this fixes the failing tests for `xarray` so I *think* it should also fix the failures for `metpy`, but it might be good to verify that before merging.

- [x] follow-up to #1660

### DIFF
--- a/pint/facets/numpy/numpy_func.py
+++ b/pint/facets/numpy/numpy_func.py
@@ -869,7 +869,6 @@ for func_str, unit_arguments, wrap_output in (
     ("max", ["a", "initial"], True),
     ("min", ["a", "initial"], True),
     ("searchsorted", ["a", "v"], False),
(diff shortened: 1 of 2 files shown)


dopplershift:
I can confirm this fixes running MetPy's test suite with Pint > 0.20.

hgrecco:
Great. My proposal would be to merge this on top of 0.21 and, if all tests run ok, release 0.21.1

Also, I would merge it on top of 0.22rc1 and release 0.22rc2

hgrecco:
ping @jules-ch

PR #1805, opened 2023-06-13
Fix Transformation type protocol #1801

### LINKED ISSUE #1801: Type hinting: `Transformation` type incompatible with usage in `add_transformation`
Thanks for adding type annotations to `pint`!

However, there seems to be a small problem with the `Transformation` protocol. We are encountering errors with the `add_tranformation` method. 

Here's an example based on the docs: 

pint version: 0.22 
mypy version: 0.991

```python
import pint

ureg = pint.UnitRegistry()
c = pint.Context("ab")

c.add_transformation("[length]", "[time]", lambda ureg, x: x / ureg.speed_of_light)
```

running `mypy` on this script produces: 

```
pint_type_demo.py:6: error: Argument 3 to "add_transformation" of "Context" has incompatible type "Callable[[Any, Any], Any]"; expected "Transformation"  [arg-type]
```

This makes sense, because the [Transformation](https://github.com/hgrecco/pint/blob/3fb63af49df803e7bcad1fe9e6bdaab04021f4cc/pint/facets/context/objects.py#L22) protocol definition does not include the ureg, just the value. 

I think it should be this instead:

```
class Transformation(Protocol):
    def __call__(self, ureg: UnitRegistry, val: Magnitude, **kwargs) -> Magnitude:
        ...
```

Happy to make a PR, but wanted to check the intent.

mikeoconnor0308:
I've manually tested this. Running mypy on tests or examples might be a good way to catch some of these things. 

- [x] Closes #1801 

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ Pint Changelog
 0.23 (unreleased)
 -----------------
 
-- Nothing changed yet.
+- Fixed Transformation type protocol.
 
 
 0.22 (2023-05-25)
--- a/pint/facets/context/objects.py
+++ b/pint/facets/context/objects.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 
 import weakref
 from collections import ChainMap, defaultdict
-from typing import Any, Callable, Protocol, Generic, Optional
+from typing import Any, Callable, Protocol, Generic, Optional, TYPE_CHECKING
 from collections.abc import Iterable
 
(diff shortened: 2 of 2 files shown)


hgrecco:
Somehow this broke the numpy code, not sure why. Ideas?

mikeoconnor0308:
i have no idea, i guess something must have been using the `Transformation` protocol but and passing something that should have been a `kwarg` as a position argument?

PR #1819, opened 2023-07-16
Short circuit for pint arrays

ksunden:
Avoid iterating over large arrays that share unit information

We (matplotlib) got a report of large arrays taking a long time to plot.

I noticed that the conversion interface in `matplotlib.py` was doing a list 
comprehension for all iterables, including those that could be converted more directly.

This adds the short circuit from `_convert_value` to `convert` prior to list comprehension

- [n/a] Closes # (insert issue number)

### DIFF
--- a/pint/matplotlib.py
+++ b/pint/matplotlib.py
@@ -34,6 +34,9 @@ class PintConverter(matplotlib.units.ConversionInterface):
 
     def convert(self, value, unit, axis):
         """Convert :`Quantity` instances for matplotlib to use."""
+        # Short circuit for arrays
+        if hasattr(value, "units"):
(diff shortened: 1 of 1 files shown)


ksunden:
FYI, the reporter (unfortunately) needs to work on python 3.8, so if there could be a release that supports that version and has this patch, that would be ideal (otherwise they may have to more manually configure the mpl converter, which is possible, but not ideal)

(But also I do understand wanting to follow NEP 29/SPEC 0 timelines for support, so, ultimately up to you)

hgrecco:
Thanks a lot for th PR (and matplotlib!). We try to adhere to NEP 29 but let me see what we can do.

PR #1847, opened 2023-09-16
add `pint-xarray`'s CI to the downstream status page

keewis:


keewis:
I wondered if we could improve the page by formatting the list as a table, and the result is the last commit here. This does make the page look a bit nicer, although `mdformat` expands the source to be equally sized everywhere, including the huge urls, which makes editing a bit hard (does anyone know if we can avoid that somehow?).

hgrecco:
@keewis I agree that the rendered table looks much nicer! I do not know if there is a way to improve the non rendered format. Is there a way to use indirection?

keewis:
there is in [github flavored markdown](https://github.github.com/gfm/#images), not sure about others. I guess it depends on what we want to do with that page?

hgrecco:
IMO is only for github. We use it to check if everything is ok before releasing!

Issue #1792, opened 2023-05-31
Performance of wrapper


Saelyos:
I'm using pint in my library with some wrappers that give a nice user interface, but unfortunately they also take a lot of time. 
Basically I'm using a code that looks like this:  

```python
from pint import UnitRegistry

ureg = UnitRegistry()


class Foo:
    def __init__(self, q):
        self.q = q

    @property
    @ureg.wraps("meter", (None,))
    def q(self) -> ureg.Quantity:
        return self._q

    @q.setter
    @ureg.wraps(None, (None, "meter"), strict=False)
    def q(self, value: float) -> None:
        self._q = value
```
I'm then using `self._q` (without unit) for some relatively heavy computations.

Unfortunately most of the time is spent in the object creations . This is particularly frustrating with `strict=False`, when even calling `Foo(100)` takes a lot of time, as one could expect a more transparent process.

Yet I'm not sure if anything can be done to speed up the process. In case it can help, I've attached the profiler results from my application showing the time spend in pint: 

![image](https://github.com/hgrecco/pint/assets/34129209/0f6f0447-68c2-4189-829e-a0a5c910a615)

matroscoe:
I have also been running into significant performance bottle necks as documented in: https://github.com/hgrecco/pint/issues/1750

hgrecco:
Performance will be address in the next version. Would you like to help developing or improving the benchmark?

Saelyos:
I'm not sure how much time I'll be able to spend on this, but I'll try to help improving the performances

hgrecco:
I pushed some benchmarks. The results are : https://codspeed.io/hgrecco/pint/benchmarks

There are still things to polish, but is a good start.

hgrecco:
https://github.com/hgrecco/pint/discussions/1820

PR #1862, opened 2023-10-20
Wraps benchmark

Saelyos:
In relation to #1792, I added some basic tests to cover the performances of the wrapper. I also made some small changes to improve the performances (about 3x improvement):

image

There are definitely some ways to improve it again  (especially by removing the call to `bind`), but that would require bigger internal changes. 


hgrecco:
Thanks for the PR!

PR #1880, opened 2023-11-30
Avoid numpy scalar deprecation warnings in __setitem__

dopplershift:
NumPy as of 1.25 deprecated automatically converting any "scalar" with non-zero number of dimensions to a float value. Therefore, we should ensure our values have ndim == 0 before passing to math.isnan()


hgrecco:
Thanks!

PR #1889, opened 2023-12-06
Add [+-]portion.inf to upcast_types

### LINKED ISSUE #1888: Quantity intervals
I'm trying to use `pint` with [portion](https://github.com/AlexandreDecan/portion) for quantity intervals. Portion defines its own singletons for positive and negative infinity that can compare to any type. However pint quantities can be compared only with other quantities:
```py
In [1]: import pint, portion

In [2]: portion.inf > pint.Quantity("3 mg")
Out[2]: True

In [3]: -portion.inf > pint.Quantity("3 mg")
Out[3]: False

In [4]: pint.Quantity("3 mg") < portion.inf
ValueError: Cannot compare PlainQuantity and <class 'portion.const._PInf'>

In [5]: pint.Quantity("3 mg") < -portion.inf
ValueError: Cannot compare PlainQuantity and <class 'portion.const._NInf'>
```

Now pint can parse `inf` as native float but this is not exactly equivalent to `portion.inf`, for example you can create a closed interval with infinite bounds:
```py
In [6]: portion.closed(pint.Quantity("-inf mg"), pint.Quantity("inf mg"))
Out[6]: [<Quantity(-inf, 'milligram')>,<Quantity(inf, 'milligram')>]
```

I think this can be fixed on the pint side by returning `NotImplemented` instead of raising `ValueError` for unhandled comparisons. 

cc @AlexandreDecan in case this can be somehow handled on the `portion` side.

gsakkis:
- [x] Closes #1888

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Pint Changelog
 0.23 (unreleased)
 -----------------
 
+- Add [+-]portion.inf to upcast types.
 - Add _get_conversion_factor to registry with cache.
 - Homogenize input and ouput of internal regitry functions to
   facility typing, subclassing and wrapping.
--- a/pint/compat.py
+++ b/pint/compat.py
@@ -221,6 +221,8 @@ upcast_type_names = (
     "pandas.core.series.Series",
     "pandas.core.frame.DataFrame",
     "xarray.core.dataarray.DataArray",
+    "portion.const._PInf",
+    "portion.const._NInf",
 )
 
 #: Map type name to the actual type (for upcast types).


AlexandreDecan:
Notice that the two paths (`portion.const._PInf` and `portion.const._NInf`) are part of the private API of `portion`. I don't know to which extent this private API will be subject to changes in the future. At the same time, I don't see any other good alternative :-)

gsakkis:
@andrewgsavage could you take a look?

andrewgsavage:
looks fine, just needs a maintainer to merge it.



> Notice that the two paths (`portion.const._PInf` and `portion.const._NInf`) are part of the private API of `portion`. I don't know to which extent this private API will be subject to changes in the future. At the same time, I don't see any other good alternative :-)

just needs a new PR with changes if you do end up changing it.

andrewgsavage:
Why was this closed? Do you not want it merged?

PR #1910, opened 2023-12-30
to_compact: support uncertainties' Magnitudes

### LINKED ISSUE #584: Question: to_compact for Measurements
I guess this is somewhat related to #350, #566.

A `Quantity` object has the method `to_compact`, however if using a `Measurement` instance, the inherited `to_compact` method returns without any formatting because the magnitude is a ` uncertainties.ufloat` which is not a `numbers.Number` type object. This is caused by the [following lines](https://github.com/hgrecco/pint/blob/master/pint/quantity.py#L432):

```python
if not isinstance(self.magnitude, numbers.Number):
    msg = ("to_compact applied to non numerical types "
           "has an undefined behavior.")
    w = RuntimeWarning(msg)
    warnings.warn(w, stacklevel=2)
    return self
```

demo of behavior (`Python 3.6.0 (v3.6.0:41df79263a11, Dec 23 2016, 08:06:12) [MSC v.1900 64 bit (AMD64)] on win32` running on `win7x64 pro`): 
```python
# current behavior
>>> import pint
>>> M_ = ureg.Measurement
>>> m1 = M_(10, 0.5, ureg.ohm)*1e3
>>> m1.to_compact()
__main__:1: RuntimeWarning: to_compact applied to non numerical types has an undefined behavior.
<Measurement(10000.00, 500.00, ohm)>
```
Unfortunately I found the [docs](https://github.com/hgrecco/pint/blob/master/docs/measurement.rst) less than helpful in this case. Am I missing something on how to get the following output?:
```python
# desirable behavior
>>> import pint
>>> M_ = ureg.Measurement
>>> m1 = M_(10, 0.5, ureg.ohm)*1e3
>>> m1.to_compact()  #  what method to call?
<Measurement(10.00, 0.50, kiloohm)>
```

If there is no built in method/support for this I think it shouldn't be too hard to implement as a new/overridden method in the `Measurement` class.
The major issue with this approach that comes to mind is:
+ as in #566, if the `Quantity.to_compact` method changes behavior, then this new `Measurement.to_compact` method will have to be modified as well to mirror that behavior in order to provide a uniform experience.

 I wouldn't mind sitting down and taking a crack at it. But before I go poking too far into things I haven't fully read through, I figured I would ask if I was just missing the correct way of presenting Measurements in a more convenient to understand form, as just display formatting will not perform the unit conversion that is desired here. 

It looks like the `Measurement` class will be getting an overhaul coming soon so if this is something that is already planned then great. I've recently taken in interest in including tolerances/variances in demo examples to illustrate more realistic system models and pint/uncertainties has been an intuitive and flexible answer to the goal thus far.

doronbehar:
Fix #584, make to_compact also raise a `TypeError` instead of a `RuntimeWarning`.

- [x] Closes #584
- [ ] ~~Executed `pre-commit run --all-files` with no errors~~ No, but I ran `black` on the files I changed.

Haven't completed the 2 above tasks, as I seek an approval for the change in behavior.

### DIFF
--- a/pint/facets/plain/qto.py
+++ b/pint/facets/plain/qto.py
@@ -4,8 +4,6 @@ from typing import TYPE_CHECKING, Optional
 
 import bisect
 import math
-import numbers
-import warnings
 
 from ...util import infer_base_unit
 from ...compat import (
@@ -100,18 +98,8 @@ def to_compact(
     <Quantity(10.0, 'millinewton')>
     """
 
-    if not isinstance(quantity.magnitude, numbers.Number):
-        msg = "to_compact applied to non numerical types " "has an undefined behavior."
-        w = RuntimeWarning(msg)
-        warnings.warn(w, stacklevel=2)
-        return quantity
 
-    if (
-        quantity.unitless
-        or quantity.magnitude == 0
-        or math.isnan(quantity.magnitude)
-        or math.isinf(quantity.magnitude)
-    ):
+    if quantity.unitless:
         return quantity
 
     SI_prefixes: dict[int, str] = {}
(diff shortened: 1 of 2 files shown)


andrewgsavage:
you can keep the warning message like so:
https://github.com/hgrecco/pint/pull/1911/files#diff-6e7c1281b7440a613fa4346f563a9d5ab8cdbe76bdc2f4cdb19a90ede1545e05R103

PR #1926, opened 2024-01-23
Dim sort

### LINKED ISSUE #1841: Override units display order (1 kW·h displayed as 1 h·kW)

MichaelTiemannOSC:
Adapt PR#1841 to the new Pint formatter.

- [x] Closes #1841 

TODO: show we can override default in registry init file.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,9 @@ Pint Changelog
 0.24 (unreleased)
 -----------------
 
-- Nothing changed yet.
+- Add `dim_sort` function to _formatter_helpers.
+- Add `dim_order` and `default_sort_func` properties to FullFormatter.
+  (PR #????, fixes Issue #1841)
 
 
 0.23 (2023-12-08)
--- a/CHANGES
+++ b/CHANGES
@@ -6,7 +6,7 @@ Pint Changelog
 
 - Add `dim_sort` function to _formatter_helpers.
 - Add `dim_order` and `default_sort_func` properties to FullFormatter.
-  (PR #????, fixes Issue #1841)
+  (PR #1926, fixes Issue #1841)
 
 
 0.23 (2023-12-08)
--- a/pint/delegates/formatter/_format_helpers.py
+++ b/pint/delegates/formatter/_format_helpers.py
@@ -268,6 +268,61 @@ def format_compound_unit(
     return out
 
 
+def dim_sort(items: Iterable[tuple[str, Number]], registry: UnitRegistry):
+    """Sort a list of units by dimensional order (from `registry.formatter.dim_order`).
+
+    Parameters
+    ----------
+    items : tuple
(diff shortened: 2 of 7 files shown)


MichaelTiemannOSC:
This test creates a UnitsContainer with `meter` inserted before `kilogram`:
```
    def test_unit_formatting(self, subtests):
        x = self.U_(UnitsContainer(meter=2, kilogram=1, second=-1))
        for spec, result in (
            ("{}", str(x)),
            ("{!s}", str(x)),
            ("{!r}", repr(x)),
	    (
                "{:L}",
                r"\frac{\mathrm{kilogram} \cdot \mathrm{meter}^{2}}{\mathrm{second}}",
            ),
            ("{:P}", "kilogram·meter²/second"),
            ("{:H}", "kilogram meter<sup>2</sup>/second"),
            ("{:C}", "kilogram*meter**2/second"),
            ("{:Lx}", r"\si[]{\kilo\gram\meter\squared\per\second}"),
            ("{:~}", "kg * m ** 2 / s"),
            ("{:L~}", r"\frac{\mathrm{kg} \cdot \mathrm{m}^{2}}{\mathrm{s}}"),
            ("{:P~}", "kg·m²/s"),
	    ("{:H~}", "kg m<sup>2</sup>/s"),
            ("{:C~}", "kg*m**2/s"),
        ):
            with subtests.test(spec):
                assert spec.format(x) == result
```
The `UnitsContainer` formatter, by default, sorts these alphabetically, putting `kilogram` before `meter**2`.  But if the dimensional sort function is `None`, it imposes no sort, leaving insertion order to rule.  Which leaves `meter**2` before `kilogram`.

This argues for the idea that we must default sorting compound units with `sorted` (by alpha).

hgrecco:
> I don't know why that one test failed--looks like a CI hiccup.

I can fix that in my computer. But how do you think we should merge #1864 and #1926 (this?) Do want to merge them into one and I merge them?

MichaelTiemannOSC:
I think this is the one PR to merge (the other is obsoleted by this one).  We can certainly add some more tests after merging that fleshes out how complex multi-unit dimensions should be sorted, but that will only change the inner operation of the `dim_sort` function, not the interfaces (I think).

PR #1935, opened 2024-02-14
usedforsecurity=False on hash functions

### LINKED ISSUE #1934: Use `usedforsecurity=False` when calling `hashlib.blake2b` to avoid crashing on FIPS enabled systems
As reported here:
- https://github.com/simonw/datasette/issues/2270

Pint currently uses the `hashlib.blake2b` hashing function: https://github.com/hgrecco/pint/blob/3cc2d365af9093cc258caa766e39d17de69d452d/pint/_vendor/flexparser.py#L1118-L1123

On systems with FIPS enabled (I only just learned about these today, see [my notes here](https://til.simonwillison.net/python/md5-fips)) this will result in a runtime error that looks something like this:

    ValueError: [digital envelope routines] unsupported

The fix as-of Python 3.9 is to pass `usedforsecurity=False` to the `hashlib.blake2b` constructor, [documented here](https://docs.python.org/3/library/hashlib.html#hashlib.blake2b).

To continue supporting Python 3.8 you can use the pattern I demonstrate in https://til.simonwillison.net/python/md5-fips 

```python
def md5_not_usedforsecurity(s):
    try:
        return hashlib.md5(s.encode("utf8"), usedforsecurity=False).hexdigest()
    except TypeError:
        # For Python 3.8 which does not support usedforsecurity=False
        return hashlib.md5(s.encode("utf8")).hexdigest()
```

simonw:
Closes #1934

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,8 +4,7 @@ Pint Changelog
 0.24 (unreleased)
 -----------------
 
-- Nothing changed yet.
-
+- Hash functions use usedforsecurity=False for compatibility with FIPS systems (See #1943)
 
 0.23 (2023-12-08)
 -----------------
--- a/pint/_vendor/flexcache.py
+++ b/pint/_vendor/flexcache.py
@@ -283,7 +283,7 @@ class DiskCache:
 
         The naming strategy is defined by the header class used.
         """
-        hd = self._hasher()
+        hd = self._hasher(usedforsecurity=False)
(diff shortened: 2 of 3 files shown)


simonw:
Looks like this PR is obsolete according to this comment:
- https://github.com/hgrecco/pint/issues/1934#issuecomment-1943019196

PR #1940, opened 2024-02-26
have year shortened to 'yr' not 'a'

mpsheasby:

Unsure as to why year is abbreviated to 'a' and propose we abbreviate to 'yr' instead.

Have made proposal accordingly, although not tested the result of these changes, please let me know if you agree then we can test the change has the desired results.

### DIFF
--- a/pint/default_en.txt
+++ b/pint/default_en.txt
@@ -181,7 +181,7 @@ hour = 60 * minute = h = hr
 day = 24 * hour = d
 week = 7 * day
 fortnight = 2 * week
-year = 365.25 * day = a = yr = julian_year
+year = 365.25 * day = yr = julian_year
 month = year / 12
 
 # decade = 10 * year


dalito:
Why "a"? From latin "annum".

mpsheasby:
Thanks @dalito .

In my user group (structural engineers) everyone was perplexed, but I suspect this may be a question of technical background/ industry.

Thoughts from anyone else on which is better?

hgrecco:
Indeed, it is from latin "annum". I am not sure if there is a canonical way. I would say the following: if there is a "rule" (i.e. from NIST) that suggest 'yr' I would agree to change it.

dalito:
The IEEE-Style manual for example suggests "a".  https://journals.ieeeauthorcenter.ieee.org/your-role-in-article-production/ieee-editorial-style-manual/

The current SI brochure (and NIST SI-guides) suggest to not abbreviate units. Note that "year" is not among the accepted non-SI units so no unit symbol is suggested in the SI brochure.

LLMs (gemini/chatGPT) are insisting on "yr" and cite sources incorrectly (e.g. gemini suggested the IEEE-Style guide that explicitly lists "a" as abbreviation as a source for "yr").

So maybe "yr" could be added as another abbreviation but it should not replace "a".

mpsheasby:
Thank you for looking into this further. I have some more considerations to add:

See also this wikipedia article 
https://en.wikipedia.org/wiki/Year

"In English, the [unit of time](https://en.wikipedia.org/wiki/Unit_of_time) for year is commonly abbreviated as "y" or "yr". The symbol "a" (for [Latin](https://en.wikipedia.org/wiki/Latin_language): annus, year) is sometimes used in scientific literature, though its exact duration may be inconsistent."

If you search "what does "a" mean unit" in google you don't get anything suggesting "year". 

If you search "what does "yr" mean unit" in google the top results point to "year".

All of this (and the LLM results) point to "yr" or "y" as being the commonly used and understood, and "a" and understood and difficult to understand if you were not already familiar (not even coming up on google search results). So from a useability perspective, I would prefer "yr".

mpsheasby:
To add, despite the summary at the start, there is some fairly detailed discussion further down that Wikipedia article, which does offer some more conflicting points. 

https://en.wikipedia.org/wiki/Year#Symbols_and_abbreviations

hgrecco:
While I do personally prefer `yr`, I agree with @dalito. We should keep yr as secondary abbreviation but keep `a` as default. Having said that, we need a better way to modify small things within the definitions programatically.

PR #1944, opened 2024-03-04
Enrico/bug context

em1208:


em1208:
@hgrecco Yes, the problem is that when the context is enabled the cache should be rebuilt otherwise currently an exception is raised.

hgrecco:
@em1208 This is the proposal https://github.com/hgrecco/pint/discussions/1779

PR #1954, opened 2024-03-12
remove all mentions of `cumproduct`

keewis:

`numpy=2.0` will bring a lot of breaking changes, including the removal of `cumproduct`. I didn't check how much `pint` will be affected by this, the removal of `cumproduct` is what I saw in `xarray`'s CI.

### DIFF
--- a/pint/facets/numpy/numpy_func.py
+++ b/pint/facets/numpy/numpy_func.py
@@ -965,7 +965,7 @@ def implement_single_dimensionless_argument_func(func_str):
         return a._REGISTRY.Quantity(func(a_stripped, *args, **kwargs))
 
 
-for func_str in ("cumprod", "cumprod", "nancumprod"):
+for func_str in ("cumprod", "nancumprod"):
     implement_single_dimensionless_argument_func(func_str)
 
 # Handle single-argument consistent unit functions
--- a/pint/facets/numpy/numpy_func.py
+++ b/pint/facets/numpy/numpy_func.py
@@ -965,7 +965,7 @@ def implement_single_dimensionless_argument_func(func_str):
         return a._REGISTRY.Quantity(func(a_stripped, *args, **kwargs))
 
 
-for func_str in ("cumprod", "cumproduct", "nancumprod"):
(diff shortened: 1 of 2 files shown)


hgrecco:
I think this break current numpy support, right? 

I was thinking that after 0.24 (which will be released early april), we do a release with all the numpy 2.0 upgrades + finishing discussing and taking action on the QuantityScalar/QuantityArray issue.

mattwthompson:
> I think this break current numpy support, right?

No, `numpy.cumprod` is in 1.x: https://numpy.org/doc/stable/reference/generated/numpy.cumprod.html

`numpy.cumproduct` is already deprecated in favor of `numpy.cumprod` in 1.25:

> np.cumproduct is deprecated. Use np.cumprod instead.

([gh-23314](https://github.com/numpy/numpy/pull/23314))

https://numpy.org/doc/stable/release/1.25.0-notes.html

pint/facets/numpy/numpy_func.py
```diff
@@ -965,7 +965,7 @@ def implementation(a, *args, **kwargs):
         return a._REGISTRY.Quantity(func(a_stripped, *args, **kwargs))
 
 
-for func_str in ("cumprod", "cumproduct", "nancumprod"):
+for func_str in ("cumprod", "cumprod", "nancumprod"):
```
mattwthompson, line 968:
```suggestion
for func_str in ("cumprod", "nancumprod"):
```

hgrecco:
Not sure while the test are failing. In master the docs/bench are failing but lint/ci are green. Here I am getting a weird error related to runtime warnings that started to appear in main after pytest was upgraded and required me to fix certain warnings. Somehow, this is not applying the PR To the current master but to the old. Any ideas?

mattwthompson:
This branch might be out of date with upstream? https://github.com/keewis/pint/commits/cumproduct/

pint/facets/numpy/numpy_func.py
keewis: thanks, I totally missed that `cumprod` was already there.

PR #1959, opened 2024-04-02
avoid calling str on array

### LINKED ISSUE #1957: `PlainQuantity.ndim` stringifys `magnitude`, which is expensive for arrays
Hi,

When profiling a poorly performing script built on pint-xarray, I noticed that most of the time was spent in [`PlainQuantity.ndim`](https://github.com/hgrecco/pint/blob/f2e4081aee38f850938048beac7fb69c4908bc5e/pint/facets/plain/quantity.py#L140C9-L140C13), specifically in the `str(self.magnitude)` call. When `magnitude` is a (rather large) array, stringifying it is actually pretty expensive. Probably we can do better here, no?

Code to reproduce the problem:

```python
Python 3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import pint

In [2]: ureg = pint.UnitRegistry()

In [3]: import numpy as np

In [4]: a = np.random.rand(200, 500, 800) * ureg.m

In [9]: %timeit a.ndim
414 µs ± 5.12 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [10]: %timeit a.magnitude.ndim
46.8 ns ± 0.808 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [11]: pint.__version__
Out[11]: '0.23'
```

I'm not quite sure why the `str(self.magnitude)` call is necessary, with a bit of context I could probably prepare a PR which fixes this.

Thanks and cheers,

Mika

andrewgsavage:
- [x] Closes #1957 

### DIFF
--- a/pint/facets/plain/quantity.py
+++ b/pint/facets/plain/quantity.py
@@ -140,7 +140,7 @@ class PlainQuantity(Generic[MagnitudeT], PrettyIPython, SharedRegistryObject):
     def ndim(self) -> int:
         if isinstance(self.magnitude, numbers.Number):
             return 0
-        if str(self.magnitude) == "<NA>":
+        if str(type(self.magnitude)) == "NAType":
             return 0
         return self.magnitude.ndim
 


mikapfl:
Nice, this reduces the runtime of my tests in a library using pint-xarray from ~50 s down to less than 10 s.

Issue #676, opened 2018-08-16
Feature request: What kind of quantity a unit represents


sigvaldm:
Hi, it would be really cool if it was possible to get what kind of physical quantity a certain dimensionality represents, for instance if a unit is kg*m/s^2 you could get the string "force". For example:
```
F = ureg['N']
"{:what}".format(F.u) # returns "force"
# or
F.u.what() # returns ["force"]
```
I realize this is not trivial. It would require some sort of dictionary of combination of units, and mappings to what they are. Perhaps some combinations could be several things? In which case you would need to return a list. All possible combinations could of course not be covered, but one could cover relatively standard ones as a beginning: length, speed, acceleration, distance, force, temperature, magnetic flux density, and so forth.

hgrecco:
This somehow similar to the discussion that came up when building `to_compact` and finding the most compact representation was too complex for general quantities. 

But we can cover a limited, common cases building on what we have. This is partly covered by the `dimensionality` attribute in the quantity, but it returns everything in terms of base dimensions. For example:
```python
>>> ureg('meter').dimensionality
<UnitsContainer({'[length]': 1.0})>
>>> ureg('are').dimensionality
<UnitsContainer({'[length]': 2.0})>
```
We could try to match this to other derived dimensionalities:
```
[area] = [length] ** 2
```
under a different attribute name. Once we have that, we could create a format specificer. What do you think?

sigvaldm:
TL;DR: I like your suggestion @hgrecco. 

Concerning the example from WolframAlpha where kg\*m/s^2 can be both force, weight, thrust and tension: Weight is a kind of force, but a force is not necessarily a weight. The same goes for thrust and tension. I suppose everything that's measured in kg*m/s^2 must be a force, no? As far as I can imagine, Coulomb always measures charge, m/s always velocity and eV always energy (can you think of any counter-example?). Then I see two options, either 1) the attribute is the "always correct" string or 2) a list of compatible ones. I'm inclined towards the first option. If we opt for 1 but need 2, it will be discovered soon enough. If we opt for 2 but don't need it, we have introduced meaningless complexity, and it may never be reviewed again.

If we do go for option 2 I would still insist that there should be a "preferred" dimensionality (e.g. the first one). As far as I am concerned this is not just for debugging. It can also be used to automatically label things, but it would be unfortunate if some quantity is labeled as "tension" if it is in fact a thrust. Labeling it as a force wouldn't be too bad, though.

[...]

PR #1967, opened 2024-04-10
QuantityKinds

andrewgsavage:

### DIFF
--- a/pint/default_en.txt
+++ b/pint/default_en.txt
@@ -256,7 +256,7 @@ sverdrup = 1e6 * meter ** 3 / second = sv
 galileo = centimeter / second ** 2 = Gal
 
 # Force
-[force] = [mass] * [acceleration]
+[force] = [mass] * [acceleration] = N
 newton = kilogram * meter / second ** 2 = N
 dyne = gram * centimeter / second ** 2 = dyn
 force_kilogram = g_0 * kilogram = kgf = kilogram_force = pond
@@ -265,7 +265,7 @@ force_metric_ton = g_0 * metric_ton = tf = metric_ton_force = force_t = t_force
 atomic_unit_of_force = E_h / a_0 = a_u_force
 
 # Energy
-[energy] = [force] * [length]
+[energy] = [force] * [length] = J
 joule = newton * meter = J
 erg = dyne * centimeter
 watt_hour = watt * hour = Wh = watthour
@@ -329,7 +329,8 @@ centimeter_H2O = centimeter * water * g_0 = cmH2O = cm_H2O
 sound_pressure_level = 20e-6 * pascal = SPL
 
 # Torque
-[torque] = [force] * [length]
+[moment_arm] = [length] = m
+[torque] = [force] * [moment_arm] = N m
 foot_pound = foot * force_pound = ft_lb = footpound
 
 # Viscosity
--- a/pint/delegates/txt_defparser/plain.py
+++ b/pint/delegates/txt_defparser/plain.py
@@ -235,8 +235,8 @@ class DerivedDimensionDefinition(
         if not (s.startswith("[") and "=" in s):
             return None
 
-        name, value, *aliases =  (p.strip() for p in s.split("="))
-        
+        name, value, *aliases = (p.strip() for p in s.split("="))
+
         preferred_unit = None
         if aliases:
             if aliases[0] == "_":
--- a/pint/delegates/txt_defparser/plain.py
+++ b/pint/delegates/txt_defparser/plain.py
@@ -221,11 +221,11 @@ class DerivedDimensionDefinition(
 ):
     """Definition of a derived dimension::
 
-        [dimension name] = <relation to other dimensions>
+        [dimension name] = <relation to other dimensions> [= <preferred unit>]
 
     Example::
 
-        [density] = [mass] / [volume]
+        [density] = [mass] / [volume] = kilogram / meter ** 3
     """
 
     @classmethod
@@ -235,7 +235,16 @@ class DerivedDimensionDefinition(
         if not (s.startswith("[") and "=" in s):
             return None
 
-        name, value, *aliases = s.split("=")
+        name, value, *aliases =  (p.strip() for p in s.split("="))
+        
+        preferred_unit = None
+        if aliases:
+            if aliases[0] == "_":
+                aliases = aliases[1:]
+            else:
+                preferred_unit, *aliases = aliases
+
+            aliases = tuple(alias for alias in aliases if alias not in ("", "_"))
 
         if aliases:
             return common.DefinitionSyntaxError(
@@ -251,7 +260,7 @@ class DerivedDimensionDefinition(
             )
 
         try:
-            return cls(name.strip(), reference)
+            return cls(name.strip(), reference, preferred_unit)
         except Exception as exc:
             return common.DefinitionSyntaxError(str(exc))
 
--- a/pint/facets/plain/definitions.py
+++ b/pint/facets/plain/definitions.py
@@ -234,6 +234,8 @@ class DerivedDimensionDefinition(DimensionDefinition):
 
     #: reference dimensions.
     reference: UnitsContainer
+    # preferred unit 
+    preferred_unit: str | None = None
 
     @property
     def is_base(self) -> bool:
--- a/pint/facets/plain/definitions.py
+++ b/pint/facets/plain/definitions.py
@@ -234,7 +234,7 @@ class DerivedDimensionDefinition(DimensionDefinition):
 
     #: reference dimensions.
     reference: UnitsContainer
-    # preferred unit 
+    # preferred unit
     preferred_unit: str | None = None
 
     @property
--- a/pint/facets/plain/quantity.py
+++ b/pint/facets/plain/quantity.py
@@ -72,7 +72,19 @@ T = TypeVar("T", bound=Magnitude)
 
 def ireduce_dimensions(f):
     def wrapped(self, *args, **kwargs):
+        print(f)
         result = f(self, *args, **kwargs)
+        print(args[0].kind, self.kind, f.__name__)
+        # TODO: Implement this for div, maybe move to mul and div funcs
+        if hasattr(self, "kind") and hasattr(args[0], "kind") and f.__name__ in ["_mul_div"]:
+            result_units_container = UnitsContainer({self.kind: 1, args[0].kind: 1})
+            print(args[0].kind, self.kind)
+            for dim in self._REGISTRY._dimensions.values():
+                if hasattr(dim, "reference") and dim.reference == result_units_container:
+                    return result.to_kind(dim.name)
+            raise ValueError(
+                f"Cannot multiply quantities with kinds {self.kind} and {args[0].kind}"
+            )
         try:
             if result._REGISTRY.autoconvert_to_preferred:
                 result.ito_preferred()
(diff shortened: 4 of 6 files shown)


andrewgsavage:
I got this behaviour working as I wanted:
```python3
from pint import UnitRegistry
ureg = UnitRegistry()
Q_ = ureg.Quantity
moment_arm = Q_(1, "m").to_kind("[moment_arm]")
force = Q_(1, "lbf").to_kind("[force]")
# to_kind converts to the preferred_unit of the kind
assert force.units == ureg.N

# both force and moment_arm have kind defined.
# Torque is defined in default_en:
# [torque] = [force] * [moment_arm]
# the result is a quantity with kind [torque]
torque = force * moment_arm
assert torque.units == ureg.N * ureg.m
assert torque.kind == "[torque]"

# Energy is defined in default_en:
# [energy] = [force] * [length] = J
distance = Q_(1, "m").to_kind("[length]")
energy = force * distance
assert energy.kind == "[energy]"
assert energy.units == ureg.J

# Torque is not energy so cannot be added
with pytest.raises(ValueError):
    energy + torque
```

Here I've used an attribute `Quantity.kind`. Another option is to create a subclass `QuantityKind` that inherits from `Quantity`.

One constraint is the multiplication of `kind_a` * `kind_b` must be unique, ie you cannot have`[energy] = [force] * [distance]` and 
`[torque] = [force] * [distance]`
You can have multiple definitions that result in a quantity, `[energy] = [force] * [distance] = [power] * [time]`, but needs implementing.
I think formulas resulting from 3 quantitykinds are possible, ie `[energy] = [mass] * [specific_heat_capacity] * [delta_temperature]`

I imagine there's a better way to do the lookup when multiplying.

What other examples would be good to test?

andrewgsavage:
Will need a `KindContainer` like `UnitContainer` to hold kinds to allow definitions in terms of 3 parameters, eg
` [specific_heat_capacity] = [energy] / [temperature] / [mass] `

Another thought is the order of operations could matter eg if specific energy is defined,
` [specific_energy] = [energy] / [mass]` 

```
energy = Q_(1, "J").to_kind("[energy]")
mass = Q_(1, "kg").to_kind("[mass]")
temperature = Q_(1, "K").to_kind("[temperature]")

shc = (energy / mass) / temperature)
shc.kind
[specific_energy] / [temperature]
```

andrewgsavage:
Added a `Kind` object similar to `Unit`. I should make it handle mathematical operations, ie `Kind("[force]") * Kind("[distance]") = Kind("[torque] ")` and then use could multiplying/adding `Kind` in `Quantity.__mul__` and other operators to determine output kinds.

I am thinking a QuantityKind would be worth adding, mainly so `isinstance` can be used rather than `hasattr(obj, "_kinds")`. That would be needed for downstream libraries that shouldn't be using private attributes.

andrewgsavage:
functionality to address #676 


```python
force = ureg.Kind("[force]")

# Valid units based on dimensionality
force.compatible_units()
```

    




    frozenset({<Unit('atomic_unit_of_force')>,
               <Unit('dyne')>,
               <Unit('force_gram')>,
               <Unit('newton')>,
               <Unit('force_kilogram')>,
               <Unit('force_metric_ton')>})




```python
# Kinds that when multiplied together with the specified exponents give ['force']
force.kind_relations()
```




    {<UnitsContainer({'[area]': 1.0, '[pressure]': 1.0})>,
     <UnitsContainer({'[energy]': 1.0, '[length]': -1.0})>,
     <UnitsContainer({'[gaussian_charge]': 1.0, '[gaussian_magnetic_field]': 1.0})>,
     <UnitsContainer({'[acceleration]': 1.0, '[mass]': 1.0})>,
     <UnitsContainer({'[moment_arm]': -1.0, '[torque]': 1.0})>}




```python
newton = ureg.N
# Kinds that have the same dimenionality as the unit
newton.compatible_kinds()
```




    {'[force]'}




```python
joule = ureg.J
# A better example, (only force is defined with dimensions of kg m s-2)
joule.compatible_kinds()
```




    {'[energy]', '[torque]'}

dalito:
@andrewgsavage Adding support for qunatity kinds would be great. Thanks a lot for starting this. 

Above you wrote

> Here I've used an attribute Quantity.kind. Another option is to create a subclass QuantityKind that inherits from Quantity.

Then you added the `QuantityKind` class instead of going the attribute-route. What was the reason for this decision?
Having just one Quantity class instead of two (`Quantity and `QuantityKind`) would be easier to learn/use.

The name `QuantityKind` was confusing me initially because it is not a class for a *quantity kind* but for a *quantity with kind support*.

I was expecting that QuantityKind instances support quantity methods, but when I tried I got

```
In [1]: from pint import UnitRegistry
   ...: ureg = UnitRegistry()
   ...: Q_ = ureg.Quantity
   ...: K_ = ureg.Kind
In [3]: l = Q_(2, "m").to_kind("[length]")
In [4]: l
Out[4]: <QuantityKind(2, meter, [length])>
In [5]: l.to("foot")
Cell In[5], line 1
----> 1 l.to("foot")

File ~\MyProg_local\gh\pint\pint\facets\plain\quantity.py:538, in PlainQuantity.to(self, other, *contexts, **ctx_kwargs)
    534 other = to_units_container(other, self._REGISTRY)
    536 magnitude = self._convert_magnitude_not_inplace(other, *contexts, **ctx_kwargs)
--> 538 return self.__class__(magnitude, other)

File ~\MyProg_local\gh\pint\pint\facets\kind\objects.py:85, in QuantityKind.__new__(cls, value, kinds, units)
     79     raise TypeError(
     80         "kinds must be of type str, KindKind or "
     81         "UnitsContainer; not {}.".format(type(kinds))
     82     )
     84 if units is None:
---> 85     kk = inst._REGISTRY.Kind(kinds)
     86     if kk.preferred_unit:
     87         units = kk.preferred_unit

File ~\MyProg_local\gh\pint\pint\facets\kind\objects.py:211, in KindKind.__init__(self, kinds)
    209     self._kinds = self._REGISTRY.parse_kinds(kinds)._kinds
    210 else:
--> 211     raise TypeError(
    212         "kinds must be of type str, UnitsContainer; not {}.".format(type(kinds))
    213     )

TypeError: kinds must be of type str, UnitsContainer; not <class 'pint.util.UnitsContainer'>.
```

Other examples that I tried worked fine. Playing with this confirmed how useful quantity kinds are for correct handling of quantities in even more cases.

andrewgsavage:
> > Here I've used an attribute Quantity.kind. Another option is to create a subclass QuantityKind that inherits from Quantity.
> 
> Then you added the `QuantityKind` class instead of going the attribute-route. What was the reason for this decision? Having just one Quantity class instead of two (`Quantity and `QuantityKind`) would be easier to learn/use.

There were a lot of `hasattr(obj, "kind")` checks which didn't feel as 'correct' as `isinstance(obj, QuantityKind)`. I saw the `Measurement` object in the uncertainty facet and moved the kind related code to its own facet. This is nicer as all the kind related code is obvious and in smaller files. I think code for `__new__` is simplier when `QuantityKind` its own class. I added `Quantity.to_kind` for this reason - a way to initialise without modifying `__new__`

I am liking the seperationso far; `Quantity` only provides units/dimensionality checking. `QuantityKind` provides both units/dimensionality and kind checking. You can switch between the two as needed, when methods are implemented. (Although would this actually be needed? Ideally you would be using QuantityKinds only.) Having two classes makes it more obvious as to whether you have objects with kind support or not.

> The name `QuantityKind` was confusing me initially because it is not a class for a _quantity kind_ but for a _quantity with kind support_.
> 
> I was expecting that QuantityKind instances support quantity methods, but when I tried I got
> 
> ```
> In [1]: from pint import UnitRegistry
>    ...: ureg = UnitRegistry()
>    ...: Q_ = ureg.Quantity
>    ...: K_ = ureg.Kind
> In [3]: l = Q_(2, "m").to_kind("[length]")
> In [4]: l
> Out[4]: <QuantityKind(2, meter, [length])>
> In [5]: l.to("foot")
> ---------------------------------------------------------------------------
> TypeError                                 Traceback (most recent call last)
> Cell In[5], line 1
> ----> 1 l.to("foot")
> 
> File ~\MyProg_local\gh\pint\pint\facets\plain\quantity.py:538, in PlainQuantity.to(self, other, *contexts, **ctx_kwargs)
>     534 other = to_units_container(other, self._REGISTRY)
>     536 magnitude = self._convert_magnitude_not_inplace(other, *contexts, **ctx_kwargs)
> --> 538 return self.__class__(magnitude, other)
> 
> File ~\MyProg_local\gh\pint\pint\facets\kind\objects.py:85, in QuantityKind.__new__(cls, value, kinds, units)
>      79     raise TypeError(
>      80         "kinds must be of type str, KindKind or "
>      81         "UnitsContainer; not {}.".format(type(kinds))
>      82     )
>      84 if units is None:
> ---> 85     kk = inst._REGISTRY.Kind(kinds)
>      86     if kk.preferred_unit:
>      87         units = kk.preferred_unit
> 
> File ~\MyProg_local\gh\pint\pint\facets\kind\objects.py:211, in KindKind.__init__(self, kinds)
>     209     self._kinds = self._REGISTRY.parse_kinds(kinds)._kinds
>     210 else:
> --> 211     raise TypeError(
>     212         "kinds must be of type str, UnitsContainer; not {}.".format(type(kinds))
>     213     )
> 
> TypeError: kinds must be of type str, UnitsContainer; not <class 'pint.util.UnitsContainer'>.
> ```
> 
> Other examples that I tried worked fine. Playing with this confirmed how useful quantity kinds are for correct handling of quantities in even more cases.

Ah I hadn't realised that `to` isn't working. Anything that uses `self.__class__` won't work either. I would expect that to return a `<QuantityKind ... ft, [length]>`. I haven't had to convert units as I've been setting preferred units in defintions, which feels very nice!

andrewgsavage:
I found a c++ library thats being written with kinds.

It has some good documentation on how it's dealing with kinds:
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3045r0.html#systems-of-quantities 

And also points, vectors, tensors ... which they are calling character
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3045r0.html#operations-in-the-affine-space

interestingly they are using a single class for quantity/quantitykind

dalito:
Thanks for the reference! This is the best in depth discussion I read so far. They also mentioned the [lack of kind handling in pint](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3045r0.html#quantity-kinds-improve-safety).

dalito:
Wouldn't it be an option to initialize Quantities directly with their kind?

So instead of converting
```
moment_arm = Q_(1, "m").to_kind("[moment_arm]")
```
initialize directly as
```
moment_arm = Q_(1, "m", "[moment_arm]")
```
This would mean adding a new argument `kinds: Optional[KindLike] = None`. What disadvantage would this API have? It seems simpler but may have other implications e.g. for performance.

andrewgsavage:
You can do
`moment_arm = QK_(1, "[moment_arm]","m")`
Or (assume the qk is initiated with the preferred unit for the kind) (this shortcut makes the ordering mag, kind, unit while I think mag, unit, kind is more intuitive)
`moment_arm = QK_(1, "[moment_arm]")

Wha lt would you expect the result of a quantity* quantitykind to be? Should it error, or drop the kind?

I saw that too. The thought that some units can only be of one kind was not obvious to me

andrewgsavage:
Another thing to note is the information they need for their definitions (
point/vector/tensor) isn't included in qudt. Adding kinds adds any more
definitions and relations which will take some work to set up. they dont
have a definition file parse like pint, units are defined in code.

dalito:
The mp-units blog has a good [series of articles](https://mpusz.github.io/mp-units/latest/blog/2024/10/07/international-system-of-quantities-isq-part-1---introduction/) on the international system of quantitites (ISQ) and quantitiy kinds. I did not yet look into how exactly they implement quantity kinds (it is in C++).

PR #1971, opened 2024-04-26
Avoid trapz DeprecationWarning on numpy 2

dopplershift:
I haven't exhaustively gone through the changes in numpy 2, but this is enough to allow MetPy's test suite to pass with no new warnings on numpy 2.

### DIFF
--- a/pint/facets/numpy/numpy_func.py
+++ b/pint/facets/numpy/numpy_func.py
@@ -740,8 +740,11 @@ def _base_unit_if_needed(a):
             raise OffsetUnitCalculusError(a.units)
 
 
+# Can remove trapz wrapping when we only support numpy>=2
 @implements("trapz", "function")
+@implements("trapezoid", "function")
 def _trapz(y, x=None, dx=1.0, **kwargs):
+    trapezoid = np.trapezoid if hasattr(np, "trapezoid") else np.trapz
     y = _base_unit_if_needed(y)
     units = y.units
     if x is not None:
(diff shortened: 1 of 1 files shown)


keewis:
I thought it was `cumproduct`, not `cumprod`, that was removed?

dopplershift:
> I thought it was `cumproduct`, not `cumprod`, that was removed?

Apparently, yes, it was actually just that one and apparently you got that in #1954. This is what I get for starting this from an install of pint 0.24. I'll drop that commit and that will leave only avoiding the deprecation warning if we want.

pint/facets/numpy/numpy_func.py
```diff
@@ -741,21 +741,23 @@ def _base_unit_if_needed(a):
 
 
 @implements("trapz", "function")
```
keewis, line 743: not sure what the policy on deprecations is, but maybe add a comment about removing this once `numpy>=2` is the minimum supported version? That usually makes it much more likely to get cleaned up.

keewis:
if you're fine with making this a general "`numpy>=2` compat" PR, there's also the `ptp` method that got removed.

The implementation itself forwards to `numpy.ndarray`'s methods, so nothing else is necessary, but the method test should be skipped if `numpy>=2` is installed.

pint/facets/numpy/numpy_func.py
dopplershift: Sounds good.

dopplershift:
@keewis I'm ok with doing that, except that skipping the tests should probably go with adding Numpy 2 to the CI matrix, and I really have no clue how best to do that, unless it's a specific `include` on the workflow? If so, what should the other options be?

keewis:
I believe you can add the release candidate to the build matrix:
```yaml
include:
  # numpy>=2
  - python-version: "3.12"
    numpy: "numpy>=2.0.0rc1"
    uncertainties: null
```
(not sure, though)

andrewgsavage:
only 4 failures before this PR, so I'll merge this and finish it up in #1984

Issue #690, opened 2018-09-05
Timedeltas Quantities not acting as expected


dalymople:
I have been playing around with this package (which is fantastic by the way!) and I have stumbled across some issues with the support for datetime.timedetas. Some of the issues appear to be due to the underlying implementation of the datetime.timedelta objects rather than Pint, but the way they are handled does seem a bit strange.

Expanding on the example in issue #608 

```
>>> ureg = UnitRegistry()
>>> a = datetime.timedelta(seconds=4)
>>> t = ureg.Quantity(a, units='seconds')
>>> t
<Quantity(0:00:04, 'second')>
```

Looks good, lets try some unit conversions.

```
>>> t.to('minutes')
<Quantity(0:00:00.066667, 'minute')>
>>> t.to('hours')
<Quantity(0:00:00.001111, 'hour')>
>>> t.to('days')
<Quantity(0:00:00.000046, 'day')>
```

All good, try some more.

```
>>> t.to('milliseconds')
<Quantity(1:06:40, 'millisecond')>
>>> t.to('microseconds')
<Quantity(46 days, 7:06:40, 'microsecond')>
```

That looks strange, but seems to be simply due to how it is displayed.

The bigger issue to do with functionality is the following;

```
>>> d = ureg.Quantity(10, 'meters')
>>> t = ureg.Quantity(datetime.timedelta(seconds=4), units='seconds')
>>> d/t
TypeError: unsupported operand type(s) for /: 'int' and 'datetime.timedelta'
```

That is not completely unexpected, the exception is raised by the underlying datetime.timedelta  implementation rather than Pint. However it doesn't feel right... the only way around this seems to be not to use a datetime.timedelta like so;

```
>>> t = ureg.Quantity(a.total_seconds(), 'seconds')
>>> d/t
<Quantity(2.5, 'meter / second')>
```

Would a pull request addressing these issues be accepted? I have never submitted a PR before but I would be happy to give it a go if there is support for these changes?

jondoesntgit:
What API would you suggest? Show some sample output perhaps?

hgrecco:
Closing. Feel free to reopen if an API is suggested.

PR #1978, opened 2024-05-06
support timedelta in Q_.init

### LINKED ISSUE #608: Convert timedelta to quantity
As per #510, this currently does not work:

```python
>>> ureg = UnitRegistry()
>>> t = ureg.Quantity(datetime.timedelta(seconds=4))
>>> t
<Quantity(0:00:04, 'dimensionless')>
>>> t  == ureg.parse_expression("4 seconds")
False
```
  
Can we get support for converting a timedelta to a quantity?

andrewgsavage:
- [x] Closes #608 #690


pint/compat.py
```diff
@@ -293,6 +298,26 @@ def is_duck_array_type(cls: type) -> bool:
...
+    if isinstance(obj, ndarray) and obj.dtype == np_timedelta64:
+        return True
+
+
+def to_seconds(obj: Any) -> float:
+    """Convert a timedelta object to seconds."""
+    if isinstance(obj, datetime.timedelta):
+        return obj.total_seconds()
+    elif isinstance(obj, np_timedelta64) or obj.dtype == np_timedelta64:
+        return obj.astype(float)
```
keewis, line 317: this is not going to work in general: `timedelta64` has its own units, which can range from days (or years?) to nanoseconds, and `astype(float)` is the equivalent of `magnitude`. I don't have any advice on how to extract that, though, other than parsing the string repr of the dtype.

keewis (commented):
I've got a couple of comments, mostly regarding `timedelta64` and where to put the code handling it.

I would also like to have a way to convert a quantity with dimension `[time]` to a `timedelta64`, but that can be its own PR (the API is not quite clear to me: should `to` special-case units of `timedelta64[us]`, for example?).

jakirkham:
It looks like this is ready for another review

andrewgsavage:
poke @keewis

andrewgsavage:
> I would also like to have a way to convert a quantity with dimension `[time]` to a `timedelta64`, but that can be its own PR (the API is not quite clear to me: should `to` special-case units of `timedelta64[us]`, for example?).

I think to do we should make time into a nonmultiplactive unit like temperature, so there is then a delta_time unit
the new c++ library does this for all units with two different classes, a point and a vector
https://github.com/hgrecco/pint/issues/2120

keewis (commented):
sorry for disappearing for quite a while.

Looks good to me in general, but I have two comments.

PR #1985, opened 2024-05-12
numpy2

andrewgsavage:

### DIFF
--- a/pint/testsuite/test_numpy.py
+++ b/pint/testsuite/test_numpy.py
@@ -448,10 +448,12 @@ class TestNumpyMathematicalFunctions(TestNumpyMethods):
 
     @helpers.requires_array_function_protocol()
     def test_trapzoid(self):
-        helpers.assert_quantity_equal(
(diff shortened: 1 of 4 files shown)


pint/testsuite/test_numpy.py
```diff
@@ -758,9 +768,12 @@ def test_minimum(self):
             np.minimum(self.q, self.Q_([0, 5], "m")), self.Q_([[0, 2], [0, 4]], "m")
         )
 
+    # NP2: Can remove Q_(arr).ptp test when we only support numpy>=2
     def test_ptp(self):
-        assert self.q.ptp() == 3 * self.ureg.m
+        if not np.lib.NumpyVersion(np.__version__) >= "2.0.0b1":
```
keewis, line 773: I would probably use `@pytest.mark.skipif(np.lib.NumpyVersion(np.__version__) >= "2.0.0b1, reason="removed in numpy2")` instead here

hgrecco: agree

PR #1988, opened 2024-05-14
use pytest skip for numpy2 test trapezoid

andrewgsavage:

### DIFF
--- a/pint/testsuite/test_numpy.py
+++ b/pint/testsuite/test_numpy.py
@@ -13,6 +13,14 @@ from pint.testsuite import helpers
 from pint.testsuite.test_umath import TestUFuncs
 
 
+def requires_numpy_2(reason=None):
+    if reason is None:
+        reason = "requires numpy >= 2"
+    return pytest.mark.skipif(
+        np.lib.NumpyVersion(np.__version__) < "2.0.0b1", reason=reason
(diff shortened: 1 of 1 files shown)


pint/testsuite/test_numpy.py
```diff
@@ -447,13 +447,16 @@ def test_trapz(self):
         )
 
     @helpers.requires_array_function_protocol()
+    # NP2: Remove this when we only support np>=2.0
+    @pytest.mark.skipif(
+        np.lib.NumpyVersion(np.__version__) < "2.0.0b1",
+        reason="trapezoid added in numpy2",
+    )
```
keewis, line 454: we might need to make this a function. Something like
```python
def requires_numpy_2(reason=None):
    if reason is None:
        reason = "requires numpy >= 2"
    return pytest.mark.skipif(np.lib.NumpyVersion(np.__version__) < "2.0.0b1", reason=reason)

...

@requires_numpy_2(reason="trapezoid added in numpy2")
def test_trapezoid(self):
    ...
```
could work. Otherwise it is possible to skip the entire test module if `numpy` is not available:
```python
pytest.importorskip("numpy")
```

andrewgsavage:
of course there was a helper for it all along...

PR #2011, opened 2024-06-11
Fix custom formatters needing registry

### LINKED ISSUE #2009: What does `registry=None` mean in custom formatter ?
Hi!

With pint 0.24, the internal workings of `register_unit_format` have changed and the `registry` argument of the wrapped function can now be "None". I do not see in the doc the meaning of that new possible value. Wouldn't a `Unit` object always be related to a registry, even if it's the default one ?

This is causing issues in `cf-xarray` (and downstream) for example (xarray-contrib/cf-xarray#507) where we used `registry` to convert the `UnitContainer` back to a `Unit` (see https://github.com/xarray-contrib/cf-xarray/blob/24dd81b2748db59c3ede68735a200b63b8a3d2d2/cf_xarray/units.py#L17-L38)

EDIT: Also, the way  `register_unit_format` is written, wouldn't `registry` _always_ be None ?

aulemahal:
- [x] Closes #2009

Fixes #2009 according to the suggestions in the thread. I also added a note to the doc to describe the arguments passed to the custom formatters. And added a line to the test.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ Pint Changelog
 0.25 (unreleased)
 -----------------
 
-Nothing added yet.
+- Fix custom formatter needing the registry object. (PR #2011)
 
 
 0.24 (2024-06-07)
--- a/docs/user/formatting.rst
+++ b/docs/user/formatting.rst
@@ -98,7 +98,8 @@ formats:
    '2.3e-06 kilogram ** -1 * meter ** 3 * second ** -2'
 
 where ``unit`` is a :py:class:`dict` subclass containing the unit names and
-their exponents.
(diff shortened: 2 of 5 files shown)


hgrecco:
@aulemahal Have you tested against `cf-xarray`?  Does it fix the issue?

aulemahal:
Yes it does!

```python3
from cf_xarray import units

q = units.units.Quantity(1, 'm')

f"{q.units:cf}"
```
fails with the current `master` and works with this branch.

hgrecco:
Then I think we can merge it and release 0.24.1 What do you think @andrewgsavage ?

andrewgsavage:
Yes , once I get the formatting of dimensionless to match previous behaviour

PR #2012, opened 2024-06-11
format dimensionless

### LINKED ISSUE #2007: Default format of dimensionless Unit: undocumented change in 0.24
Consider the following:
```py
import pint

q = pint.Quantity(1.0, "")
print(f"{q.units}")
```

With pint 0.23:
- The output is `'dimensionless'`.
- With instead `f"{q.units:D}"`, the output is `'dimensionless'`.
- With instead `f"{q.units:~}"`, the output is `''`.

With pint 0.24:
- The output is `''`.
- It seems none of the format specifiers described on the [String formatting specifications](https://pint.readthedocs.io/en/stable/user/formatting.html) can be used to get `'dimensionless'` as output.

This change is not mentioned in the change log, and there was not a PendingDeprecationWarning or FutureWarning to indicate this behaviour would change with 0.24.

andrewgsavage:
- [ ] Closes #2007 

### DIFF
--- a/pint/delegates/formatter/_format_helpers.py
+++ b/pint/delegates/formatter/_format_helpers.py
@@ -211,7 +211,7 @@ def formatter(
             neg_terms.append(power_fmt.format(key, fun(value)))
 
     if not pos_terms and not neg_terms:
-        return ""
+        return "dimensionless"
 
     if not as_ratio:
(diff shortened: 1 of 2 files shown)


andrewgsavage:
how does this look @hgrecco ?

PR #2013, opened 2024-06-12
Added "mu" and "mc" prefixes.

SPKorhonen:
Added "mu" and "mc" as alternatives for SI micro prefix. They are mostly used in chemistry and biomedical fields.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ Pint Changelog
 0.25 (unreleased)
 -----------------
 
-Nothing added yet.
+Added mu and mc as alternatives for SI micro prefix
 
 
 0.24 (2024-06-07)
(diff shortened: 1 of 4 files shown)


andrewgsavage:
do you have a reference showing where theses alternatives are used?

SPKorhonen:
Salutem Dicit.

See Wikipedia: The prefix "mc" is commonly used in healthcare or when the character "μ" is not available; for example, "mcg" commonly denotes a microgram.[2]<https://en.wikipedia.org/wiki/Micro-#cite_note-ISMP_2021-2> The letter "u" is sometimes used instead of "μ" when non-Latin<https://en.wikipedia.org/wiki/Latin_alphabet> characters are not available. (https://en.wikipedia.org/wiki/Micro-)

PR #2014, opened 2024-06-12
Add ℓ as alternative for liter

SPKorhonen:
Added ℓ as an alternative symbol for liter

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ Pint Changelog
 0.25 (unreleased)
 -----------------
 
-Nothing added yet.
+Added ℓ as alternative for liter
 
 
 0.24 (2024-06-07)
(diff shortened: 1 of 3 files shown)


andrewgsavage:
do you have a reference where this symbol is used for liter?

SPKorhonen:
Salutem Dicit.

See Wikipedia: The litre (British English<https://en.wikipedia.org/wiki/British_English> spelling) or liter (American English<https://en.wikipedia.org/wiki/American_English> spelling) (SI symbols L and l,[1]<https://en.wikipedia.org/wiki/Litre#cite_note-SIbrochure-1> other symbol used: ℓ) is a metric unit<https://en.wikipedia.org/wiki/Metric_units> of volume<https://en.wikipedia.org/wiki/Volume>. (https://en.wikipedia.org/wiki/Litre)

PR #2032, opened 2024-07-06
Fix cli uncertainty package import

### LINKED ISSUE #2016: pint-convert fails with uncertainties related error if numpy is not installed as well - document numpy dependence or give a better error message
I'm experiencing the following error when trying out the `pint-convert` executable:

```
$ pint-convert '32 m/s' 'mile/hour'
  File "/nix/store/c46r9a4x6c67dwr7mb0l6d00apb723l2-python3.11-pint-0.24/bin/.pint-convert-wrapped", line 6, in <module>
    from pint.pint_convert import main
  File "/nix/store/c46r9a4x6c67dwr7mb0l6d00apb723l2-python3.11-pint-0.24/lib/python3.11/site-packages/pint/pint_convert.py", line 106, in <module>
    (R_i, g_e, m_u, m_e, m_p, m_n) = uncertainties.correlated_values_norm(
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'uncertainties' has no attribute 'correlated_values_norm'
```

My uncertainties version is `3.2.1` - upstream's latest. I also experienced this error with uncertainties `3.1.7`. The `uncertainties` version should ideally be specified in `pyproject.toml`.

daspk04:
- [x] Closes #2016

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ Pint Changelog
 0.25 (unreleased)
 -----------------
 
-- Nothing added yet.
+- Fix the default behaviour for pint-convert (cli) for importing uncertainties package
 
 
 0.24.1 (2024-06-24)
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@ Pint Changelog
 
 - Fix the default behaviour for pint-convert (cli) for importing uncertainties package
 - Added ℓ as alternative for liter
+- Fix the default behaviour for pint-convert (cli) for importing uncertainties package (PR #2032, Issue #2016)
 - Support permille units and `‰` symbol (PR #2033, Issue #1963)
 
 
--- a/docs/dev/pint-convert.rst
+++ b/docs/dev/pint-convert.rst
@@ -82,7 +82,7 @@ The uncertainty can be enabled with `-U` (by default it is not enabled):
 .. code-block:: console
 
     $ pint-convert -p 20 -U Eh eV
-    1 hartree = 27.211386245988471444 eV
+    1 hartree = 27.211386245988(52) eV
 
 .. code-block:: console
 
--- a/docs/dev/pint-convert.rst
+++ b/docs/dev/pint-convert.rst
@@ -77,36 +77,39 @@ With the `uncertainties` package, the experimental uncertainty in the physical
 constants is considered, and the result is given in compact notation, with the
 uncertainty in the last figures in parentheses:
 
-.. code-block:: console
-
-    $ pint-convert Eh eV
-    1 hartree = 27.21138624599(5) eV
-
-The precision is limited by both the maximum number of significant digits (`-p`)
-and the maximum number of uncertainty digits (`-u`, 2 by default)::
-
-    $ pint-convert -p 20 Eh eV
-    1 hartree = 27.211386245988(52) eV
-
-    $ pint-convert -p 20 -u 4 Eh eV
-    1 hartree = 27.21138624598847(5207) eV
-
-The uncertainty can be disabled with `-U`):
+The uncertainty can be enabled with `-U` (by default it is not enabled):
 
 .. code-block:: console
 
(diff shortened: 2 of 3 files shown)


docs/dev/pint-convert.rst
```diff
@@ -77,36 +77,39 @@ With the `uncertainties` package, the experimental uncertainty in the physical
 constants is considered, and the result is given in compact notation, with the
 uncertainty in the last figures in parentheses:
 
+The uncertainty can be enabled with `-U` (by default it is not enabled):
+
 .. code-block:: console
 
-    $ pint-convert Eh eV
+    $ pint-convert -p 20 -U Eh eV
+    1 hartree = 27.211386245988471444 eV
```
doronbehar, line 85: This conversion and the conversions below are a bit confusing to me. The [CODATA value seems to be `27.211386245981(30) eV`](https://en.wikipedia.org/wiki/Hartree) - it does include uncertainty (and it is also a bit different then this value, but this difference is a topic out of scope for this discussion).

Here you do use the `-U` flag, but no uncertainty is printed?

pint/pint_convert.py
```diff
@@ -171,7 +172,8 @@ def convert(u_from, u_to=None, unc=None, factor=None):
     if factor:
         q *= ureg.Quantity(factor)
         nq *= ureg.Quantity(factor).to_base_units()
-    prec_unc = use_unc(nq.magnitude, fmt, args.prec_unc)
+    if args.unc:
```
doronbehar, line 175: There is a place in this file where this logic happens:

```
if args.unc:
    import uncertainties
```

Perhaps you could put the `import` here, with a `try except` that will print a nice user friendly error asking the user to install `uncertainties` and `numpy`?

doronbehar (commented):
My 2 small comments

pint/pint_convert.py
daspk04: Thanks @doronbehar ! 

I'll add it. Currently it will raise `Import` error, which I assumed was enough for people to understand and install `uncertainties`.

I think I understand the problem that you encountered is an issue with `uncertainties`  package not with `pint`. In my understanding `uncertainties` has `numpy` as optional dependency so even if you import `uncertainties` here in pint, the calculations seems to doesn't work.  The problem is basically [here](https://github.com/lmfit/uncertainties/blob/1238d37c6ed87d018e696589716de488e7d9eca7/uncertainties/core.py#L108), at this place the `uncertainties` should raise an warning not pass it.

docs/dev/pint-convert.rst
daspk04: Thanks @doronbehar !

I think I missed to update this one. Hmm, I tried this an currently this raises the same error  as yours, which is as because of [uncertainties](https://github.com/hgrecco/pint/pull/2032#discussion_r1668622802).

daspk04: I have added a fix for it

pint/pint_convert.py
daspk04: I have added an expection to be raised if `uncertainties` cannot be imported.

CHANGES
```diff
@@ -4,8 +4,10 @@ Pint Changelog
 0.25 (unreleased)
 -----------------
 
+- Fix the default behaviour for pint-convert (cli) for importing uncertainties package
 - Added mu and mc as alternatives for SI micro prefix
 - Added ℓ as alternative for liter
+- Fix the default behaviour for pint-convert (cli) for importing uncertainties package (PR #2032, Issue #2016)
```
doronbehar, line 10: Looks like a duplicate.

doronbehar (commented):
Nice job! Thanks for working on this :).

CHANGES
daspk04: Thanks @doronbehar !

I think, I did it during rebase. Related PR https://github.com/hgrecco/pint/pull/2034

PR #2036, opened 2024-07-12
2035 pandas3

mutricyl:
- [x] Closes 2035 

It is difficult to add test to `is_upcast_type` function since it would need `pandas`, `pint-pandas` an `xarray` to be installed.

I removed one duplicate in the list.

### DIFF
--- a/pint/compat.py
+++ b/pint/compat.py
@@ -240,6 +240,8 @@ upcast_type_names = (
     "xarray.core.variable.Variable",
     "pandas.core.series.Series",
     "pandas.core.frame.DataFrame",
+    "pandas.Series",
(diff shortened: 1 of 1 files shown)


pint/compat.py
```diff
@@ -240,7 +240,8 @@ def _to_magnitude(value, force_ndarray=False, force_ndarray_like=False):
     "xarray.core.variable.Variable",
     "pandas.core.series.Series",
     "pandas.core.frame.DataFrame",
-    "xarray.core.dataarray.DataArray",
```
andrewgsavage, line 243: no need to delete this

mutricyl: ok, reverted

PR #2049, opened 2024-08-21
CODATA 2022 update

Jellby:
Update physical constant values to the latest CODATA 2022 recommended values (from https://pml.nist.gov/cuu/Constants/). Small change, but there's no reason to stay outdated :grin: 


Jellby:
Is there anything I should fix? I don't think any documentation (beyond the changelog) or any new test is needed. And surely the PR is not going to be rejected for those µs in the benchmark?

andrewgsavage:
No, nothing to fix. It needs a maintainer to take the time to confirm the
numbers are all correct.

pint/constants_en.txt
```diff
@@ -46,22 +46,21 @@ wien_wavelength_displacement_law_constant = ℎ * c / (k * wien_x)
...
 # The uncertainty in the last significant digits is given in parentheses as a comment.
 
 newtonian_constant_of_gravitation = 6.67430e-11 m^3/(kg s^2) = _ = gravitational_constant  # (15)
-rydberg_constant = 1.0973731568160e7 * m^-1 = R_∞ = R_inf                                  # (21)
-electron_g_factor = -2.00231930436256 = g_e                                                # (35)
-atomic_mass_constant = 1.66053906660e-27 kg = m_u                                          # (50)
-electron_mass = 9.1093837015e-31 kg = m_e = atomic_unit_of_mass = a_u_mass                 # (28)
-proton_mass = 1.67262192369e-27 kg = m_p                                                   # (51)
-neutron_mass = 1.67492749804e-27 kg = m_n                                                  # (95)
-lattice_spacing_of_Si = 1.920155716e-10 m = d_220                                          # (32)
```
andrewgsavage, line 61: you've removed this and the following K_ constants from the default units (but they're available in pint-convert). Is there a reason why?

Jellby: I've removed them from `pint-convert` too, haven't I?

Those constants, as far as I can see, are only relevant to define the "X units" and the Å*, but the NIST website does not give values or correlations for them. So I decided it's better to define the units as constants directly.

andrewgsavage: thanks that makes senes

PR #2052, opened 2024-09-04
Adding GitHub content protection rule to CI

john-science:
## What is the change?

I have constrained the permissions of the GitHub Actions in this repo.

The permissions I set are maximally constrained so that a bad actor can't use a PR into your repo to do various things like: commit to your main branch, modify a GH "Discussion" in your repo, or author an official release.

## A Little Background

GitHub added [this security feature](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#permissions) in 2021.

This is not a complete solution in itself, of course. But it pairs really well with the setting pint has where Pint team members have to "approve a workflow" from a non-team member, to run CI.

For a shorter discussion of the topic than the full GitHub docs, see [this blog post](https://dev.to/azu/how-to-set-github-actions-s-permissions-hln).

## The Checklist

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -10,6 +10,9 @@ on:
   # performance analysis in order to generate initial data.
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   benchmarks:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@ name: CI
 
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
 jobs:
   test-linux:
     strategy:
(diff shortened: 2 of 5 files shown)


dopplershift:
I'll just note that an alternative is to set default permissions on the whole repo in the settings:

image

john-science:
> I'll just note that an alternative is to set default permissions on the whole repo in the settings:

Sure!

As it happens, I can tell that is not being done for this repo, so I am just trying to help.

There is a GitHub Workflow for this repo (`lint-autoupdate`) that _does_ actually need the write permissions. So, if you set the repo-global permissions, that CI will fail and you will need to make a change to that YAML file to allow it to run.

So, either approach works. But for this repo in particular, a PR will need to be made to allow all the Workflows to run.

PR #2077, opened 2024-11-17
Use SPDX license identifier

cdce8p:
Add the SPDX license identifier as recommended by [PEP 639](https://peps.python.org/pep-0639/).
https://spdx.org/licenses/BSD-3-Clause.html


cdce8p:
This PR has been open without a response for some time now. Going to close it. Please feel free to reuse any part of it if you like.

PR #2090, opened 2024-12-12
Fix return type of PlainQuantity.to

### LINKED ISSUE #2088: Typing: to method returns a PlainQuantity
Running `mypy` on the file below results in the following error: `Incompatible return value type (got "PlainQuantity[Any]", expected "Quantity")`

```py
from pint import Quantity, UnitRegistry

ureg = UnitRegistry()

def quantity_to_mm(quantity: Quantity) -> Quantity:
    return quantity.to(ureg.mm)
```

denialhaag:
- [x] Closes #2088

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@ Pint Changelog
 
 - Add docs to the functions in ``pint.testing`` (PR #2070)
 - Fix round function returning float instead of int (#2081)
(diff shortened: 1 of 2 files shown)


denialhaag:
@andrewgsavage, thanks for fixing the import problem for Python < 3.11! :)

PR #2094, opened 2024-12-16
Humanize time output

### LINKED ISSUE #2080: Create option to humanize time output
This touches on #1459. 

Again, I am (again) trying to get a human friendly version of a time. Somewhat with the same intention why https://pypi.org/project/humanize/ exists. 

Background is I want to show time data in a human friendly way. k, M etc are not helpful here.

102123123 s, very few humans will be able to comprehend. I can do ureg.Quantity('102123123 s').to('a') and get 3.23... a. Nice but I will have to know apriory what the meaningful time-unit is here.

I am proposing to create a modification `to_compact` which will automatically convert units of dimension [time] to a suitable humanfriendly time unit. Suitable time unit is the time unit which leaves the magnitude with at least one but the least amount of integer parts.

Example: 

1001 ms should convert to 1.001 s
59 s should still be 59 s
but 60 s should be 1.0 min 
61 min should be 1.01... h 
and so on. 
I'm not sure if pint implements a decade but it does implement a century  
so finally 
ureg.Quantity('10231231232s ').to_compact(humantime=True) -> 3.23 century
(again millenium seem not to be implemented, eon is, maybe that's another topic to pick up: https://en.wikipedia.org/wiki/Unit_of_time, but not to dilute this post). 

As demonstrated above, `to_compact()`, nor `to_preferred()` nor `to_base_units()` are helpful but don't get the job done out of the box.

andrewgsavage:
- [ ] Closes #2080

@mmarras

I've made this it's own function as:
- the logic is different to to_compact
- there are instances where you may wish to use to_compact and to_human, or to_human alone

I'm open to other names than to_human!


andrewgsavage:
thanks for looking,

>to_compact() works by changing the SI prefix
to_human() works by changing the unit itself

To me this is the key difference between the to_... functions. Though I don't think this is explained well at the moment. I added a page here comparing them https://pint--2289.org.readthedocs.build/en/2289/user/quantity-to.html#to-preferred-section

The logic difference is internal to the function. to_compact deals with SI prefixes so relies on them being powers of 1000 and deals with the unit to apply the unit having an exponent (eg m^3 -> mm^3) to find the best SI prefix. Whereas in to_human I've converted and picked the most suitable - so its more expenmsive to run.

PR #2113, opened 2025-01-05
Add conductivity dimension

### LINKED ISSUE #2112: Add conductivity dimension
I would like to be able to check a quantity for compatibility with conductivity:

```python
from pint import UnitRegistry
ureg = UnitRegistry()
value = 5 * ureg.mS/ureg.cm
value.check("[conductivity]")
```

Currently I fix this by defining the conductivity dimension myself:
`ureg.define("[conductivity] = [conductance]/[length]")`

According to [Wikipedia](https://en.wikipedia.org/wiki/Electrical_resistivity_and_conductivity]) conductivity can have multiple meanings, however only one of these, namely, electrical conductivity `[conductivity] = [conductance]/[length]`, is associated with the topic of units. For this reason I think it is sufficient to define `[conductivity] = [conductance]/[length]`and not `[electrical_conductivity] = [conductance]/[length]`.

Br Eskild

nneskildsf:
- [x] Closes #2112 (insert issue number)

### DIFF
--- a/pint/testsuite/test_issues.py
+++ b/pint/testsuite/test_issues.py
@@ -1349,9 +1349,3 @@ def test_issue2044():
     q = (ufloat(10_000, 0.01) * ureg.m).to_compact()
     assert_almost_equal(q.m.n, 10.0)
     assert q.u == "kilometer"
-
-
-def test_issue2112():
-    ureg = UnitRegistry()
-    conductivity = 1.0 * ureg.mS / ureg.cm
-    assert conductivity.check("[conductivity]") is True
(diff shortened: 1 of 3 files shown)


pint/testsuite/test_issues.py
```diff
@@ -1349,3 +1349,9 @@ def test_issue2044():
     q = (ufloat(10_000, 0.01) * ureg.m).to_compact()
     assert_almost_equal(q.m.n, 10.0)
     assert q.u == "kilometer"
+
+
```
andrewgsavage, line 1353: don't need tests for adding to the default registry

PR #2132, opened 2025-02-14
Upgrade infrastructure to pixi/uv/ruff, bump minimum python version to 3.11 Develop

hgrecco:


.github/workflows/ci.yml
```diff
@@ -1,58 +1,47 @@
-name: CI
+name: Test and release when tagged
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
```
jules-ch, line 7: Might be a good idea to use `os` in the matrix so that we define everything in one job.

.github/workflows/ci.yml
```diff
@@ -1,58 +1,47 @@
...
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        environment:
+          - test-py39
+          - test-py310
+          - test-py311
+          - test-py312
+          - test-py313
         numpy: [null, "numpy>=1.23,<2.0.0", "numpy>=2.0.0rc1"]
         uncertainties: [null, "uncertainties==3.1.6", "uncertainties>=3.1.6,<4.0.0"]
```
jules-ch, line 24: Dunno if its still relevant to test multiple versions of uncertainties

pint/compat.py
```diff
@@ -392,3 +193,216 @@ def zero_or_nan(obj: Any, check_all: bool) -> bool | Iterable[bool]:
     if check_all and is_duck_array_type(type(out)):
         return out.all()
     return out
+
+
+# TODO: remove this warning after v0.10
+class BehaviorChangeWarning(UserWarning):
```
jules-ch, line 199: This is dead code I think we can safely remove it

.github/workflows/ci.yml
```diff
@@ -1,58 +1,47 @@
...
+      - run: pixi run --environment lint lint
+
   test-linux:
+    runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        environment:
+          - test-py39
```
jules-ch: You've removed support for 3.9 and 3.10 in pyproject.toml.

Maybe keep 3.10 for now according to https://numpy.org/neps/nep-0029-deprecation_policy.html
We can still support it until April 04 2025.

pint/util.py
```diff
@@ -304,7 +302,7 @@ def pi_theorem(quantities: dict[str, Any], registry: UnitRegistry | None = None)
 
 def solve_dependencies(
     dependencies: dict[TH, set[TH]],
-) -> Generator[set[TH], None, None]:
+) -> Generator[set[TH]]:
```
jules-ch, line 305: Change this to `Iterator[set[TH]]` from collection.abc

jules-ch: My bad it's still valid

hgrecco: I think they changed this in 3.10 or 3.11, it was not ok before.

.github/workflows/ci.yml
jules-ch: Saw that you're using 3.11 only features though. Well I'mnot against of moving forward. Only thing might be some downstream that still support 3.10 like Metpy and others.

jules-ch:
We could remove requirements.txt to since all is handled by pyproject.toml + PEP621 no ?

pyproject.toml
```diff
@@ -1,106 +1,174 @@
...
+codspeed = ["pint[test-full]", "pytest-codspeed"]
 numpy = ["numpy >= 1.23"]
 uncertainties = ["uncertainties >= 3.1.6"]
 babel = ["babel <= 2.8"]
 pandas = ["pint-pandas >= 0.3"]
 xarray = ["xarray"]
 dask = ["dask"]
 mip = ["mip >= 1.13"]
+matplotlib = ["matplotlib"]
+full = [
```
jules-ch, line 52: I'd use all, most packages use this terminology, 

See:

https://github.com/pandas-dev/pandas/blob/6bcd30397d67c3887288c7a82c2c235ce8bc3c7f/pyproject.toml#L85
https://github.com/pola-rs/polars/blob/947dc072724e8e6c7a0e9084fc97213e7643d526/py-polars/pyproject.toml#L82
https://github.com/fastapi/fastapi/blob/eea196f4a505c2584fafda691f2a0033c4db6622/pyproject.toml#L79

.github/workflows/ci.yml
hgrecco: not sure. My suggestion is that we use this commit to migrate as is, and the use future commits to do this.

pint/compat.py
hgrecco: I think so too. The same hapens with certain imports in compat (Self, Never, etc) but I would rather keep it as is until downstream libraries are aware. Again my suggestion is that we use this commit to migrate as is

.github/workflows/ci.yml
hgrecco: The new Pint version will be released on > April 05, I am not sure if we should. Maybe we can list what would be the burden of keeping Python 3.10

pyproject.toml
hgrecco: Good point! I will change this

.github/workflows/ci.yml
hgrecco: Agreed, but My I will use this commit to migrate as is, and then use future commits to do this.

PR #2141, opened 2025-02-18
Fix lint task in CI

maurosilber:
The lint task (`pixi run lint`) runs `pre-commit run`, which only runs on staged files. It was not doing anything on CI jobs. I added the `--all-files` flag based on the [pre-commit action](https://github.com/pre-commit/action/blob/576ff52938d158a24ac7e009dfa94b1455e7df99/action.yml#L7).

The other commits correspond to running each pre-commit task separately.


andrewgsavage:
thanks, I've used most of this to fix the lint task

PR #2142, opened 2025-02-19
Number comparison with quantity

simutisernestas:

Haven't added tests yet, just a proposal enabling:

```python
import pint

d = pint.Quantity(5, "m")
print(d > 6)
d = pint.Quantity(10, "m")
print(d > 6)
```

For now, `pint` gives an error:

```sh
ValueError: Cannot compare PlainQuantity and <class 'int'>
```

This is something of interest within the integration of `pydantic` and `pint`

### DIFF
--- a/pint/facets/plain/quantity.py
+++ b/pint/facets/plain/quantity.py
@@ -1383,6 +1383,10 @@ class PlainQuantity(Generic[MagnitudeT], PrettyIPython, SharedRegistryObject):
                         return op(self.to_base_units()._magnitude, other)
                     else:
                         raise OffsetUnitCalculusError(self._units)
+            elif isinstance(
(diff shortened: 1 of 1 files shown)


dopplershift:
I wouldn't expect this to ever "just work" because it violates the unit safety. The problem here is that e.g. `d > 6` doesn't specify what the units of `6` are, so there's no way to know what the correct answer is. If you just want to compare the magnitudes (again, completely bypassing pint's unit safety), you can do `d.m > 6`.

simutisernestas:
True. I was more thinking of simple validation checks done by `pydantic` against plain numbers, for instance, making sure physical quantity is positive. In that case it requires some boilerplate to make this work rather than simply going through with the comparison. But in the majority of cases, the unit safety is going to be more important.

Issue #1947, opened 2024-03-04
mypy error: defining a Quantity as multiplication of list and unit raises an error


Luna137:
With the newest version of pint, running mypy on the following script

```
import pint

ureg = pint.UnitRegistry()
position_vector: pint.Quantity = [1, 1, 1] * ureg.m
```
returns the following error:

```
error: Incompatible types in assignment (expression has type "list[int]", variable has type "Quantity")  [assignment]
error: Unsupported operand types for * ("list[int]" and "Unit")  [operator]
```

Is this meant to happen, or is this a bug?

PR #2155, opened 2025-03-08
ensure Measurement handles operations with Quantity

### LINKED ISSUE #2154: Operations between Quantity and Measurement
The [documentation](https://pint.readthedocs.io/en/stable/advanced/measurement.html) states that : 

> Mathematical operations with Measurements, return new measurements following the [Propagation of uncertainty](http://en.wikipedia.org/wiki/Propagation_of_uncertainty) rules.

But it is not clear what is the result of an operation between a Quantity and a Measurements. My own experiments lead to counterintuitive results:

```python
from pint import UnitRegistry, Quantity, Measurement
ureg = UnitRegistry()

basic_quantity = 10 * ureg.m
basic_measurement = (10 * ureg.m).plus_minus(1)

# quantity + quantity = quantity --> All good
assert isinstance(basic_quantity + basic_quantity, Quantity)

# measurement + measurement = measurement --> All good
assert isinstance(basic_measurement + basic_measurement, Measurement)

# quantity + measurement = quantity (?) --> Not what I would expect
mixed = basic_quantity + basic_measurement
assert isinstance(mixed, Quantity)

print(mixed) 
# Output: "20.0+/-1.0 meter"

print(mixed.magnitude.std_dev)
# Output: "1.0"
``` 

The `mixed` variable is a Quantity, while I would expect it to be a Measurement since the operation contains uncertainty. And weirdly, this Quantity contains the appropriate uncertainty value and the magnitude is structured as a Measurement.

Also the string formatting is not working properly with this variable:

```python
print(f"{mixed:P}")
# Output: "20.0+/-1.0 meter"

print(f"{basic_measurement:P}")
# Output: "(10.0 ± 1.0) meter"
```

andrewgsavage:
- [ ] Closes #2154 

### DIFF
--- a/pint/facets/plain/quantity.py
+++ b/pint/facets/plain/quantity.py
@@ -94,6 +94,13 @@ def check_implemented(f):
         other = args[0]
         if is_upcast_type(type(other)):
             return NotImplemented
+        # ensure Measurement handles operations with Quantity
+        if (
+            HAS_UNCERTAINTIES
+            and not type(self).__name__ == "Measurement"
+            and type(other).__name__ == "Measurement"
+        ):
+            return NotImplemented
         # pandas often gets to arrays of quantities [ Q_(1,"m"), Q_(2,"m")]
         # and expects PlainQuantity * array[PlainQuantity] should return NotImplemented
         elif isinstance(other, list) and other and isinstance(other[0], type(self)):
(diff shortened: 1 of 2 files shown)


pint/facets/plain/quantity.py
```diff
@@ -831,8 +836,7 @@ def __iadd__(self, other: datetime.datetime) -> datetime.timedelta:  # type: ign
         ...
 
     @overload
-    def __iadd__(self, other) -> PlainQuantity[MagnitudeT]:
-        ...
+    def __iadd__(self, other) -> PlainQuantity[MagnitudeT]: ...
```
josh-bauer-kobold, line 839: I assume similar things need to be done for multiplication/division?

reteps: Typing/doing this would also close #1947

PR #2173, opened 2025-05-13
Initial attempt to correct support use extra arguments in coversion by default contexts

### LINKED ISSUE #2172: Enabled context application appears to be inconsistent
In an attempt to add non-standard unit transformations, I have found that context application with extra arguments seems inconsistent, at least as I understand it.  I am attempting to enable a context in all the code as shown in the documentation as: 

```python 
>>> ureg.enable_contexts('sp', 'boltzmann')
>>> q.to('Hz')
<Quantity(5.99584916e+14, 'hertz')>
```

However, I am attempting to add an additional (required) keyword argument things become less clear. Here is an example that shows the problem I am having. In this example the extra keyword argument does not get implied w/o including the context name, even though it is enabled.

```
import pint
ureg = pint.UnitRegistry()

def mass_to_volume(u, value, *, density=None):
    """Convert mass to volume using density."""
    density = density or 1000 * u.kilogram / u.meter**3
    return value / density


context = pint.Context("Testing")
context.add_transformation("[mass]", "[volume]", mass_to_volume)
ureg.add_context(context)

ureg.enable_contexts(context.name)
m = ureg.Quantity(1000, "kg")
print(m.to("m**3")) 
print(m.to("m**3", density=999 * ureg.kilogram / ureg.meter**3)) # this density is not applied
print(m.to("m**3", "Testing", density=999 * ureg.kilogram / ureg.meter**3)) # this works
```

If the density in the transformation function is required, you get a TypeError on missing keyword.

```
def mass_to_volume2(u, value, *, density):
    """Convert mass to volume using density."""
    return value / density


context = pint.Context("Testing2")
context.add_transformation("[mass]", "[volume]", mass_to_volume2)
ureg.add_context(context)

ureg.enable_contexts(context.name)
print(m.to("m**3", "Testing2", density=999 * ureg.kilogram / ureg.meter**3)) # this works
print(m.to("m**3", density=999 * ureg.kilogram / ureg.meter**3)) #  TypeError
```

Is this the desired behavior?

aeslaughter:
- [ ] Closes #2172 


aeslaughter:
I am not sure this is the correct solution, but it did work on my local machine for the included test. If this an acceptable approach I can clean up the code and update the documentation.

PR #2187, opened 2025-07-29
fix: flip left/right in np.interp wrapper

### LINKED ISSUE #2184: Bug in interaction with left/right arguments of numpy.interp
Unexpected behavior with 
```
numpy             2.3.1 (and also 1.26.4)
pint              0.24.4
```

Reproduction
```
import numpy as np
from pint import Quantity

x = Quantity(1.0, "meter")
xp = Quantity([2.0, 3.0, 4.0], "meter")
fp = [10, 20, 30]

# BUG: Should return 0 (left value) since x < xp[0], but returns 100 (right value)
result = np.interp(x, xp, fp, left=0, right=100)
print(f"Result: {result} (expected: 0, got: 100)")

```

nshankar:
- [ ] Closes #2184


nshankar:
Having some trouble with `pre-commit run --all-files`, it changed several extra files

andrewgsavage:
nice thanks

PR #2206, opened 2025-08-11
Support numpy.geomspace

oxygen-dioxide:

This PR adds support for the numpy.geomspace function:

## Explanation:

`numpy.geomspace(start, stop, ...)` generates numbers spaced evenly on a log scale (a geometric progression) between start and stop.
For physical quantities, both start and stop must have compatible units. The output should have the same units as the inputs.
This registration ensures that Pint will automatically convert start and stop to consistent units and wrap the output array with the correct unit.

## Example code
```python
import numpy as np
from pint import UnitRegistry

ureg = UnitRegistry()
start = 1 * ureg.meter
stop = 1 * ureg.kilometer
result = np.geomspace(start, stop, num=4)
print(result)
```

### DIFF
--- a/pint/facets/numpy/numpy_func.py
+++ b/pint/facets/numpy/numpy_func.py
@@ -903,6 +903,7 @@ for func_str, unit_arguments, wrap_output in (
     ("append", ["arr", "values"], True),
     ("compress", "a", True),
     ("linspace", ["start", "stop"], True),
+    ("geomspace", ["start", "stop"], True),
     ("tile", "A", True),
     ("lib.stride_tricks.sliding_window_view", "x", True),
     ("rot90", "m", True),


oxygen-dioxide:
I noticed that Pint supports temperature units such as degrees Celsius and Fahrenheit, which have non-zero offsets (i.e., their zero points do not correspond to absolute zero).

This raises an ambiguity when using geomspace with offset units. For example:

```py
import numpy as np
from pint import UnitRegistry
Q_ = ureg.Quantity

start = Q_(26.85, ureg.degC)  # 300 K
end = Q_(89.85, ureg.degC)    # 363 K
result = np.geomspace(start, end, num=3)
```
There are three possible behaviors for handling this case:

1. **Raise an error**: Disallow geometric spacing for offset units like Celsius or Fahrenheit.
2. **Physically correct approach**: Convert the values to Kelvin, perform the geometric spacing, and then convert the results back to Celsius. This would yield [26.85, 56.85, 89.85] °C.
3. **Naive numerical approach**: Ignore the offset and perform the geometric spacing directly on the numerical values in Celsius, resulting in [26.85, 49.1169, 89.85] °C.

Personally, I prefer the second (physically correct) approach, as it is more useful in real engineering scenarios.

Which behavior would you recommend for Pint?

oxygen-dioxide:
I've implemented `geomspace` with approach 1, as `dot` is also implemented in this way (raising error if input has offset).

andrewgsavage:
good catch, we'll also want to do logspace in the same way in future

I'll merge when at the end of the week after the freeze for the next release

andrewgsavage:
I suspect that issue exists for many other functions too. That can be done in a future PR.

PR #2209, opened 2025-08-15
fix numpy.dot error if the first parameter is a pint quantity and the second parameter isn't a pint quantity

oxygen-dioxide:

Before this change, if we run:
```python
from pint import UnitRegistry;ureg=UnitRegistry();import numpy
numpy.dot([1,2], [3,4]*ureg.m)
```
it will throw `AttributeError: 'list' object has no attribute '_is_multiplicative'`

After this change, it will work correctly and output `<Quantity(11, 'meter')>`

### DIFF
--- a/pint/testsuite/test_numpy.py
+++ b/pint/testsuite/test_numpy.py
@@ -477,7 +477,7 @@ class TestNumpyMathematicalFunctions(TestNumpyMethods):
         expected = 3 * self.ureg.m
         result1 = np.dot(a, b)
         helpers.assert_quantity_equal(result1, expected)
-        result2 = np.dot(a, b)
+        result2 = np.dot(b, a)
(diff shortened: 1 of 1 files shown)


andrewgsavage:
what do you think about moving the check to `_base_unit_if_needed`, and construction a dimensionless quantity when needed? ie 
```python3
def _base_unit_if_needed(a):
    if not _is_quantity(a):
        return registry.Quantity(a, "dimensionless")
    else:
            if a.units._REGISTRY.autoconvert_offset_to_baseunit:
                return a.to_base_units()
            else:
                raise OffsetUnitCalculusError(a.units)
```


https://github.com/hgrecco/pint/blob/4a62a4148840fa80b8d92b34aee6a4416ff00b3a/pint/facets/numpy/numpy_func.py#L749

PR #2214, opened 2025-09-06
fix numpy.dot error if the first parameter is a pint quantity and the second parameter isn't a pint quantity

andrewgsavage:


andrewgsavage:
@oxygen-dioxide what do you think of this approach? This should make it sasier to fix the same issue in other functions

PR #2217, opened 2025-09-12
Fix definitions of quarter and drachm

acottuli:
1 quarter = 28 pounds not 28 stones

1 drachm = 1/8 ounce apothecaries not 1/8 ounce avoirdupois

Reference
- https://www.legislation.gov.uk/uksi/1994/2867/schedule/made

### DIFF
--- a/pint/default_en.txt
+++ b/pint/default_en.txt
@@ -616,11 +616,11 @@ neper = 1 ; logbase: 2.71828182845904523536028747135266249775724709369995; logfa
 @end
 
 @group Avoirdupois
-    dram = pound / 256 = dr = avoirdupois_dram = avdp_dram = drachm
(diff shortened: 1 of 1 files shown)


acottuli:
@andrewgsavage could you please review this PR and the following three PRs (https://github.com/hgrecco/pint/pull/2218, https://github.com/hgrecco/pint/pull/2219, https://github.com/hgrecco/pint/pull/2220) when you get a chance?

All four PRs are just small atomic fixes for units defined in Pint's default unit definition file.

andrewgsavage:
yes I will get to them at some point, don't be discouraged if I take a while

How are you finding these? Are you using QUDT?
Ideally we'd like to add other information like the reference to be accessible within pint, which the toml format should allow

PR #2219, opened 2025-09-15
Fix alias for decibelmicrowatt

acottuli:
dBu is a voltage level not a power level.

0 dBu is defined as the RMS voltage that would
dissipate 0 dBm (1 mW) in a 600 ohm load.

dBuW is an alias which is used for decibelmicrowatt.

References
- (dBu) https://en.wikipedia.org/wiki/Decibel#List_of_suffixes
- (dBuW) https://spectrumcompact.com/uploads/2024/05/Spectrum_Compact_v2_03-43_GHz__User_Manual_v43.pdf

### DIFF
--- a/pint/default_en.txt
+++ b/pint/default_en.txt
@@ -528,7 +528,7 @@ LMH = L / m**2 / h
 
 decibelwatt = watt; logbase: 10; logfactor: 10 = dBW
 decibelmilliwatt = 1e-3 watt; logbase: 10; logfactor: 10 = dBm
-decibelmicrowatt = 1e-6 watt; logbase: 10; logfactor: 10 = dBuW
+decibelmicrowatt = 1e-6 watt; logbase: 10; logfactor: 10 = dBuW = dBµW = dBμW  # NOTE: µ = U+00B5, μ = U+03BC
 
 decibel = 1 ; logbase: 10; logfactor: 10 = dB
 # bell = 1 ; logbase: 10; logfactor: = B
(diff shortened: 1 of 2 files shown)


andrewgsavage:
I am confused trying to get my head around this. On the wiki page under
https://en.wikipedia.org/wiki/Decibel#List_of_suffixes_in_alphabetical_order
> dBu or dBv 
RMS voltage relative to sqrt(0.6) V ...

there's no dBuW in that list

acottuli:
> I am confused trying to get my head around this. On the wiki page under https://en.wikipedia.org/wiki/Decibel#List_of_suffixes_in_alphabetical_order
> 
> > dBu or dBv
> > RMS voltage relative to sqrt(0.6) V ...
> 
> there's no dBuW in that list

Yes I know. It's because dBuW is not a commonly used alias (like dBW and dBm) and so is not explicitly listed in [ITU-R V.574-5](https://www.itu.int/dms_pubrec/itu-r/rec/v/R-REC-V.574-5-201508-I!!PDF-E.pdf). That's why I labelled the first reference as (dBu) and the second reference as (dBuW).

The first reference provides a definition for dBu (which should never be used as an alias for decibel microwatt) and the second reference provides an example where dBuW is used as an alias for decibel microwatt.

If you google `"dBuW" decibel microwatt` you'll find many more references where dBuW / dBµW / dBμW is used as an alias for decibel microwatt.

acottuli:
For reference, another relevant standard for dealing with logarithmic quantities will be [IEC 80000-15 Quantities and units – Part 15: Logarithmic and related quantities](https://www.iso.org/standard/76692.html), however it's currently under development and I couldn't find a committee draft available online.

PR #2231, opened 2025-09-17
Allow pint.registry.Quantity to be used as a generic class

### LINKED ISSUE #1770: Typing in 0.22
This issue is to address typing related topics that we should for 0.22. The objective is not to fix every typing aspects but those exposed to the users

zschaller27:
- [x] Closes #1770

### DIFF
--- a/pint/facets/context/registry.py
+++ b/pint/facets/context/registry.py
@@ -427,7 +427,7 @@ class GenericContextRegistry(
 
 
 class ContextRegistry(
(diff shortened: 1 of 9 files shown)


zschaller27:
> Any idea why this doesn't work
> 
> ```python
> ureg = UnitRegistry()
> length = ureg.Quantity[int]
> ```
> 
> TypeError: <class 'pint.Quantity'> is not a generic class

I'd forgotten to change the `UnitRegistry` class signature to make use of the `facets.MagnitudeT` typevar. I've pushed that small change so this should work now.

Issue #2075, opened 2024-11-08
Type error when creating Quantity in 0.24.4


brearleyt:
After upgrading to `0.24.4`, I get a type error for this example:

Code:
```py
from pint import UnitRegistry

ureg = UnitRegistry()

distance = 24.0 * ureg.meter
```

Error:
```
example.py:5:1 - error: Type of "distance" is partially unknown
  Type of "distance" is "Unit | Unknown | PlainQuantity[Any] | _NotImplementedType" (reportUnknownVariableType)
```

I am using pyright `1.1.388` on strict mode.

When using `0.24.3`, pyright was able to infer the type of `distance` as `_NotImplementedType | Quantity` and did not raise any errors.

reteps:
I think this is because `__mul__` is not typed.

PR #2243, opened 2025-10-26
typing for unit __mul__

### LINKED ISSUE #1947: mypy error: defining a Quantity as multiplication of list and unit raises an error

andrewgsavage:
- [ ] Closes #1947 #2075


andrewgsavage:
```python3
    @overload
    def __mul__(self, other) -> PlainQuantity: ...
```
gives the error 
`pint\testsuite\typing_tests\quantity.py:5: error: Incompatible types in assignment (expression has type "PlainQuantity[Any]", variable has type "Quantity[list[float]]")  [assignment]`

whereas any of the following run ok
```python3
    @overload
    def __mul__(self, other) -> QuantityT: ...

    @overload
    def __mul__(self, other) -> QuantityT[MagnitudeT]: ...
```

andrewgsavage:
Locally I've added 
```python3

    @overload
    def __rmul__(self, other: PlainUnit) -> PlainUnit: ...  # type: ignore[misc]

    @overload
    def __rmul__(self, other) -> QuantityT[MagnitudeT]: ...

    def __rmul__(self, other):
        return self.__mul__(other)
```
however this doesn't allow `b = 5 * ureg.meter` to be recognised as a `Quantity` with mypy
gpt5:
> Because mypy types binary * from the left operand, it picks the list __mul__ signature (which is typed to return list[T]) and never models the runtime fallback to the unit's __rmul__. At runtime Python does call PlainUnit.rmul and returns a Quantity, but mypy's static rules conclude the expression has type list[int], so the assignment to Quantity[...] is flagged.

PR #2291, opened 2026-03-26
Fix regression #2290

### LINKED ISSUE #2290: Regression: Custom string formats broken after 0.25.3 update
After updating pint from 0.25.2 to 0.25.3 my custom unit formatter stopped working. Here is a minimized example to reproduce:
```python
import pint

@pint.register_unit_format("X")
def custom_formatter(unit, registry, **options):
    return pint.formatter(
        unit.items(),
        {},
        power_fmt="{} ^ {}",
        **options,
    )

area = pint.Quantity("1.23456789 m^2")
print(f"{area:#~.4gX}")
print(f"{area:.4gX}")
print(f"{area:#~X}")
print(f"{area:#~.4gX}")
```
Expected output:
```
1.235 m ^ 2
1.235 meter ^ 2
1.23456789 m ^ 2
1.235 m ^ 2
```
Actual output:
```
1.235 m ** 2
1.235 meter ** 2
1.23456789 m ^ 2
1.235 m ^ 2
```

The issue is that the custom formatter is not recognized when used in combination with certain modifiers. Here, the issue seems to be the `.4g` modifier used for rounding the number to 4 significant digits. After the custom formatter is used once without these problematic modifiers, subsequent usages of the custom formatter work as expected.

eltos:
Reverts a small change introduced in #2268: The hardcoded list of modifiers is too restrictive and does not capture magnitude modifiers.

- [x] Closes #2290
  --> Already documented in user/formatting under "Magnitude modifiers"

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@ Pint Changelog
 
 - Add beats per minute (`beats_per_minute`, `bpm`) (#2286)
 - Bump minimum numpy version to 2.0.0
+- Fix regression when using magnitude format specs in string formats (#2291)
 
 0.25.3 (2026-03-19)
 -----------------
--- a/pint/delegates/formatter/full.py
+++ b/pint/delegates/formatter/full.py
@@ -104,9 +104,10 @@ class FullFormatter(BaseFormatter):
             if k in spec:
                 return v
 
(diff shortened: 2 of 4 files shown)


mutricyl:
Thanks for fixing this @eltos, sorry for this regression.

PR #2293, opened 2026-04-01
Updating `repr` for various objects

### LINKED ISSUE #1495: Quantity repr
We should use magnitude repr for Quantity repr as follow

```python

<Quantity(array(2.), 'dimensionless')>
<Quantity(array(2., dtype=float32), 'dimensionless')>
<Quantity(array([2., 3., 4., ..., 2., 3., 4.]), 'dimensionless')>
<Quantity(dask.array<array, shape=(3,), dtype=float64, chunksize=(3,), chunktype=numpy.ndarray>, 'dimensionless')>
<Quantity(2.932515646146465, 'dimensionless')>
<Quantity(Decimal('2.9325156461464651456456545649'), 'dimensionless')>
```
using `units=` might be a good idea too.

_Originally posted by @jules-ch in https://github.com/hgrecco/pint/issues/1342#issuecomment-1057058129_

leakec:
- [x] Closes #1495

This updates the `repr` values to use `repr(magnitude)` rather than printing the `magnitude` with a formatter. This also means that `repr` values are valid Python statements, so `eval(repr(...))` works; I've added unit tests that do this to confirm. These changes also make it so they work with things like `pybind11-stubgen`. See #1495 for more details.

### DIFF
--- a/pint/testsuite/test_quantity.py
+++ b/pint/testsuite/test_quantity.py
@@ -138,6 +138,10 @@ class TestQuantity(QuantityTestCase):
         assert repr(x_arr) == 'Quantity(array([1., 2., 3.]), "meter ** 2")'
 
         # Imports needed for the eval
+        from numpy import array  # noqa: F401
+
+        from pint import Quantity  # noqa: F401
+
         assert eval(repr(x)) == x
         assert np.all(eval(repr(x_arr)) == x_arr)
 
--- a/pint/testsuite/test_quantity.py
+++ b/pint/testsuite/test_quantity.py
@@ -130,13 +130,10 @@ class TestQuantity(QuantityTestCase):
         assert self.Q_(1000, "millimeter") == self.Q_(1, "meter")
         assert self.Q_(1000, "millimeter/min") == self.Q_(1000 / 60, "millimeter/s")
 
-    @helpers.requires_numpy
     def test_quantity_repr(self):
         x = self.Q_(4.2, UnitsContainer(meter=1))
-        x_arr = self.Q_(np.array([1.0, 2.0, 3.0]), UnitsContainer(meter=2))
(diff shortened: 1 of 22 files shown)


pint/testsuite/test_quantity.py
```diff
@@ -132,8 +132,18 @@ def test_quantity_comparison_convert(self):
 
     def test_quantity_repr(self):
         x = self.Q_(4.2, UnitsContainer(meter=1))
+        x_arr = self.Q_(np.array([1.0, 2.0, 3.0]), UnitsContainer(meter=2))
```
andrewgsavage: can you move this and the numpy bits below to a new test function that uses `@helpers.requires_numpy`?

andrewgsavage:
the doctests will need changing too

pint/testsuite/test_quantity.py
leakec: Okay, I believe I've got this one all sorted.

leakec:
> the doctests will need changing too

@andrewgsavage  What do you mean by the `doctests`? Can you please point me to the file(s) where these are located and I'll update them.

EDIT: Nevermind, found them in various rst files. Will update these.

EDIT 2: I only found one `rst` file that uses repr when doing a `grep` over all `rst` files for `repr`. Please let me know if there is anywhere else this might be.

andrewgsavage:
yea you can see the failures in the CI doc build, https://github.com/hgrecco/pint/actions/runs/23922418554/job/69772924257?pr=2293

leakec:
@andrewgsavage Okay, I think I've got them all now.

andrewgsavage:
nice thanks

leakec:
> nice thanks

No problem. Thanks for the review and the help merging!

PR #2302, opened 2026-04-26
Improve type annotations for `PlainQuantity`

RBerga06:
Closes #2281.

If `PlainQuantity`'s methods return `PlainQuantity`, subclass information is lost. To improve the status quo, two options are available:
1. When the return type matches the instance type (i.e. `PlainQuantity[MagnitudeT]`), just return `Self`;
2. When this is not possible, the only route I see is to copy method signatures in the relevant subclasses (e.g. `Quantity`).

For now, I have chosen to restrict myself to type methods that allow for option 1 to be used.

---
_**Edit**_: Other improvements:
- Marked the `BabelKwds` `TypedDict` non-total, because those keyword arguments are optional (`TypedDict`s are total by default, see the examples [in the spec](https://typing.python.org/en/latest/spec/typeddict.html))
- Changed annotations for the `DimensionalityError.dim*` attributes because inside of `PlainQuantity._iadd_sub(...)` they were assigned a `UnitsContainer` (i.e. `PlainQuantity.dimensionality`) instead of a `str`

### DIFF
--- a/pint/delegates/formatter/_compound_unit_helpers.py
+++ b/pint/delegates/formatter/_compound_unit_helpers.py
@@ -46,7 +46,7 @@ SortFunc: TypeAlias = Callable[
 ]
 
 
-class BabelKwds(TypedDict):
+class BabelKwds(TypedDict, total=False):
     """Babel related keywords used in formatters."""
 
     use_plural: bool
--- a/pint/errors.py
+++ b/pint/errors.py
@@ -12,6 +12,9 @@ from __future__ import annotations
 
 import typing as ty
 
+if ty.TYPE_CHECKING:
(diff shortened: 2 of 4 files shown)


andrewgsavage:
thanks. could you open a PR with 

> When this is not possible, the only route I see is to copy method signatures in the relevant subclasses (e.g. Quantity).

It would be good to see what you mean

RBerga06:
@andrewgsavage Thank you!
> could you open a PR with (...) It would be good to see what you mean

Yes, I have just opened #2303. At the time of writing it's a draft PR because I'd like to annotate more methods (and I still have to check that everything works as intended).

PR #2303, opened 2026-04-30
Improve type annotations for `Quantity`'s arithmetic methods

### LINKED ISSUE #2257: PlainQuantity from_list and from_sequence typing is wrong
`from_list` and `from_sequence` are explicitly meant to create a Quantity wrapping an ndarray.

Currently the typing indicate it return a Quantity whose magnitude is of the same type as the type as the input Quantities (see https://github.com/hgrecco/pint/blob/master/pint/facets/plain/quantity.py#L361) which is wrong.

RBerga06:
Follow-up of PR #2302.

Here I've duplicated method signatures to propagate subclass information through (some) arithmetic methods that do not simply return `Self`.

This is a suboptimal solution, because if anyone wants to define their own `PlainQuantity` subclass, they have to essentially copy-paste the methods definitions in `Quantity`'s body in order to achieve the same type checking behaviour. Nevertheless, I don't think the current tools in the Python type system allow for more ergonomic solutions.

---

**Edit**: Additional changes:
- Added `optype` as an optional dependency
- Switched to the new (PEP 695) syntax for type variables

---

- [x] Closes #2257, #2258

### DIFF
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -11,6 +11,8 @@ need.
 :copyright: 2022 by Pint Authors, see AUTHORS for more details.
 :license: BSD, see LICENSE for more details.
 """
+# pyright: reportNoOverloadImplementation=none
+# pyright: reportInvalidTypeArguments=information
 
 from __future__ import annotations
 
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -14,9 +14,7 @@ need.
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Generic, SupportsAbs, TypeVar, overload
-
-from typing_extensions import override
+from typing import TYPE_CHECKING, SupportsAbs, overload, override
 
 from . import facets, registry_helpers
 from ._typing import Magnitude
@@ -24,9 +22,11 @@ from .compat import TypeAlias
 from .util import logger, pi_theorem
 
 if TYPE_CHECKING:
-    import optype as opt
+    from ._typing import Magnitude
+    from ._typing import Quantity as _Quantity
+    from ._typing import Unit as _Unit
 
-T = TypeVar("T", bound=Magnitude)
+    import optype as opt
 
 # To build the Quantity and Unit classes
 # we follow the UnitRegistry bases
@@ -55,7 +55,7 @@ class Quantity[MagnitudeT: Magnitude](
         ) -> Quantity[T]: ...
 
         @override
-        def __abs__(self: Quantity[SupportsAbs[T]]) -> Quantity[T]: ...
+        def __abs__[T: Magnitude](self: Quantity[SupportsAbs[T]]) -> Quantity[T]: ...
 
 
 class Unit(
@@ -70,15 +70,14 @@ class Unit(
     pass
 
 
-class GenericUnitRegistry(
-    Generic[facets.QuantityT, facets.UnitT],
-    facets.GenericSystemRegistry[facets.QuantityT, facets.UnitT],
-    facets.GenericContextRegistry[facets.QuantityT, facets.UnitT],
-    facets.GenericDaskRegistry[facets.QuantityT, facets.UnitT],
-    facets.GenericNumpyRegistry[facets.QuantityT, facets.UnitT],
-    facets.GenericMeasurementRegistry[facets.QuantityT, facets.UnitT],
-    facets.GenericNonMultiplicativeRegistry[facets.QuantityT, facets.UnitT],
-    facets.GenericPlainRegistry[facets.QuantityT, facets.UnitT],
+class GenericUnitRegistry[QuantityT: _Quantity, UnitT: _Unit](
+    facets.GenericSystemRegistry[QuantityT, UnitT],
+    facets.GenericContextRegistry[QuantityT, UnitT],
+    facets.GenericDaskRegistry[QuantityT, UnitT],
+    facets.GenericNumpyRegistry[QuantityT, UnitT],
+    facets.GenericMeasurementRegistry[QuantityT, UnitT],
+    facets.GenericNonMultiplicativeRegistry[QuantityT, UnitT],
+    facets.GenericPlainRegistry[QuantityT, UnitT],
 ):
     pass
 
@@ -203,7 +202,7 @@ class UnitRegistry[MagnitudeT: Magnitude](
     check = registry_helpers.check
 
 
-class LazyRegistry(Generic[facets.QuantityT, facets.UnitT]):
+class LazyRegistry[QuantityT: Quantity, UnitT: Unit]:
     def __init__(self, args=None, kwargs=None):
         self.__dict__["params"] = args or (), kwargs or {}
 
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -17,7 +17,16 @@ need.
 from __future__ import annotations
 
 from collections.abc import Iterator, Sequence
-from typing import TYPE_CHECKING, Any, Self, TypeAlias, overload, override
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Generic,
+    Self,
+    TypeAlias,
+    TypeVar,
+    overload,
+    override,
+)
 
 from . import facets, registry_helpers
 from .util import logger, pi_theorem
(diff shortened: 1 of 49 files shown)


andrewgsavage:
Have you seen https://jorenham.github.io/optype/ ? It may be better to use here

andrewgsavage:
We can also support 3.12+ since Apr 24, 2026
https://numpy.org/neps/nep-0029-deprecation_policy.html
so can use the newer typing syntax

RBerga06:
Thanks @andrewgsavage for taking a look at this!
> Have you seen [jorenham.github.io/optype](https://jorenham.github.io/optype/) ? It may be better to use here

I did not know about this library, I'll have a look and see if it can help. 

**Edit**: Yes, it does look very useful, can I add it as a dependency?

> We can also support 3.12+ since Apr 24, 2026
> [numpy.org/neps/nep-0029-deprecation_policy.html](https://numpy.org/neps/nep-0029-deprecation_policy.html)
> so can use the newer typing syntax

Ah, I see. Should I submit a separate PR updating `pyproject.toml` then? I guess it would be a breaking change.

andrewgsavage:
> **Edit**: Yes, it does look very useful, can I add it as a dependency?

yea

> 
> > We can also support 3.12+ since Apr 24, 2026
> > [numpy.org/neps/nep-0029-deprecation_policy.html](https://numpy.org/neps/nep-0029-deprecation_policy.html)
> > so can use the newer typing syntax
> 
> Ah, I didn't notice numpy 2.0 was required, I've assumed we supported 3.11+ per the `requires-python` field (and the presence of the `py311` test environment) in `pyproject.toml`. Should I submit a separate PR updating `pyproject.toml` then?

numpy is optional. I haven't got round to updating min version to 3.12. that'll need changing CI settings so don't feel you need to submit a PR

RBerga06:
For now I've chosen to always import `optype` under `if TYPE_CHECKING`. This way it's always an optional dependency, unless when type checking, and it won't affect runtime performance.

RBerga06:
> that'll need changing CI settings so don't feel you need to submit a PR

I ended up needing to make some changes locally in order to run tools such as `ruff` and `pixi`. Since I've also updated some definitions with the new typing syntax, I figured it would've made sense to publish my changes as a PR anyway (#2304). Feel free to close or edit it, if necessary.

This way, this PR won't be overloaded with syntax upgrades and can contain only the new annotations.

RBerga06:
I've resolved all conflicts with master (4bf26dd15f585de71a64ada7e7660539937714ef), but I still have to annotate the binary arithmetic methods (`__add__`, `__sub__`, etc.).

However, I don't understand why does `UnitRegistry` need to be generic over `MagnitudeT: Magnitude`. Isn't the same `UnitRegistry` used for all associated quantities?

andrewgsavage:
> However, I don't understand why does `UnitRegistry` need to be generic over `MagnitudeT: Magnitude`. Isn't the same `UnitRegistry` used for all associated quantities?

What do you mean by this? It is generic over Quantity and Unit
```python
class GenericPlainRegistry[QuantityT: PlainQuantity, UnitT: PlainUnit](
    metaclass=RegistryMeta
):```

**Comment by RBerga06** [comment]
> What do you mean by this? It is generic over `Quantity` and `Unit`

Ah yes, sorry, that's right. What I meant is, we technically wouldn't need unit registries to be generic over the complete (fully specialized) quantity type, because with the same registry one can construct different quantities with different magnitude types (e.g. `ureg.Quantity(32, "m") -> Quantity[int]`, `ureg.Quantity(np.array([3.2, 3.4, 3.5])) -> Quantity[np.ndarray[...]]`, etc.).

Because of the `ureg.Quantity()` constructor, what we'd *actually* need to be generic over is a higher kinded type `QuantityT[~MagnitudeT: Magnitude]: PlainQuantity[MagnitudeT]` (`~` meaning invariant). Sadly, we don't have HKTs in Python yet (https://github.com/python/typing/issues/548), so that doesn't look feasible for now.

The problem with the current approach is that the type aliases `GenericUnitRegistry[QuantityT, UnitT].Quantity = QuantityT` are not generic, which means that, for example, `SystemRegistry.Quantity = SystemRegistry[SystemQuantity[Any], SystemUnit].Quantity = SystemQuantity[Any]` is considered (at least by pyright) as already specialized, which triggers a fair amount of errors in the `class Quantity[MagnitudeT: Magnitude](facets.SystemRegistry.Quantity[MagnitudeT], ...)` definition.

**Edit**: In fact, pyright reports an error in the `test_quantity.py`:`TestGenericQuantityTyping.test_generic_type_use` test.

I think the simplest approach would be to switch to generic type aliases, like:
```python
# pint/facets/plain/registry.py
class GenericUnitRegistry[UnitT: PlainUnit]:
    type Quantity[MagnitudeT] = PlainQuantity[MagnitudeT]

    def parse_expression(...) -> Quantity[Any]
```
This unfortunately requires re-annotating `parse_expression` in subclasses (in a potential bright future, the [associated types proposal](https://discuss.python.org/t/associated-types/105401/6) I saw on Discourse would fix that).

Furthermore, these new type aliases don't support subclassing, so this won't really work. Fortunately, we just need an alias for the class itself, and that's achievable with `Quantity: TypeAlias = PlainQuantity`, which is very similar to what we're currently doing, but slightly different, in that we'd have to remove the `QuantityT` type variable.

I might try and experiment with this in a separate PR, but I think I'll prioritize the work on arithmetic methods.

andrewgsavage:
Let's discuss that in it's own issue as there's already a lot here! and people tend to read issues more

andrewgsavage:
Is there anything more you want to do on this?

RBerga06:
Yes, sorry, I've been quite busy this week. I'd like to annotate the remaining arithmetic methods (the binary ones, `__add__`, `__mul__` and so on) and also annotate some tests to verify that my approach to annotations works correctly.

RBerga06:
Ok so I found out my simple `optype.Can*`-based approach to correctly typing unary methods (`__bool__`, `__iter__`, `__int__`, `__float__`, `__complex__`, `__abs__`, `__round__`) ~does not currently work the way I was hoping for (only tested with pyright) because magnitude information is lost, the type checker is not smart enough to understand what I was doing.~ **Edit**: It does work with pyright, at least for library users, although it reports errors at method definitions (but I think we can live with that).

The core problem is that the `MagnitudeT` type variable is bound to a concrete type: yesterday I've opened [a thread on Discourse](https://discuss.python.org/t/allowing-specializing-a-bound-covariant-typevar-with-a-protocol-as-a-limited-form-of-intersection/107403) about this.

RBerga06:
There is still a typing error in `testsuite.test_quantity:TestQuantity.test_round`, but I think it's either a problem with optype (https://github.com/jorenham/optype/issues/596) or pyright, our `__round__` annotation is correct.

pint/registry.py
```diff
@@ -51,7 +352,16 @@ class Unit(
...
+    if TYPE_CHECKING:
+
+        @overload
+        def __mul__(self, other: Self) -> Self: ...
+        @overload
+        def __mul__[T: Magnitude](self, other: T) -> Quantity[T]: ...
+        @overload
+        def __mul__(self, other: str) -> Quantity[Any]: ...
+
+        __rmul__ = __mul__
```
andrewgsavage, line 364: does this need other ops like div?

andrewgsavage:
does this need doing for the other facets too?

pint/registry.py
RBerga06: In my mind, since this is already quite a big PR, typing `PlainUnit`/`Unit` methods too was out of scope. I only annotated this method because it's one of the most human-readable ways of creating `PlainQuantity`/`Quantity` (`3.14 * ureg.m` for example). However, if it's preferable to annotate also `__truediv__` or similar for consistency, I can look into it.

RBerga06:
> does this need doing for the other facets too?

To be precise, yes, it would, for exactly the same reason that `Quantity` needs those annotations. However, if I've read correctly [the documentation](https://pint.readthedocs.io/en/stable/advanced/custom-registry-class.html) (_please correct me if I'm mistaken_), one is supposed to use the facets they need by subclassing the corresponding `Quantity`s, in which case they'd still need to annotate their own class anyway.

> [!NOTE]
> In a perfect world, one would only subclass the right `Registry` classes and use the associated `ureg.Quantity` directly. However, I still haven't checked if this works well with typing. To be fair, I don't expected it to work: I think we'll first need to resolve #2306. I might experiment a bit in the next weeks and see how to make it work.

At the very least, `PlainQuantity` automatically provides to all subclasses the arithmetic methods, although in a suboptimal form (`-> PlainQuantity` instead of the right subclass) - that's already an improvement.

pint/registry.py
andrewgsavage: sure that makes sense, what you've done is fine

andrewgsavage:
Thanks for explaining. I've not seen any issues related to subclassing or creating custom classes so I wonder whether that functionality is used! Having the typing in for `Quantity` in registry.py is what matters most :D

RBerga06:
Thank you for the review!

PR #2314, opened 2026-06-03
fix: format_unit() with ^ flag for use with negative powers for standalone units

### LINKED ISSUE #2313: `format_unit()` with `^` flag should use negative powers for standalone units

Example:

- currently returns `1/s`

juhuebner:
- [x] Closes #2313


pint/testsuite/test_formatting.py
```diff
@@ -72,3 +72,183 @@ def format_custom(unit, registry, **options):
...
+
+        # Magnitude should be identical
+        assert f"{q:~^P}".split()[0] == "5.5"
+        # Only unit part differs between with/without '^'
+        assert f"{q:~P}" == "5.5 1/s"
+        assert f"{q:~^P}" == "5.5 s⁻¹"
+
+
+if __name__ == "__main__":
+    pytest.main([__file__, "-v"])
```
andrewgsavage, line 254: what's this for?

PR #2318, opened 2026-06-06
Fix offset (affine) unit conversion of Decimal magnitudes (#2305)

### LINKED ISSUE #2305: Scale Conversion with Decimal and complex Magnitudes fails
Hei there
I looked for an open issue like this but couldn't find any, so please forgive me if this is already documented anywhere else.

## Problem
As the title says, converting a `Quantity` that requires a scale conversion from one unit to another fails if the magnitude of the `Quantity` is a `Decimal` or `complex` type.

Example:

```python
from pint import Quantity
from decimal import Decimal

Quantity(Decimal(10), "degC").to("K")

# TypeError: unsupported operand type(s) for +: 'decimal.Decimal' and 'float'
```
This fails in `pint.facets.nonmultiplicative.definitions.OffsetConverter.to_reference` because `value` is of type `Decimal`, whereas `self.scale` and `self.offset` are of type `float`.

The same problem exists if you transform a `complex` type value or a `Quantity` where the magnitude is an `np.array` that holds `Decimal` or `complex` values.

This is an example for the `OffsetConverter` but it affects all `Converter` classes equally in their `to_reference` and `from_reference` functions.

The converters are covered by unittests but these unittests do not take the various input types for values into account. `pint.testsuite.test_converters.TestConverter.test_multiplicative_converter` could be changed like this to cover these cases:

```python
    def test_multiplicative_converter(self):
        c = ScaleConverter(20.0)
        assert c.is_multiplicative
        assert not c.is_logarithmic
        for possible_type in [complex, float, int, Decimal, Fraction]:
            val = possible_type(100)
            assert c.from_reference(c.to_reference(val)) == val
            assert c.to_reference(c.from_reference(val)) == val
```
This test will currently fail for the aforementioned input types.

## Possible fix
The naive way to fix this would be to check for the type of `value` on converting and adapting the type of the `scale` and `offset` values if necessary. 

For example:

```python
@dataclass(frozen=True)
class OffsetConverter(ScaleConverter):
    """An affine transformation."""

    def to_reference(self, value: Magnitude, inplace: bool = False) -> Magnitude:
            scale = Decimal(self.scale)
            offset = Decimal(self.offset)
        elif isinstance(value, Fraction):
            scale = Fraction(self.scale)
            offset = Fraction(self.offset)
        else:
            scale = self.scale
            offset = self.offset
        if inplace:
            value *= scale
            value += offset
        else:
            value = value * scale + offset
        return value
```

I'd be willing to provide a PR and adapt the unittests if this is the way forward. If there is a better way please let me know.

gaoflow:
Fixes #2305.

## Problem

Converting a `Quantity` with a `Decimal` magnitude across an affine (offset) unit raises `TypeError`:

```python
from pint import UnitRegistry
from decimal import Decimal

ureg = UnitRegistry()
ureg.Quantity(Decimal(10), "degC").to("K")
# TypeError: unsupported operand type(s) for +: 'decimal.Decimal' and 'float'
```

`OffsetConverter.to_reference` / `from_reference` combine the magnitude with `self.scale` and `self.offset` directly:

```python
value = value * self.scale + self.offset
```

The scale and offset are parsed as `float`, and `Decimal` does not support arithmetic with `float`, so the conversion fails (`*` for units with a non-unit scale like `degF`, `+`/`-` for the offset, in both directions).

Setting `non_int_type=Decimal` (as suggested in the issue) only moves the problem: scale/offset then become `Decimal`, so `Decimal` magnitudes work but plain `float` magnitudes fail instead.

## Fix

Coerce the scale/offset to a type compatible with the magnitude before the arithmetic (`_matching_scalar`): promote to `Decimal` for a `Decimal` magnitude, and demote a `Decimal` scale/offset to `float` for any other magnitude. Other numeric types (`float`, `complex`, `Fraction`, numpy arrays) already interoperate and are untouched, so this is a no-op for the common case. `Decimal` magnitudes now keep their type through the conversion, and the mirror `float`-in-a-`Decimal`-registry case works too.

## Tests

Added `test_issue2305` covering `degC`/`degF`/`K` conversions of `Decimal` magnitudes in both directions, the float-in-Decimal-registry case, and that float magnitudes are unaffected. Fails on `master` (TypeError), passes here.

`test_issues.py`, `test_converters.py`, `test_non_int.py`, `test_numpy.py`, `test_quantity.py` all pass (no regressions). `ruff check` and `ruff format` clean.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Pint Changelog
 0.26.0 (unreleased)
 -------------------
 
+- Fix offset (affine) unit conversion of `Decimal` magnitudes raising `TypeError` from mixing `Decimal` with the float scale/offset (e.g. `Quantity(Decimal(10), "degC").to("K")`), and the mirror case of a plain `float` magnitude in a `Decimal` registry (#2305)
 - Add `system` parameter to `to_base_units` and `ito_base_units` to allow converting to the base units of a specific unit system (#2297)
 - Add beats per minute (`beats_per_minute`, `bpm`) (#2286)
 - Drop support for Python 3.11 in favour of Python 3.14 (#2304)
--- a/pint/compat.py
+++ b/pint/compat.py
@@ -14,6 +14,7 @@ import math
 import sys
 from collections.abc import Callable, Iterable, Mapping
 from decimal import Decimal
+from fractions import Fraction
 from importlib import import_module
 from importlib.util import find_spec
 from numbers import Number
@@ -29,6 +30,27 @@ if sys.version_info >= (3, 13):
     from warnings import deprecated  # noqa
 else:
     from typing_extensions import deprecated  # noqa
+
+
+def coerce_scalar(value, scalar):
+    """Coerce a scalar (a conversion scale or offset) to be arithmetically
+    compatible with value (magnitude).
+
+    Decimal refuses arithmetic with float (both directions raise TypeError);
+    Fraction loses its type when combined with float. Promote the scalar to
+    match a Decimal/Fraction magnitude, and demote a Decimal scalar to float
+    when value is a plain float (the only case where float OP Decimal fails).
+    int OP Decimal works fine and is left unchanged.
+    """
+    if isinstance(value, Decimal):
+        if not isinstance(scalar, Decimal):
+            return Decimal(str(scalar))
+    elif isinstance(value, Fraction):
+        if not isinstance(scalar, Fraction):
+            return Fraction(str(scalar))
+    elif isinstance(value, float) and isinstance(scalar, Decimal):
+        return float(scalar)
+    return scalar
 
 
 def missing_dependency(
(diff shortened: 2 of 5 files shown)


andrewgsavage:
I saw normal this issue doesn't occur for plain units (m -> in) and found similar logic in plain.registry, so I've made both plain and nonmulti share the same logic. How does it look?

binggao1230:
Looks good to me. Pulling `coerce_scalar` into `compat` and sharing it between the plain `_convert` and `OffsetConverter` is cleaner than the duplicated branches, and it's a strict improvement over my version: a `Fraction` magnitude now stays exact instead of being pushed through `float` (e.g. `Fraction(100)` °C→°F comes back a `Fraction`), and the `Decimal` #2305 case still round-trips correctly. I pulled the branch locally — the #2305 regression plus the full `test_converters` and `test_issues` suites pass and ruff is clean. Happy to have it merged as-is.

andrewgsavage:
nice one thanks

Issue #2157, opened 2025-03-20
Support for scipy.sparse matrices


johannes-moegerle:
It would be nice if `pint` could handle sparse matrices from `scipy.sparse` similarly as it does with `numpy.ndarray`.

A first straightforward step would be to add `scipy.sparse.spmatrix` to the `Magnitude` Union. This would allow users to specify the type of the magnitude of a `PlainQuantity` as a sparse matrix, and `pint` would be able to infer the type of the magnitude of the `PlainQuantity` from the type hint.

Example:
```
from typing import Any, reveal_type

import numpy as np
from scipy.sparse import csr_matrix

import pint
from pint.facets.plain import PlainQuantity

ureg = pint.UnitRegistry()

a: PlainQuantity[float] = ureg.Quantity(1, "m")
reveal_type(a.magnitude)  # Type of "a.magnitude" is "float"

array: PlainQuantity[np.typing.NDArray[Any]] = ureg.Quantity(np.array([1, 2, 3]), "m")
reveal_type(array.magnitude)  # Type of "array.magnitude" is "ndarray[tuple[int, ...], dtype[Any]]"

sparse: PlainQuantity[csr_matrix] = ureg.Quantity(csr_matrix(np.array([1, 2, 3])), "m")
reveal_type(sparse.magnitude)  # Type of "sparse.magnitude" is "MagnitudeT@PlainQuantity"
# Fixed type hint: # Type of "sparse.magnitude" is "csr_matrix"

```

A minimal implementation would look like this:
https://github.com/johannes-moegerle/pint/commit/1b9e6abf8bf9c7a9fda0ab8871ecdc1e574b60d7

andrewgsavage:
that would be good to add, would you like to submit a PR? Is there a way to allow a package to add to the magnitude union?

spanag:
My use case (sparse matrix `A` @ numpy vector `v`) fails at this line because it [wrongly]( https://github.com/scipy/scipy/issues/15547 ) delegates to `np.matmul`:
https://github.com/hgrecco/pint/blob/0.25.3/pint/facets/plain/quantity.py#L1048  
But `A@v` does work perfectly as:
```python
A._mul_div(v, operator.matmul, operator.mul)
```
Could the offendlng line be replaced with the above? It uses `_mul_div` like the other operators do.

andrewgsavage:
feel free to submit a PR

Issue #859, opened 2019-08-30
Missing `__matmul__` @ operator. 


agoose77:
I'm not too sure on how pint works, but I imagine we can directly delegate to `np.matmul(self, other)`?

jthielen:
Oops, I forgot to list this as closed by #905.

PR #2319, opened 2026-06-09
Support matmul for non numpy matrices

spanag:
Replace np.matmul with common _mul_div, using matmul for the numbers and mul for the units.  
Allows eg scipy.sparse matrix @ vector.

- [X] Related: #859, #2157
- [X] Executed `pre-commit run --all-files` or `pixi run lint --all-files` with no errors
- [ ] The change is fully covered by automated unit tests  - *Should it be tested with eg a sparse matrix?*
- [ ] ~~Documented in docs/ as appropriate~~ - *not a new feature really*
- [ ] ~~Added an entry to the CHANGES file~~ - *not a new feature plus #859 was not mentioned to begin with*


andrewgsavage:
nice thanks

Issue #1097, opened 2020-05-12
unit redefinition in a context breaks transform in another context (again)


crusaderky:
Reopen of #1062, which I thought I had fixed it in #1064.
I just realised that my PR only moved the problem further: before #1064 the POC was failing at the second iteration; now it fails on the third one.

```python
ureg = pint.UnitRegistry()
for i in range(3):
    print(f"run {i}")
    ctx = pint.Context.from_lines(["@context _", "cal = 4 J"])
    with ureg.context("sp", ctx):
        u = ureg.Unit("nm")
        q = ureg.Quantity(1, u)
        q.to('J')
```
Output:
```
run 0
run 1
run 2
KeyError: 'nanometer'
```

PR #2321, opened 2026-06-12
Fix parse_units cache never hit for symbol/prefixed spellings (#2320)

### LINKED ISSUE #2320: parse_units cache is never hit for symbol/prefixed spellings (kg, ms, kWh) — `input_string in self._units` guard
While profiling an application that calls `ureg.kg`-style attribute access in hot loops, we found that pint's own parse cache (`registry._cache.parse_unit`) is populated for every spelling but **permanently unreachable for symbol forms and prefixed forms** — i.e. the spellings most users write (`kg`, `ms`, `kWh`, `GB`…).

### Cause

`PlainRegistry._parse_units_as_container` guards the cache hit with:

```python
cache = self._cache.parse_unit
# Issue #1097: it is possible, when a unit was defined while a different context
# was active, that the unit is in self._cache.parse_unit but not in self._units.
# If this is the case, force self._units to be repopulated.
if as_delta and input_string in cache and input_string in self._units:
    return cache[input_string]
```

`self._units` contains canonical names and registered aliases, but not symbol forms (`kg` is the symbol of `kilogram`) nor prefixed forms (`kWh`, `ms`). So for those strings the guard fails on every call and the full parse pipeline (`ParserHelper.from_string` → `parse_unit_name` → `_yield_unit_triplets` → `_dedup_candidates`) runs again — the cache entry exists but is never read.

### Repro (pint 0.25.3, Python 3.13)

```python
from time import perf_counter
import pint

ureg = pint.UnitRegistry()

for name in ("kilogram", "kg", "kilowatt_hour", "kWh", "millisecond", "ms"):
    ureg.parse_units(name)  # prime the cache
    s = perf_counter()
    for _ in range(2000):
        ureg.parse_units(name)
    dt = (perf_counter() - s) / 2000 * 1e6
    print(f"{name:14s} in cache: {name in ureg._cache.parse_unit!s:5s} "
          f"in _units: {name in ureg._units!s:5s} per call: {dt:5.1f} us")
```

```
kilogram       in cache: True  in _units: True  per call:   0.7 us
kg             in cache: True  in _units: False per call:  14.3 us
kilowatt_hour  in cache: True  in _units: True  per call:   0.7 us
kWh            in cache: True  in _units: False per call:  14.5 us
millisecond    in cache: True  in _units: True  per call:   0.7 us
ms             in cache: True  in _units: False per call:  14.9 us
```

A ~20× penalty for the symbol spelling of the same unit. Since `ureg.<attr>` (`__getattr__` → `Unit(item)` → `parse_units`) goes through the same path, every `ureg.kg` attribute access pays the full parse. In our application (hourly-timeseries carbon modeling, thousands of `u.kg` accesses per render) this guard alone accounted for ~25% of a user-facing operation before we worked around it by caching resolved Units app-side.

### Suggested fix

Keep the #1097 staleness protection but validate the **cached result** instead of the raw input string, e.g.:

```python
if as_delta and input_string in cache:
    cached = cache[input_string]
    if all(name in self._units for name in cached):
        return cached
```

The constituent names of a cached `UnitsContainer` are canonical, so they *are* `self._units` keys whenever the definitions are still live; a unit removed/invalidated after a context switch still fails the check and re-parses, preserving the #1097 behavior. Happy to turn this into a PR if the approach looks right to you.

gaoflow:
Fixes #2320.

### Problem

`PlainRegistry._parse_units_as_container` populates `self._cache.parse_unit` for every spelling, but the cache-hit guard required the raw input string to also be a `self._units` key:

```python
if as_delta and input_string in cache and input_string in self._units:
    return cache[input_string]
```

`self._units` holds canonical names and registered aliases, but **not** symbol forms (`kg` is the symbol of `kilogram`) nor prefixed forms (`kWh`, `ms`, `GB`). So for exactly the spellings most users write, the guard fails on every call and the full parse pipeline re-runs even though a cache entry exists. Since `ureg.kg` attribute access goes through the same path, every such access pays it. Reporter measured a ~20x penalty (≈0.7 µs cached vs ≈15 µs for `kg`).

### Fix

Validate the **cached result** instead of the raw input string. The cached `UnitsContainer` always holds canonical names, which are `self._units` keys exactly while the definitions are live, so this keeps the issue #1097 staleness protection (a unit removed after a context switch leaves its canonical name absent from `self._units`, so the cache correctly falls through and re-parses).

While doing so I found the cache was also not case-sensitivity aware: the key was the raw input only, and the old `input_string in self._units` guard had been incidentally masking that (a mis-cased spelling like `"Meter"` is never a `self._units` key). To keep `case_sensitive=True/False` correct, the cache is now keyed on `(input_string, case_sensitive)`.

### Verification

`kg`, `kWh`, `ms`, and `degC**2` now hit the cache (0 re-parses after priming, confirmed by spying on `ParserHelper.from_string`), dropping to ≈0.8 µs. New regression test `test_issue2320` covers both the cache-hit and the case-sensitivity behavior; it fails on `master` (`kg re-parsed instead of hitting the cache`) and passes here. Existing `test_issues`, `test_unit`, `test_contexts`, `test_quantity`, and `test_application_registry` suites stay green.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,7 @@ Pint Changelog
 - Bump minimum numpy version to 2.0.0
 - Fix `Unit.__eq__` when comparing with strings using unit aliases (e.g. `Unit('metre') == 'meter'`) (#634)
 - Fix regression when using magnitude format specs in string formats (#2291)
+- Fix `parse_units` cache never being hit for symbol and prefixed spellings (e.g. `kg`, `kWh`, `ms`), making those very common lookups re-parse on every call (#2320)
 - Updated repr for various objects so that eval(repr(...))) works (#1495)
 
 0.25.3 (2026-03-19)
--- a/pint/facets/plain/registry.py
+++ b/pint/facets/plain/registry.py
@@ -127,7 +127,9 @@ class RegistryCache:
         self.dimensionality: dict[UnitsContainer, UnitsContainer] = {}
 
         #: Cache the unit name associated to user input. ('mV' -> 'millivolt')
-        self.parse_unit: dict[str, UnitsContainer] = {}
+        #: Keyed on (input string, case_sensitive) because the same spelling can resolve
+        #: differently under each mode (e.g. "Meter" is undefined when case-sensitive).
+        self.parse_unit: dict[tuple[str, bool], UnitsContainer] = {}
 
         self.conversion_factor: dict[
             tuple[UnitsContainer, UnitsContainer], Scalar | DimensionalityError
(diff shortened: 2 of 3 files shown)


binggao1230:
Closing in favor of @VincentVillet, who reported #2320 and offered to turn it into a PR. I jumped in before noticing that — apologies. I left a note on the issue with one extra case (the cache needs to be keyed on `(input, case_sensitive)`) that may be useful for your PR.

PR #2322, opened 2026-06-13
Raise DefinitionSyntaxError instead of AssertionError for an operator-only expression (#2124)

### LINKED ISSUE #2124: Calling ureg('-') causes an AssertionError
I'm reading in text files and trying to identify units in them. I just randomly discovered that calling `ureg('-')` throws an `AssertionError` rather than a `pint.errors.UndefinedUnitError`. It appears to do the same thing for other symbols, like `ureg('*')`.

I could understand maybe throwing an error _other_ than an `UndefinedUnitError`, but an `AssertionError` is clearly a bad choice here.

vineethsaivs:
Fixes #2124.

Parsing an expression that is only an operator raised a bare `AssertionError` rather than a pint error:

```python
import pint
ureg = pint.UnitRegistry()
ureg('-')   # before: AssertionError()
ureg('*')   # before: AssertionError()
```

In `pint_eval._build_eval_tree`, an operator with no operand leaves `result is None`, which tripped `assert result is not None`. The surrounding malformed-token paths already raise `DefinitionSyntaxError`, so this does the same:

```python
ureg('-')   # after: pint.errors.DefinitionSyntaxError: missing operand
```

Valid expressions are unchanged (`ureg('3 m / s')`, `ureg('m**2')`, ...), and an unknown unit name still raises `UndefinedUnitError` (`ureg('foobar')`).

Added `test_build_eval_tree_missing_operand_raises_definition_syntax_error` (parametrized over both tokenizers and several operators); it fails on `master` (AssertionError) and passes with this change. All 145 tests in `test_pint_eval.py` pass.

---
*Disclosure: prepared with AI assistance; reviewed by me and I can explain every line.*


andrewgsavage:
Any idea why the benchmarks are showing slower results for getitem and
parse_expression?

vineethsaivs:
I don't think this PR is actually responsible for those numbers. The only runtime change is a one-line guard in the parse error path: `assert result is not None` becomes `if result is None: raise DefinitionSyntaxError("missing operand")` at the point where `_build_eval_tree` closes recursion. That runs at most once per parse rather than per token, and an `if` costs about the same as the `assert` it replaces, so it shouldn't move `parse_expression`, and it never runs on the `getitem` path at all.

CodSpeed flagged the cause on this run itself: "Unknown Walltime execution environment detected" and "Different runtime environments detected ... compared across different runtime environments". The slowdown is also pretty uniform (around 25 to 28%) across benchmarks that exercise completely different code (`parse_expression`, `getitem`, `op2`, `finding_meter_getitem`), which is what you'd expect when BASE and HEAD land on different runners, not from a localized change in the eval-tree builder.

A re-run on a consistent runner should bring it back in line. Happy to push an empty commit to re-trigger, or to acknowledge it on CodSpeed, whichever you prefer.

PR #2324, opened 2026-06-17
fix: respect ^ flag in format_unit for negative powers (fixes #2313)

### LINKED ISSUE #2313: `format_unit()` with `^` flag should use negative powers for standalone units

gaoflow:
`FullFormatter.format_unit()` was not passing `as_ratio=False` when the `^` flag was present in the format spec, unlike `format_quantity()` which handles this correctly. This caused standalone units to always use ratio form (`1/s`) instead of negative powers (`s⁻¹`) even when `^` was explicitly requested.

## Before
```python
>>> format(ureg.Unit("second")**-1, "~^P")
'1/s'
```

## After
```python
>>> format(ureg.Unit("second")**-1, "~^P")
's⁻¹'
```

The fix mirrors the existing `^` flag handling in `format_quantity()`.


binggao1230:
CI triage note: the current red checks do not appear to be caused by this formatter change.

Observed failures:
- `docbuild` / `doctest`: fail while rendering docs because Graphviz executable `dot` is missing from PATH (`ExecutableNotFound: failed to execute PosixPath(dot)`).
- `test-linux (test-py312, numpy>=2.0.0,<2.1.0, babel==2.15 matplotlib==3.9.0)` and `test-windows (test-py313, numpy>=2.4.0)`: fail during pixi/uv environment resolution with a distribution cache lock timeout before tests run.
- CodSpeed reports a broad walltime regression while also warning about inconsistent/different runtime environments; the affected benchmarks are unrelated parse/getitem paths, not formatter paths touched by this PR.

Local verification of the touched behavior passes:
- `uv run --no-project --python 3.12 --with-editable . --with pytest --with pytest-subtests --with pytest-benchmark python -m pytest pint/testsuite/test_formatting.py -q` -> 35 passed
- `uv run --no-project --python 3.12 --with ruff ruff check pint/delegates/formatter/full.py pint/testsuite/test_formatting.py`
- `uv run --no-project --python 3.12 --with ruff ruff format --check pint/delegates/formatter/full.py pint/testsuite/test_formatting.py`
- `git diff --check upstream/master...HEAD`

PR #2326, opened 2026-06-25
Fix wraps raising errors with VAR_POSITIONAL (*args) parameters (#2254)

### LINKED ISSUE #2254: ureg.wraps does not work with var-positional arguments
I want to wrap the `math.gcd` function using `ureg.wraps` utility function (decorator). The wrapped function works only with one parameter. Called without parameters and with more than one parameter I get various behaviors / errors:

```python
>>> import pint
>>> import math
>>> gcd = ureg.wraps('=A', '=A')(math.gcd)
>>> gcd(1)
<Quantity(1, 'dimensionless')>
>>> gcd(ureg.Quantity(1))
<Quantity(1, 'dimensionless')>
>>> gcd(ureg.Quantity(1, 'm'))
<Quantity(1, 'meter')>
>>> gcd(ureg.Quantity(1), ureg.Quantity(2))
  File "pint/pint/registry_helpers.py", line 293, in wrapper
    result = func(*new_values, **new_kw)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'Quantity' object cannot be interpreted as an integer
>>> gcd()
  File "pint/pint/registry_helpers.py", line 289, in wrapper
                                         ^^^^^^^^^^
  File "pint/pint/registry_helpers.py", line 128, in _converter
    values.append(kw[param_name])
                  ~~^^^^^^^^^^^^
KeyError: 'integers'
>>> gcd(1, 2)
<Quantity(1, 'dimensionless')>
```

When I create a function with `*args` myself the same error occurs called without parameters and another error occurs with more than one pint quantity parameters:

```python
>>> import pint
>>> def f(*x: int) -> int:
...     return sum(x)
...
>>> f_wrapped = ureg.wraps('=A', '=A')(f)
>>> f_wrapped()
  File "pint/pint/registry_helpers.py", line 289, in wrapper
                                         ^^^^^^^^^^
  File "pint/pint/registry_helpers.py", line 128, in _converter
    values.append(kw[param_name])
                  ~~^^^^^^^^^^^^
KeyError: 'x'
>>> f_wrapped(1)
<Quantity(1, 'dimensionless')>
>>> f_wrapped(1, 2)
<Quantity(3, 'dimensionless')>
>>> f_wrapped(ureg.Quantity(1, 'm'))
<Quantity(1, 'meter')>
>>> f_wrapped(ureg.Quantity(1, 'm'), ureg.Quantity(2, 'm'))
  File "pint/pint/registry_helpers.py", line 293, in wrapper
    result = func(*new_values, **new_kw)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<stdin>", line 2, in f
  File "pint/pint/facets/plain/quantity.py", line 874, in __add__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pint/pint/facets/plain/quantity.py", line 101, in wrapped
    return f(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "pint/pint/facets/plain/quantity.py", line 750, in _add_sub
    raise DimensionalityError(self._units, "dimensionless")
```

I have python 3.12.3, pint 0.25rc2.dev28+g9325ea272 (master branch).

Labib-Bin-Salam:
## Problem

`ureg.wraps` fails in two distinct ways when wrapping a function whose parameter list contains `*args` (`VAR_POSITIONAL`):

**Calling with 0 arguments raises `KeyError`:**
```python
@ureg.wraps("=A", "=A")
def f(*values):
    return sum(values)

f()  # KeyError: 'values'
```
`_converter` iterated over `sig.parameters` and, for any index `i ≥ len(positional_values)`, attempted `values.append(kw[param_name])`.  A `VAR_POSITIONAL` parameter never appears in `kw`, so the lookup failed.

**Calling with multiple `Quantity` arguments raises `DimensionalityError`:**
```python
f(1 * ureg.meter, 200 * ureg.cm)  # DimensionalityError
```
`_parse_wrap_args` produced exactly one entry in `args_as_uc` (one spec per declared parameter). Only `values[0]` was processed by the unit-conversion passes; `values[1]` was left as a raw `Quantity` and forwarded to the function unchanged, causing `sum(1, Quantity(200, 'cm'))` to raise.

## Fix

Two targeted changes inside `_converter` (the closure returned by `_parse_wrap_args`):

1. **Skip `VAR_POSITIONAL`/`VAR_KEYWORD` in the kwargs-packing and kwargs-unpacking loops.** Variadic positional args are collected as a tuple in `values`, not as individual keyword args, so attempting `kw[param_name]` is always wrong for them.

2. **Expand the unit-spec tables for extra `*args` values.** When `len(values) > len(args_as_uc)` and the function has a `VAR_POSITIONAL` parameter, copy the unit spec for that parameter to every additional index. When the original spec was a reference definition (e.g. `"=A"`), the first `*args` value continues to define the reference unit and the extras become dependent on it, so they are all converted to a consistent unit before being forwarded.

## Tests

Regression test added in `pint/testsuite/test_issues.py::test_issue2254` covering:
- single `Quantity` arg (pre-existing behaviour preserved)
- two `Quantity` args of the same dimension but different units (new)
- plain scalars without units (new)
- fixed-unit spec with two args (new)

All existing `test_wraps` and `test_wrap_referencing` cases still pass.

Closes #2254

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Pint Changelog
 0.26.0 (unreleased)
 -------------------
 
+- Fix `wraps` raising `KeyError` when calling a `*args` function with 0 arguments, and `DimensionalityError` when calling it with more than one `Quantity` argument (#2254)
 - Fix offset (affine) unit conversion of `Decimal` magnitudes raising `TypeError` from mixing `Decimal` with the float scale/offset (e.g. `Quantity(Decimal(10), "degC").to("K")`), and the mirror case of a plain `float` magnitude in a `Decimal` registry (#2305)
 - Add `system` parameter to `to_base_units` and `ito_base_units` to allow converting to the base units of a specific unit system (#2297)
 - Add beats per minute (`beats_per_minute`, `bpm`) (#2286)
--- a/pint/registry_helpers.py
+++ b/pint/registry_helpers.py
@@ -120,34 +120,77 @@ def _parse_wrap_args(args, registry=None):
     def _converter(ureg, sig, values, kw, strict):
         len_initial_values = len(values)
 
-        # pack kwargs
+        # pack kwargs, skipping VAR_POSITIONAL/VAR_KEYWORD params since they
+        # collect positional/keyword args and don't appear in kw themselves
         for i, param_name in enumerate(sig.parameters):
-            if i >= len_initial_values:
+            param = sig.parameters[param_name]
+            if i >= len_initial_values and param.kind not in (
+                Parameter.VAR_POSITIONAL,
+                Parameter.VAR_KEYWORD,
+            ):
                 values.append(kw[param_name])
+
+        # For VAR_POSITIONAL parameters: expand unit specs to cover all actual
+        # positional args passed beyond the single spec defined for *args.
+        local_args_as_uc = args_as_uc
+        local_defs_ndx = defs_args_ndx
+        local_dep_ndx = dependent_args_ndx
+        local_unit_ndx = unit_args_ndx
+
+        if len(values) > len(args_as_uc):
+            var_pos_ndx = next(
+                (
+                    i
+                    for i, pname in enumerate(sig.parameters)
+                    if sig.parameters[pname].kind == Parameter.VAR_POSITIONAL
+                ),
+                None,
+            )
+            if var_pos_ndx is not None:
+                local_args_as_uc = list(args_as_uc)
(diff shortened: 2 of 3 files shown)


andrewgsavage:
does https://github.com/hgrecco/pint/pull/2336 fix this? plrease reopen if not

PR #2327, opened 2026-06-30
Fix `unit in registry` membership test for Unit objects

### LINKED ISSUE #2156: `Unit` in `UnitRegistry` gives error
If I create a unit from the unit registry, I would expect that `unit in ureg` returns `True`.

Instead, I get an error:

```python
import pint
ureg = pint.UnitRegistry()
unit = ureg.Unit('J')
unit in ureg  #  AttributeError: 'Unit' object has no attribute 'endswith'
```

I think this is unintended; I would expect the statement to return `True`.

Currently, to correctly do the check, I must convert to string:

```python
str(unit) in ureg # True
```

vineethsaivs:
## Summary

Checking unit membership with a `Unit` object raised an `AttributeError` instead of returning `True`:

```python
import pint
ureg = pint.UnitRegistry()
unit = ureg.Unit('J')
unit in ureg        # AttributeError: 'Unit' object has no attribute 'endswith'
str(unit) in ureg   # True  (the string path already worked)
```

Closes #2156.

## Root cause

`GenericPlainRegistry.__contains__` passes its argument straight to `__getattr__`, which assumes a string and eventually calls `unit_name.endswith(...)` during unit-name parsing. A `Unit` object has no `endswith`, so an uncaught `AttributeError` escapes instead of the method returning a `bool`.

## Fix

Convert the item to its string form before delegating, so a `Unit` (or anything with a sensible unit string) reuses the existing, already-correct string path. `str(str)` is the identity, so plain string input is unchanged.

```python
def __contains__(self, item: str) -> bool:
    """Support checking prefixed units with the `in` operator"""
    try:
        self.__getattr__(str(item))
        return True
    except UndefinedUnitError:
        return False
```

Verified behaviour after the fix:

| check | before | after |
|---|---|---|
| `ureg.Unit('J') in ureg` | AttributeError | `True` |
| `(ureg.Unit('m') / ureg.Unit('s')) in ureg` | AttributeError | `True` |
| `'joule' in ureg` | `True` | `True` |
| `'kJ' in ureg` (prefixed) | `True` | `True` |
| `'definitely_not_a_unit' in ureg` | `False` | `False` |

## Test

Added `test_issue2156` to `pint/testsuite/test_issues.py`. It fails before the change (`AttributeError: 'Unit' object has no attribute 'endswith'`) and passes after. The full `test_issues.py` suite still passes (81 passed) and `ruff` is clean.


andrewgsavage:
this should be an isinstance check, see https://github.com/hgrecco/pint/issues/2156

vineethsaivs:
Thanks, that is a fair point, and reworked in `87ded69`.

`str(item)` made anything stringifiable resolvable through `in`, which widens `__contains__` well past what #2156 asks for. Checking the type instead keeps it to the reported case:

```python
def __contains__(self, item: str | UnitT) -> bool:
    """Support checking prefixed units with the `in` operator"""
    if isinstance(item, self.Unit):
        item = str(item)
    try:
        self.__getattr__(item)
        return True
    except UndefinedUnitError:
        return False
```

```
Unit object in registry : True     # the #2156 case
string in registry      : True     # unchanged
prefixed string ("kJ")  : True     # unchanged
undefined string        : False    # unchanged
unrelated object        : raises AttributeError, as it did before
```

So strings behave exactly as before, a `Unit` now resolves, and nothing else is silently coerced.

I have left this closed rather than reopening it myself. Happy to reopen it or raise a fresh PR, whichever you prefer, and equally happy to drop it if you would rather fix #2156 elsewhere.

andrewgsavage:
yea that looks fine, either works

PR #2332, opened 2026-07-10
Fix to_compact() crashing on units whose name is ambiguous with a prefix (dtex)

### LINKED ISSUE #2246: dtex with .to_compact() fails
pint 0.25.1

```python
quant = pint.Quantity(100, 'dtex') #decitex, fiber unit.
quant.to_compact()
```
```bash
AssertionError                            Traceback (most recent call last)
Cell In[57], [line 1](vscode-notebook-cell:?execution_count=57&line=1)
----> [1](vscode-notebook-cell:?execution_count=57&line=1) dpm_ureg.Quantity(100,'dtex').to_compact()

File c:\Users\xxx\AppData\Local\miniconda3\envs\datamore\Lib\site-packages\pint\facets\plain\qto.py:137, in to_compact(quantity, unit)
    134 SI_bases = [item[1] for item in SI_prefixes_list]
    136 if unit is None:
--> [137](file:///C:/Users/xxx/AppData/Local/miniconda3/envs/datamore/Lib/site-packages/pint/facets/plain/qto.py:137)     unit = infer_base_unit(quantity, registry=quantity._REGISTRY)
    138 else:
    139     unit = infer_base_unit(quantity.__class__(1, unit), registry=quantity._REGISTRY)

File c:\Users\xxx\AppData\Local\miniconda3\envs\datamore\Lib\site-packages\pint\util.py:1097, in infer_base_unit(unit_like, registry)
   1095 for unit_name, power in original_units.items():
   1096     candidates = registry.parse_unit_name(unit_name)
-> [1097](file:///C:/Users/xxx/AppData/Local/miniconda3/envs/datamore/Lib/site-packages/pint/util.py:1097)     assert len(candidates) == 1
   1098     _, base_unit, _ = candidates[0]
   1099     d[base_unit] += power

AssertionError:
```

```python
pint.Quantity(100,'dtex').to('g/m').to_compact()
```

works fine.

vineethsaivs:
## Summary

`Quantity.to_compact()` crashes with a bare `AssertionError` on a unit whose name is ambiguous with a prefixed unit:

```python
import pint
ureg = pint.UnitRegistry()
ureg.Quantity(100, "dtex").to_compact()
# AssertionError  (pint/util.py, infer_base_unit)
```

`to_compact()` calls `infer_base_unit`, which loops over the unit names and does `candidates = registry.parse_unit_name(unit_name); assert len(candidates) == 1`. For `dtex` the parser returns **two** candidates: `('', 'dtex', '')` (dtex is itself a registered unit, `dtex = decitex`) and `('deci', 'tex', '')`. `_dedup_candidates` cannot collapse them (it only drops a no-prefix name equal to the literal `prefix + unit` concatenation, and `deci` + `tex` = `decitex` != `dtex`), so `len(candidates) == 2` and the assert fires on a perfectly valid unit.

The sibling `Quantity._get_unprefixed` (`pint/facets/plain/qto.py`) handles the exact same `parse_unit_name` result by just taking `candidates[0]` with no assert, and a `TODO` there notes the two functions should be merged. This makes the assert an inconsistent oversight.

## Fix

Drop the assert and take `candidates[0]`, matching the sibling. `Quantity(100, "dtex").to_compact()` now returns `100 dtex`, and the documented `to_compact` examples are unchanged.

Fixes #2246.

## Test

`pint/testsuite/test_infer_base_unit.py::test_infer_base_unit_ambiguous_prefixed_unit` asserts `Quantity(100, "dtex").to_compact()` no longer raises and that `infer_base_unit` resolves it. Raises `AssertionError` before the fix, passes after. Added a `CHANGES` entry.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Pint Changelog
 0.26.0 (unreleased)
 -------------------
 
+- Fix `Quantity.to_compact()` raising `AssertionError` for units whose name is ambiguous with a prefixed unit (e.g. `dtex`) (#2246)
 - Fix offset (affine) unit conversion of `Decimal` magnitudes raising `TypeError` from mixing `Decimal` with the float scale/offset (e.g. `Quantity(Decimal(10), "degC").to("K")`), and the mirror case of a plain `float` magnitude in a `Decimal` registry (#2305)
 - Add `system` parameter to `to_base_units` and `ito_base_units` to allow converting to the base units of a specific unit system (#2297)
 - Add beats per minute (`beats_per_minute`, `bpm`) (#2286)
--- a/pint/testsuite/test_infer_base_unit.py
+++ b/pint/testsuite/test_infer_base_unit.py
@@ -133,3 +133,13 @@ def test_volts(sess_registry):
     b = infer_base_unit(r)
     assert b == sess_registry.Quantity(1, "V").units
     helpers.assert_quantity_almost_equal(r, sess_registry.Quantity(1, "uV"))
+
+
+def test_infer_base_unit_ambiguous_prefixed_unit(sess_registry):
+    # "dtex" parses both as the registered unit dtex and as deci + tex, so
+    # parse_unit_name returns two candidates; infer_base_unit / to_compact must
+    # take the first (like Quantity._get_unprefixed) rather than assert and crash.
+    # See https://github.com/hgrecco/pint/issues/2246
+    q = sess_registry.Quantity(100, "dtex")
+    assert q.to_compact() == q
+    assert infer_base_unit(q) == sess_registry.Quantity(1, "dtex").units
--- a/pint/util.py
+++ b/pint/util.py
@@ -1086,7 +1086,10 @@ def infer_base_unit(
 
     for unit_name, power in original_units.items():
         candidates = registry.parse_unit_name(unit_name)
-        assert len(candidates) == 1
+        # A unit name can be parsed as more than one (prefix, unit, suffix)
+        # candidate when it collides with a prefixed unit (e.g. "dtex" is both
+        # the registered unit dtex and deci + tex); take the first, matching the
+        # sibling logic in Quantity._get_unprefixed (facets/plain/qto.py).
         _, base_unit, _ = candidates[0]
         d[base_unit] += power
 


andrewgsavage:
can you look at changing the definition file instead

vineethsaivs:
Fair, and it does work for the reported unit. I tried it, and it resolves `dtex` cleanly but not the general case, so here is what I found in case it is useful.

**`dtex` via the definition file: yes.** `deci-` already has the symbol `d-`, so `d` + `tex` resolves through the prefix system on its own. That makes line 706 redundant, and it is the thing creating the second candidate:

```
dtex candidates before : ((, dtex, ), (deci, tex, ))
dtex candidates after  : ((deci, tex, ),)
```

Removing `    dtex = decitex` from `default_en.txt` is behaviour-preserving:

```
100 dtex        = 100 decitex
to_compact()    = 10.0 tex
100 dtex == 100 decitex -> True
```

**But it is not only `dtex`.** I swept the whole registry for names that parse to more than one candidate: 38 of them do, and `_dedup_candidates` already collapses 36. Two survive and hit the bare assert in `infer_base_unit`:

```
multi-candidate names                          : 38
actually crash to_compact() with AssertionError: 2  -> [dtex, rads]
```

`rads` is the one the definition file cannot fix. It is not a prefix collision, it collides with the plural of `radian`:

```
rads candidates : ((, rads, ), (, radian, ))
```

`rads` is the radiation-absorbed-dose unit at line 383 (`rads = 0.01 * gray`) and both spellings are legitimate, so neither side can simply be dropped.

So the definition change fixes `dtex`, and `to_compact()` still asserts on `rads`. Happy to send the one-line `default_en.txt` change on its own if you want just that, and to leave `infer_base_unit` alone. If you would also rather it not fail with a bare `AssertionError` for genuinely ambiguous names like `rads`, I can raise that separately so the two are not entangled. Whichever you prefer.

andrewgsavage:
I didn't realise it would affect other units so thanks for looking into that.

I think this needs a bit of discussion over how best to deal with it in the issue. There's a few options:  this PR, giving a warning when there's multiple candidates, disallowing ambuigity in the unit file

vineethsaivs:
Agreed, moved it to #2246 with the numbers so the options can be compared there: https://github.com/hgrecco/pint/issues/2246#issuecomment-5085980335

The short version, in case it saves a click. Of 38 ambiguous names only 2 actually raise, `dtex` and `rads`. The definition-file route fixes `dtex` (the `deci-` prefix already carries `d-`, so the explicit alias is redundant) but cannot fix `rads`, because that one collides with the generated plural of `radian` rather than with a prefix, so there is no line to delete. That rules out the option that looks easiest, which seemed worth knowing before anyone picks a direction.

No objection to this staying closed. Happy to implement whichever of the three you land on.

PR #2333, opened 2026-07-12
Fix NumPy offset unit mul/div guards

### LINKED ISSUE #1335: using np.array circumvents ambiguity error for offset units
pint: 0.17
numpy: 1.20.3
python: 3.9.1
________
I observe the following:
```
ureg.autoconvert_offset_to_baseunit
```
\>>>False
```
myoven = (400 * ureg.K).to('degC')
```
\>>>126.85000000000002 degree_Celsius

```
2* myoven 
```
\>>>OffsetUnitCalculusError: Ambiguous operation with offset unit (degree_Celsius, ). See https://pint.readthedocs.io/en/latest/nonmult.html for guidance.
```
np.array(2)*myoven
```
\>>>253.70000000000005 degree_Celsius

Not sure if is a bug or intended, but it is somewhat inconsistent imho.

therealclvn:
Closes #1335.

## Summary

`np.array(2) * quantity_with_offset_unit` and related NumPy ufunc paths were bypassing the offset-unit checks that `Quantity` scalar arithmetic already applies. With `autoconvert_offset_to_baseunit=False`, scalar arithmetic raises `OffsetUnitCalculusError`, but the NumPy array protocol path stripped units, computed the magnitude, and then rebuilt a result with the offset unit.

This adds a small validation helper inside the NumPy output-unit path for `mul`/`div`, so these ufuncs reuse the same `_ok_for_muldiv` guard before wrapping the NumPy result.

## Tests

Added regression coverage for:

- `np.array(2) * temperature`
- `np.multiply(temperature, 2 * ureg.m)`
- `np.array(2) / temperature`

when `temperature` has an offset unit and `autoconvert_offset_to_baseunit=False`.

Verification run locally:

- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_numpy_func.py -q`
- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_numpy.py -q`
- `.venv-pint-test/bin/python -m ruff check pint/facets/numpy/numpy_func.py pint/testsuite/test_numpy_func.py`
- `.venv-pint-test/bin/python -m ruff format --check pint/facets/numpy/numpy_func.py pint/testsuite/test_numpy_func.py`

I also ran the full `pint/testsuite`; all tests passed except `test_diskcache.py::test_auto`, which failed because this sandbox cannot create `~/Library/Caches/pint`. Re-running that single test with a writable cache root passed:

- `env XDG_CACHE_HOME=/tmp/pint-cache .venv-pint-test/bin/python -m pytest pint/testsuite/test_diskcache.py::test_auto -q`


therealclvn:
FYI, the ReadTheDocs failure here appears unrelated to this patch.

I reproduced the docs install locally with Python 3.12 and found that the docs extra could resolve `sparse` through an old `numba`/`llvmlite` stack that does not install on Python 3.12. I opened #2334 for that dependency issue, and its ReadTheDocs build is now green.

PR #2334, opened 2026-07-12
Constrain docs NumPy dependency

therealclvn:
## Summary
- constrain the docs extra to `numpy<2.5`
- avoid resolving `sparse` through an old `numba`/`llvmlite` stack on Python 3.12

## Verification
- `env UV_CACHE_DIR=/tmp/uv-cache VIRTUAL_ENV=$PWD/.venv-rtd-check uv pip install --dry-run -e '.[docs]'` resolved `numpy==2.4.6`, `numba==0.66.0`, and `llvmlite==0.48.0`
- `env UV_CACHE_DIR=/tmp/uv-cache VIRTUAL_ENV=$PWD/.venv-rtd-check uv pip install -e '.[docs]'` completed successfully on Python 3.12.13
- `.venv-rtd-check/bin/python -c 'import tomllib; tomllib.load(open("pyproject.toml", "rb")); print("pyproject ok")'`

Note: a local Sphinx build progressed past dependency installation and notebook startup, but stopped because this machine does not have the Graphviz `dot` executable. ReadTheDocs installs Graphviz through `.readthedocs.yaml`.

PR #2335, opened 2026-07-12
Honor negative-power unit format flag

### LINKED ISSUE #2313: `format_unit()` with `^` flag should use negative powers for standalone units

therealclvn:
## Summary
- pass `as_ratio=False` from `FullFormatter.format_unit()` when the unit format spec contains `^`
- add a regression test for standalone inverse units
- add a changelog entry for #2313

Fixes #2313.

## Verification
- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_issues.py::test_issue2313 pint/testsuite/test_formatter.py -q`
- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_issues.py -q`
- `.venv-pint-test/bin/python -m ruff check pint/delegates/formatter/full.py pint/testsuite/test_issues.py`
- `.venv-pint-test/bin/python -m ruff format --check pint/delegates/formatter/full.py pint/testsuite/test_issues.py`

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@ Pint Changelog
 -------------------
 
 - Fix offset (affine) unit conversion of `Decimal` magnitudes raising `TypeError` from mixing `Decimal` with the float scale/offset (e.g. `Quantity(Decimal(10), "degC").to("K")`), and the mirror case of a plain `float` magnitude in a `Decimal` registry (#2305)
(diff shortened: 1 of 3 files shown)


therealclvn:
FYI, the ReadTheDocs failure here appears to be the same docs dependency issue fixed separately in #2334, not this formatter change.

PR #2336, opened 2026-07-12
Allow wraps to pass var keyword arguments

### LINKED ISSUE #2175: `UnitRegistry.wraps` cannot handle unitless keyword arguments
Say we have the following function:
```
from pint import UnitRegistry
u = UnitRegistry()

def foo(arg1, arg2, kwarg1=0, **kwargs):
    if 'printmsg' in kwargs and kwargs['printmsg']:
        print("This is a message from foo.")
        return arg1 + arg2 + kwarg1
    else:
        return arg1 + kwarg1 + arg2
```

When run on its own, we get:
```
>>>print(foo(1, 2, arg3=3, kwarg1=5, arg4="doesn't matter"))
8
```
Wrapping the function with `u.wraps` proves challenging, as the unit registry expects a unit for the argument with label `"kwargs"` in the function signature. Still, this can be worked around by providing `None` as the unit for this argument. I assume however that this is not the intended functionality.
```
#                                                                         |
#                                                                         v
wrapped_foo = u.wraps(ret=u.centimeter, args=(u.meter, u.meter, u.meter, None), strict=False)(foo)
```
However, even though the wrapping itself works, calling the wrapped function will fail:
```
>>>print(wrapped_foo(1*u.meter, 2*u.meter, kwarg1=5*u.meter, printmsg=True))
  File File "<python-input-7>", line 1, in <module>
    print(wrapped_foo(1*u.meter, 2*u.meter, kwarg1=5*u.meter, printmsg=True))
          ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\path\to\Python\Python313\Lib\site-packages\pint\registry_helpers.py", line 289, in wrapper
                                         ~~~~~~~~~^
        ureg, sig, values, kw, strict
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "C:\path\to\Python\Python313\Lib\site-packages\pint\registry_helpers.py", line 128, in _converter
    values.append(kw[param_name])
                  ~~^^^^^^^^^^^^
KeyError: 'kwargs'
```
Wouldn't it be better to allow for arbitrary `**kwargs` which are simply passed to the function without being handled for units?

therealclvn:
## Summary
- skip `**kwargs` parameters when packing/unpacking values for `UnitRegistry.wraps()`
- allow wrapped functions with arbitrary keyword arguments without requiring a unit slot for `**kwargs`
- keep compatibility with the previous workaround of passing a trailing `None` unit for `**kwargs`
- add regression coverage and a changelog entry

Fixes #2175.

## Verification
- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_unit.py -k test_wraps -q`
- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_unit.py -q`
- `.venv-pint-test/bin/python -m ruff check pint/registry_helpers.py pint/testsuite/test_unit.py`
- `.venv-pint-test/bin/python -m ruff format --check pint/registry_helpers.py pint/testsuite/test_unit.py`


therealclvn:
FYI, the ReadTheDocs failure here appears to be the same docs dependency issue fixed separately in #2334, not this `wraps()` change.

PR #2337, opened 2026-07-12
Accept strings in preferred unit lists

### LINKED ISSUE #2185: Method `to_preferred` does not accept strings
Instead of this:
```
Quantity('g').to(Unit('kg'))
```
I can also simply do this:
```
Quantity('g').to('kg')
```
It means that the string is automatically cast to a `Unit`.

However, instead of this:
```
Quantity('g').to_preferred([Unit('kg')])
```
I cannot do this:
```
Quantity('g').to_preferred(['kg'])
```
It will raise this Exception:
```
AttributeError: 'str' object has no attribute 'dimensionality'
```
It means that the string is not automatically cast to a `Unit`.

This is probably unintended, as the [API documentation](https://pint.readthedocs.io/en/latest/api/base.html) says that `to_preferred()` accepts objects of type `UnitLike`, which includes strings.

therealclvn:
## Summary
- normalize `preferred_units` through the registry before `to_preferred()` selects units
- allow strings in preferred unit lists, matching the documented `UnitLike` type
- add regression coverage for both `to_preferred()` and `ito_preferred()`
- add a changelog entry

Fixes #2185.

## Verification
- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_quantity.py::TestQuantity::test_to_preferred_accepts_unitlike_strings -q`
- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_quantity.py -k to_preferred -q`
- `.venv-pint-test/bin/python -m ruff check pint/facets/plain/qto.py pint/testsuite/test_quantity.py`
- `.venv-pint-test/bin/python -m ruff format --check pint/facets/plain/qto.py pint/testsuite/test_quantity.py`


therealclvn:
FYI, the ReadTheDocs failure here appears to be the same docs dependency issue fixed separately in #2334, not this `to_preferred()` change.

PR #2338, opened 2026-07-12
Warn on application registry getitem

### LINKED ISSUE #2267: Improve warning about deprecation of getitem usage
When running the test, I meet this warning

```
  /tmp/.tox/py313-linux-pyqt5-no_cov/lib/python3.13/site-packages/pint/registry.py:264: DeprecationWarning: Calling the getitem method from a UnitRegistry will be removed in future versions of pint.
  use `parse_expression` method or use the registry as a callable.
    return self._registry[item]
```
It points inside the pint code:

https://github.com/hgrecco/pint/blob/6049a9c5b75052d3de22515e3e3e5076f8492e0b/pint/registry.py#L263-L264

https://github.com/hgrecco/pint/blob/6049a9c5b75052d3de22515e3e3e5076f8492e0b/pint/facets/plain/registry.py#L380-L385 

That makes it harder to debug where in our codebase there is a problem. It will be nice to add the same warning to `ApplicationRegistry.__getitem__` (to allow easier finding of problematic places) or reimplement it to use `parse_expression`. 

I might have made a contribution, but I'm not sure what you prefer.

therealclvn:
## Summary

- add the deprecation warning to `ApplicationRegistry.__getitem__()`
- call `parse_expression()` directly so the application registry emits only one warning
- add a regression test for the warning and parsed unit result

Fixes #2267

## Testing

- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_application_registry.py -q`
- `.venv-pint-test/bin/python -m ruff check pint/registry.py pint/testsuite/test_application_registry.py`
- `.venv-pint-test/bin/python -m ruff format --check pint/registry.py pint/testsuite/test_application_registry.py`


therealclvn:
FYI, the ReadTheDocs failure here appears to be the same docs dependency issue fixed separately in #2334, not this `ApplicationRegistry.__getitem__()` warning change.

PR #2339, opened 2026-07-12
Guard linalg solve against offset units

### LINKED ISSUE #2211: Bug: can perform `numpy.linalg.solve` on offset unit
Offset units (such as `degC`) shouldn't be allowed to perform `numpy.linalg.solve`. However, in current pint, the following code doesn't throw error:
```python
from pint import UnitRegistry;ureg=UnitRegistry();import numpy
numpy.linalg.solve(ureg.Quantity([[1,2],[3,4]], "degC"), [1,1])
```
and the result is `<Quantity([-1.  1.], '1 / degree_Celsius')>`

therealclvn:
## Summary

- handle `numpy.linalg.solve` with a dedicated NumPy function wrapper
- apply the existing offset-unit guard before stripping magnitudes for solve inputs
- keep the existing output unit calculation as `b / a`
- add regression coverage for offset-unit matrices

Fixes #2211

## Testing

- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_numpy.py -k solve -q`
- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_numpy.py -q`
- `.venv-pint-test/bin/python -m ruff check pint/facets/numpy/numpy_func.py pint/testsuite/test_numpy.py`
- `.venv-pint-test/bin/python -m ruff format --check pint/facets/numpy/numpy_func.py pint/testsuite/test_numpy.py`


therealclvn:
FYI, the ReadTheDocs failure here appears to be the same docs dependency issue fixed separately in #2334, not this `numpy.linalg.solve` offset-unit guard.

PR #2340, opened 2026-07-12
Fix wraps referenced units with integer arrays

### LINKED ISSUE #2109: Relative units with numpy integer division raises error
It looks like an edge case, but I guess it's an error coming from `pint`, so I raise it here.

### What obviously works

```python
def f(x, y):
    return x / y

print(f(np.array([1]), np.array([2])))
```

Output: `[0.5]`

### What raises an error

```python
from pint import UnitRegistry

ureg = UnitRegistry()


@ureg.wraps("=A/B", ("=A", "=B"))
def f(x, y):
    return x / y

print(f(np.array([1.]) * ureg.angstrom, np.array([2.]) * ureg.second)) # This one works
print(f(np.array([1]) * ureg.angstrom, np.array([2]) * ureg.second)) # This one raises an error
```

Here's the error message: `ValueError: Integers to negative integer powers are not allowed.` It's not impossible to understand what it means, but since `numpy` automatically converts it to a float, I think `pint` should also support it.

I'm using version 0.24.4

therealclvn:
## Summary

- derive referenced `wraps()` return units from argument units only
- avoid applying negative powers to integer NumPy array magnitudes while resolving units such as `=A/B`
- add regression coverage for integer array quantities in referenced return units

Fixes #2109

## Testing

- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_unit.py -k test_wrap -q`
- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_unit.py -q`
- `.venv-pint-test/bin/python -m ruff check pint/registry_helpers.py pint/testsuite/test_unit.py`
- `.venv-pint-test/bin/python -m ruff format --check pint/registry_helpers.py pint/testsuite/test_unit.py`

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@ Pint Changelog
 -------------------
 
 - Fix offset (affine) unit conversion of `Decimal` magnitudes raising `TypeError` from mixing `Decimal` with the float scale/offset (e.g. `Quantity(Decimal(10), "degC").to("K")`), and the mirror case of a plain `float` magnitude in a `Decimal` registry (#2305)
+- Fix `UnitRegistry.wraps()` referenced return units with negative powers for integer NumPy array inputs (#2109)
 - Add `system` parameter to `to_base_units` and `ito_base_units` to allow converting to the base units of a specific unit system (#2297)
 - Add beats per minute (`beats_per_minute`, `bpm`) (#2286)
 - Drop support for Python 3.11 in favour of Python 3.14 (#2304)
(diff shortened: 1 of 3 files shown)


therealclvn:
FYI, the ReadTheDocs failure here appears to be the same docs dependency issue fixed separately in #2334, not this `UnitRegistry.wraps()` referenced-unit change.

andrewgsavage:
closing in vavour of https://github.com/hgrecco/pint/pull/2362

PR #2342, opened 2026-07-12
Handle public units in quantity muldiv

### LINKED ISSUE #2046: pint.Quantity.units / pint.Unit fails
Steps to reproduce:
```
a = pint.Quantity(1, "L").units
b = pint.Unit("mL")
c = a / b
```

Expected behavior:
`c` is a dimensionless unit representing 1000

Actual behavior:
AttributeError: 'PlainUnit' object has no attribute '_get_non_multiplicative_units'

---

The offending line is in pint/facets/plain/quantity.py:999:
```
no_offset_units_other = len(other._get_non_multiplicative_units())
```
`other` does not have this method, because `other` is a `Unit`, and not a `NonMultiplicativeQuantity`. It looks like the bug is coming from quantity.py:989:
```
if isinstance(other, self._REGISTRY.Unit):
    other = 1 * other
```

This is evaluating to `False`, so `other` is remaining a Unit rather than being converted to a Quantity. I suspect this isn't quite the right logic for this conditional. I believe the intention is "is this item a Unit and attached to the same registry I am," but what it is actually checking is, "was this item constructed as an instance of my registry's inner class called Unit."

---
(Unnecessary) detail:

- `pint.Quantity(1, "L")` invokes `SharedRegistryObject.__new__`, which gives the created object a field `_REGISTRY` pointing to the application registry
- The `units()` property returns a `self._REGISTRY.Unit`, which does not have a `_REGISTRY` field but is an instance of an inner class of the application registry
- `pint.Unit("mL") invokes `SharedRegistryObject.__new__`, so again it has the application registry as a field, but it isn't an instance of that registry's inner Unit class

therealclvn:
## Summary

- recognize public `Unit` objects attached to the same registry during quantity multiplication and division
- convert those units to unity quantities before offset-unit checks, matching registry inner `Unit` handling
- add regression coverage for `*`, `*=`, `/`, and `/=` with public `Unit` objects

Fixes #2046

## Testing

- `.venv-pint-test/bin/python -m pytest pint/testsuite/test_quantity.py -k plain_unit_from_same_registry -q`
- `.venv-pint-test/bin/python -m ruff check pint/facets/plain/quantity.py pint/testsuite/test_quantity.py`
- `.venv-pint-test/bin/python -m ruff format --check pint/facets/plain/quantity.py pint/testsuite/test_quantity.py`

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@ Pint Changelog
 -------------------
 
 - Fix offset (affine) unit conversion of `Decimal` magnitudes raising `TypeError` from mixing `Decimal` with the float scale/offset (e.g. `Quantity(Decimal(10), "degC").to("K")`), and the mirror case of a plain `float` magnitude in a `Decimal` registry (#2305)
+- Allow multiplying and dividing quantities by public `Unit` objects from the same registry (#2046)
 - Add `system` parameter to `to_base_units` and `ito_base_units` to allow converting to the base units of a specific unit system (#2297)
 - Add beats per minute (`beats_per_minute`, `bpm`) (#2286)
 - Drop support for Python 3.11 in favour of Python 3.14 (#2304)
--- a/pint/facets/plain/quantity.py
+++ b/pint/facets/plain/quantity.py
@@ -978,6 +978,13 @@ class PlainQuantity(PrettyIPython, SharedRegistryObject, Generic[MagnitudeT_co])
 
         return -self._add_sub(other, operator.sub)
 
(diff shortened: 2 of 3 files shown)


therealclvn:
FYI, the ReadTheDocs failure here appears to be the same docs dependency issue fixed separately in #2334, not this public `Unit` multiplication/division change.

andrewgsavage:
closing in favour of 

https://github.com/hgrecco/pint/pull/2359

PR #2349, opened 2026-07-17
Raise DefinitionSyntaxError on unbalanced or empty parentheses

santhreal:
Parsing an expression with unbalanced or empty parentheses leaked a low-level error instead of pint's own `DefinitionSyntaxError`:

- `ureg("(m")` (unbalanced) reached the stdlib tokenizer, which raised `tokenize.TokenError`.
- `ureg("()")` (empty group) hit `assert result is not None` in `_build_eval_tree`, which raises `AssertionError` (and is stripped entirely under `python -O`, leaving a `None` to crash later).

Both now raise `DefinitionSyntaxError`, matching the other parse-error paths in this module.

```python
import pint
pint.UnitRegistry()("(m")  # was: tokenize.TokenError
pint.UnitRegistry()("()")  # was: AssertionError
```


andrewgsavage:
Please make an issue before making a PR

Don't include the fixes for CI or docs in here

I am not a fan of making the eval_tree code more complicated, it's already long and difficult to follow. I would like to see other people have this issue first before making the eval_tree code longer.

PR #2350, opened 2026-07-17
Format negative magnitudes in scientific notation with the pretty exponent

vineethsaivs:
The pretty (`~P`) and HTML (`~H`) formatters turn a magnitude's scientific notation into a nice `×10ⁿ` exponent, but only for positive magnitudes:

```python
>>> import pint
>>> u = pint.UnitRegistry()
>>> f"{2.3e-6 * u.m:~P}"
'2.3×10⁻⁶ m'
>>> f"{-2.3e-6 * u.m:~P}"
'-2.3e-06 m'      # raw e-notation, not formatted
```

The formatters detect the exponent with `_EXP_PATTERN.match(mstr)`. `re.match` anchors at the start of the string, and `_EXP_PATTERN` begins with `[0-9]`, so a negative magnitude (leading `-`) never matches and the `×10ⁿ` substitution is skipped. Positive magnitudes match at index 0 and are formatted.

## Fix

Use `_EXP_PATTERN.search(mstr)` in the plain and HTML formatters so the mantissa is found regardless of a leading sign. (The LaTeX formatter already calls `_EXP_PATTERN.sub(...)` directly and is unaffected.) After the fix:

```python
>>> f"{-2.3e-6 * u.m:~P}"
'-2.3×10⁻⁶ m'
>>> f"{-2.3e-6 * u.m:~H}"
'-2.3×10<sup>-6</sup> m'
```

Positive magnitudes are unchanged.

## Test

Added `test_negative_magnitude_scientific_notation_formatting` in `pint/testsuite/test_issues.py`, asserting matching `~P`/`~H` output for positive and negative magnitudes. It fails before this change (`'-2.3e-06 m' != '-2.3×10⁻⁶ m'`) and passes after. The existing formatter/formatting/measurement tests still pass.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Pint Changelog
 0.26.0 (unreleased)
 -------------------
 
+- Fix negative magnitudes in scientific notation not getting the pretty/HTML exponent formatting (e.g. `-1.5e-08 m` now renders as `-1.5×10⁻⁸ m`, matching the positive case) (#2350)
 - Fix offset (affine) unit conversion of `Decimal` magnitudes raising `TypeError` from mixing `Decimal` with the float scale/offset (e.g. `Quantity(Decimal(10), "degC").to("K")`), and the mirror case of a plain `float` magnitude in a `Decimal` registry (#2305)
 - Add `system` parameter to `to_base_units` and `ito_base_units` to allow converting to the base units of a specific unit system (#2297)
 - Add beats per minute (`beats_per_minute`, `bpm`) (#2286)
--- a/pint/delegates/formatter/html.py
+++ b/pint/delegates/formatter/html.py
@@ -76,7 +76,7 @@ class HTMLFormatter(BaseFormatter):
                         + "</pre>"
                     )
 
-        m = _EXP_PATTERN.match(mstr)
+        m = _EXP_PATTERN.search(mstr)
         _exp_formatter = lambda s: f"<sup>{s}</sup>"
 
         if m:
--- a/pint/delegates/formatter/plain.py
+++ b/pint/delegates/formatter/plain.py
@@ -307,7 +307,7 @@ class PrettyFormatter(BaseFormatter):
             else:
                 mstr = format_number(magnitude)
 
-            m = _EXP_PATTERN.match(mstr)
+            m = _EXP_PATTERN.search(mstr)
 
             if m:
                 exp = int(m.group(2) + m.group(3))
--- a/pint/testsuite/test_issues.py
+++ b/pint/testsuite/test_issues.py
@@ -1477,6 +1477,19 @@ def test_issue2256_2():
     assert f"{q:^}" == "2.3e-06 kilogram ** -1 * meter ** 3 * second ** -2"
 
 
+def test_negative_magnitude_scientific_notation_formatting():
+    # A negative magnitude shown in scientific notation must get the same pretty
+    # (and HTML) exponent formatting as its positive counterpart. _EXP_PATTERN
+    # was applied with .match(), which anchors at the start and so failed on the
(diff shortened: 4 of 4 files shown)


andrewgsavage:
this is the third PR to fix 
https://github.com/hgrecco/pint/issues/2313

vineethsaivs:
CI caught a regression in my own change, fixed in `6b2c080`.

Switching `_EXP_PATTERN.match()` to `.search()` fixed negative scalars but broke array magnitudes. An array formats as a single bracketed string holding several numbers, and `.search()` finds an exponent anywhere in it, so the guard passed and the substitution ran. The replacement is built once from the first match:

```python
exp = int(m.group(2) + m.group(3))
mstr = _EXP_PATTERN.sub(r"\1×10" + pretty_fmt_exponent(exp), mstr)
```

so every element got the *first* element's exponent. `[1e-16 1 1e+10]` came out as `[1×10⁻¹⁶ 1 1×10⁻¹⁶]`, wrong numbers rather than a cosmetic difference. Thanks to `test_quantity_array_format` for catching it.

The leading sign was the only thing actually missing, so it now lives in the pattern and the anchored `match()` stays:

```python
_EXP_PATTERN = re.compile(r"(-?[0-9]\.?[0-9]*)e(-?)\+?0*([0-9]+)")
```

| input | old `.match` | my `.search` | new `.match` |
|---|---|---|---|
| `2.3e-06` | yes | yes | yes |
| `-2.3e-06` | **no** | yes | yes |
| `[1e-16 1 1e+10]` | no | **yes** | no |

Scalars match with the sign captured in group 1, so `\1` carries it through the substitution, and a bracketed array string still fails at position 0 and falls through untouched, exactly as on master. Same change in `html.py`.

Also added an array assertion to the test so this cannot regress again. Locally on 3.13: `test_quantity_array_format` and the new test both pass, and reverting to the previous commit reproduces the CI failure.

vineethsaivs:
Understood on the duplicate, closing is fine by me and I should have checked #2313 for prior attempts before opening this.

Leaving one note for whoever lands the eventual fix, since it caught me and it is not obvious: the natural fix here is to relax `_EXP_PATTERN.match()` to `.search()` so the leading `-` stops blocking the match. That silently breaks array magnitudes. An array formats as a single bracketed string of several numbers, `.search()` finds an exponent inside it, and the substitution then builds its replacement once from the first match:

```python
exp = int(m.group(2) + m.group(3))
mstr = _EXP_PATTERN.sub(r"\1×10" + pretty_fmt_exponent(exp), mstr)
```

so every element gets the *first* element's exponent, and `[1e-16 1 1e+10]` renders as `[1×10⁻¹⁶ 1 1×10⁻¹⁶]`. Wrong numbers, not just wrong styling. `test_quantity_array_format` catches it.

Keeping the anchored `match()` and putting the sign in the pattern avoids that entirely, because a bracketed string still fails to match at position 0:

```python
_EXP_PATTERN = re.compile(r"(-?[0-9]\.?[0-9]*)e(-?)\+?0*([0-9]+)")
```

Thanks for the reviews today.

andrewgsavage:
oh good catch. I went with the simpler PRs, thanks for explaining here. If you put a PR with those changes in again I'll take a look.

vineethsaivs:
Opened it as #2371, rebased on current master.

Same fix, but rewritten to be array-safe: the sign is in the pattern and the anchored `match()` stays, rather than relaxing to `search()`. There is a second test guarding specifically against the `search()` version, since that is the change anyone looking at this would reach for first and it silently rewrites every element of an array with the first element's exponent.

Verified the bug is still live on master before opening: `1.5e-07` formats as `1.5×10⁻⁷` while `-1.5e-07` stays raw. Full suite green on 3.13 (2266 passed).

Issue #2356, opened 2026-07-26
Adopt Ruff 0.16's expanded default rule set (currently pinned to old defaults)


andrewgsavage:
## Background

Ruff 0.16.0 expanded its default rule set from 59 rules (`E4`, `E7`, `E9`, `F`) to 413 rules across 34 categories, adding whole categories like `B` (flake8-bugbear), `UP` (pyupgrade), `SIM` (flake8-simplify), `RUF`, `PERF`, `TRY`, `PL*`, etc. See https://astral.sh/blog/ruff-v0.16.0.

Since `pyproject.toml`'s `[tool.pixi.feature.lint.dependencies]` pins `ruff = "*"` (unpinned) and `pixi.lock` isn't committed to the repo, CI silently picked up 0.16.0 and started failing lint on ~365 pre-existing violations that have nothing to do with whatever PR happened to trigger the CI run.

To keep CI stable, `[tool.ruff.lint]` in `pyproject.toml` now pins:

```toml
select = ["E4", "E7", "E9", "F", "I"]
ignore = [
  "E402",
  "E731",
  "E741",
]
```

which reproduces Ruff's old (pre-0.16) default behavior plus the isort (`I`) rules this project already opted into.

## What's needed to unpin

Adopting the new defaults means fixing (or explicitly ignoring) the violations below, then replacing the explicit `select` with the previous `extend-select = ["I"]` (or an intentional new list). Rough breakdown of current violation counts with `ruff check --isolated .` (i.e. Ruff's real 0.16 defaults, ignoring this repo's pin):

```
48  C408    unnecessary-collection-call
39  UP031   printf-string-formatting
35  RUF012  mutable-class-default
25  RUF059  unused-unpacked-variable
16  BLE001  blind-except
15  SIM201  negate-equal-op
14  RUF100  unused-noqa
12  SIM117  multiple-with-statements
11  RUF022  unsorted-dunder-all
 9  D419    empty-docstring
 8  B015    useless-comparison
 8  PLW0127 self-assigning-variable
 8  UP032   f-string
 7  PLR0124 comparison-with-itself
 6  RUF010  explicit-f-string-type-conversion
 6  SIM202  negate-not-equal-op
 6  UP045   non-pep604-annotation-optional
 5  SIM102  collapsible-if
 5  TRY002  raise-vanilla-class
 5  UP007   non-pep604-annotation-union
 5  UP037   quoted-annotation
 4  B023    function-uses-loop-variable
[...]

PR #2359, opened 2026-07-26
Handle public Unit/Quantity objects in muldiv, from_, and wraps

andrewgsavage:
## Summary

Reworks #2342's fix for #2046 (public `Unit` objects rejected during quantity `*`/`/`), and fixes two more occurrences of the same underlying bug.

#2342 fixed `Quantity.__mul__`/`__truediv__` rejecting "public" `Unit` objects (e.g. `pint.Unit(...)`) attached to the same registry by adding `isinstance(other, self._REGISTRY.Unit)`. That check is fragile: `self._REGISTRY.Unit` is the per-instance dynamic subclass a `UnitRegistry` builds for itself in `_init_dynamic_classes`, whereas `pint.Unit`/`pint.Quantity` are the un-specialized `UnitRegistry.Unit`/`.Quantity` aliases. Both kinds of object share the same `_REGISTRY`, but a "public" object can still fail an `isinstance` check against that specific dynamic subclass.

This reworks the fix to check against the base `PlainUnit`/`PlainQuantity` classes instead — the same pattern the registry's own operators already use (`PlainUnit.__mul__`/`__eq__`/`compare` all gate on `self._check(other)`, which does a duck-typed `_REGISTRY` identity comparison, plus an `isinstance` against the base class or `self.__class__`).

Applied the same fix to two more call sites with the identical bug, found while reviewing #2342:
- `Unit.from_()` — checked `isinstance(value, self._REGISTRY.Quantity)`.
- `wraps()` / `_parse_wrap_args()` — checked `isinstance(arg, (ureg.Unit, str))` in three places (arg validation, ret validation, and the inner argument-conversion pass).

## Testing

- Added `test_issue2046` (`Unit.from_` and `wraps` with public `Unit`/`Quantity` objects) — verified it fails against the pre-rework `isinstance(x, self._REGISTRY.X)` checks and passes with the base-class checks.
- Full suite: `pixi run --environment test test` — 1637 passed, 1035 skipped, 9 xfailed.
- `pixi run --environment lint ruff check`/`format --check` on the changed files (checked against this repo's pinned pre-0.16 rule set, since `ruff` itself is unpinned on this branch — see #2356).

## Test plan

- [ ] CI: full test suite and lint pass

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@ Pint Changelog
 -------------------
 
 - Fix offset (affine) unit conversion of `Decimal` magnitudes raising `TypeError` from mixing `Decimal` with the float scale/offset (e.g. `Quantity(Decimal(10), "degC").to("K")`), and the mirror case of a plain `float` magnitude in a `Decimal` registry (#2305)
+- Allow multiplying and dividing quantities by public `Unit` objects from the same registry (#2046)
 - Add `system` parameter to `to_base_units` and `ito_base_units` to allow converting to the base units of a specific unit system (#2297)
 - Add beats per minute (`beats_per_minute`, `bpm`) (#2286)
 - Drop support for Python 3.11 in favour of Python 3.14 (#2304)
--- a/CHANGES
+++ b/CHANGES
@@ -5,7 +5,7 @@ Pint Changelog
 -------------------
 
 - Fix offset (affine) unit conversion of `Decimal` magnitudes raising `TypeError` from mixing `Decimal` with the float scale/offset (e.g. `Quantity(Decimal(10), "degC").to("K")`), and the mirror case of a plain `float` magnitude in a `Decimal` registry (#2305)
-- Allow multiplying and dividing quantities by public `Unit` objects from the same registry (#2046)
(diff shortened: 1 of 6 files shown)


RBerga06:
I also noticed this problem when adding more tests for #2427, essentially stumbling upon #2046. In case it can help, I've verified that:
```python
# tested on current master
>>> from pint import Quantity
>>> type(Quantity(3)) is Quantity
True
>>> Quantity(3)._REGISTRY.Quantity is Quantity
False

>>> from pint import UnitRegistry
>>> ureg = UnitRegistry()
>>> ureg.Quantity(3, "m").u / ureg.Unit("m")  # works fine
Unit("dimensionless")
>>> ureg.Unit("m") / ureg.Quantity(3, "m").u  # works fine
Unit("dimensionless")
```
To my understanding, the moment a registry is constructed it invokes `GenericUnitRegistry._init_dynamic_classes()` which creates new `Quantity` and `Unit` subclasses with that registry. What might be happening here is that, although the "root" `Quantity` and `Unit` classes already use the application registry by default (because of `SharedRegistryObject.__new__`), when the application registry is initialized the new classes are created anyway (although it would be unnecessary), so I think the easiest fix would be to "just" prevent `_init_dynamic_classes` from running in the application registry.

PR #2367, opened 2026-07-26
Support * and / between Quantity and datetime.timedelta

### LINKED ISSUE #2348: Support `*` and `/` between `timedelta` and `Quantity`
Following #510 and #608, it would be nice if `Quantity` supported transparently division and multiplication with `datetime.timedelta` (returning another `Quantity`). For example:
```python3
>>> from pint import Quantity
>>> from datetime import timedelta
>>> t = timedelta(seconds=10)
>>> q = Quantity(3.6, "m/s")
```
Current behaviour (7a927b4):
```python
>>> q * t
Quantity(datetime.timedelta(seconds=36), "meter / second")
>>> t * q
Quantity(datetime.timedelta(seconds=36), "meter / second")
>>> q / t
TypeError: unsupported operand type(s) for /: 'float' and 'datetime.timedelta'
>>> t / q
Quantity(datetime.timedelta(seconds=2, microseconds=777778), "second / meter")
```
Proposed behaviour:
```python
>>> q * t
Quantity(36.0, "meter")
>>> t * q
Quantity(36.0, "meter")
>>> q / t
Quantity(0.36, "meter / second ** 2")
>>> t / q
Quantity(2.7777777777777777, "second ** 2 / meter")
```

Would this be feasible?

andrewgsavage:
## Summary
- Previously, `Quantity * timedelta` (and its reflected form) wrapped the raw `timedelta` object as the magnitude, and `Quantity / timedelta` raised `TypeError`, since `timedelta` was never converted into a proper duration `Quantity` before the arithmetic.
- `*` and `/` (in both directions, and their in-place forms) now convert a `datetime.timedelta` operand into a `Quantity` in seconds first, so the result is a normal `Quantity`:
  ```python
  >>> q = Quantity(3.6, "m/s")
  >>> t = timedelta(seconds=10)
  >>> q * t
  Quantity(36.0, "meter")
  >>> q / t
  Quantity(0.36, "meter / second ** 2")
  >>> t / q
  Quantity(2.7777777777777777, "second ** 2 / meter")
  ```
- Also fixes a latent bug in the `datetime.timedelta`-to-`Quantity` `__init__` path (added in #1978) that stored `"seconds"` (a plural alias) as the raw `UnitsContainer` key instead of resolving it to the canonical `"second"` name. This surfaced once `timedelta` operands were routed through regular arithmetic, which does direct unit-definition lookups.
- Note: `numpy.timedelta64` involved in numpy ufunc dispatch (e.g. `numpy_array_quantity * np.timedelta64(...)`) has a separate, pre-existing limitation not addressed here; this PR covers `datetime.timedelta`, matching the issue's scope, including quantities backed by numpy array magnitudes.

## Test plan
- [x] Added `TestTimedelta.test_mul_div` and `test_mul_div_np` in `pint/testsuite/test_quantity.py` covering scalar and numpy-array-backed quantities, both directions of `*`/`/`, and in-place `*=`/`/=`
- [x] Full test suite passes in both the plain (`pixi run -e test`) and numpy-inclusive (`pixi run -e docs`) environments

Fixes #2348

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@ Pint Changelog
 -------------------
 
 - Fix `unit in registry` raising `AttributeError` instead of returning `True` when checking membership with a `Unit` object rather than a string (#2156)
+- Support `*` and `/` between `Quantity` and `datetime.timedelta`, converting the `timedelta` to a `Quantity` in seconds (#2348)
 - Fix `DimensionalityError` from `numpy_scalar == quantity` for incompatible dimensions; now returns `False`/`True` like `quantity == numpy_scalar` (#2182)
 - Fix `wraps` raising an `AssertionError` for arguments declared as `"dimensionless"` (affecting math functions such as `exp` or `log`) (#2255)
 - Raise ``DefinitionSyntaxError`` instead of a bare ``AssertionError`` when parsing an expression that is only an operator, e.g. ``ureg('-')`` or ``ureg('*')`` (#2124)
--- a/CHANGES
+++ b/CHANGES
@@ -5,7 +5,7 @@ Pint Changelog
 -------------------
 
 - Fix `unit in registry` raising `AttributeError` instead of returning `True` when checking membership with a `Unit` object rather than a string (#2156)
-- Support `*` and `/` between `Quantity` and `datetime.timedelta`, converting the `timedelta` to a `Quantity` in seconds (#2348)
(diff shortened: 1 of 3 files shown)


RBerga06:
Thank you! I think this might need typing support though, I'll submit a PR fixing that.

PR #2371, opened 2026-07-26
Format negative magnitudes with the pretty exponent

vineethsaivs:
Re-opening this per @andrewgsavage's "if you put a PR with those changes in again I'll take a look" on #2350. This is the same fix, rewritten to be array-safe, and rebased on current master.

### The bug, still present on master

`_EXP_PATTERN` is applied with an anchored `match()` but has no leading sign, so a negative magnitude never matches at position 0:

```python
_EXP_PATTERN = re.compile(r"([0-9]\.?[0-9]*)e(-?)\+?0*([0-9]+)")
```

```
 1.5e-07 meter  ->  1.5×10⁻⁷ meter
-1.5e-07 meter  -> -1.5e-07 meter     <- unformatted
```

Both `plain.py` (`:P`) and `html.py` (`:H`) carry the same pattern and the same problem.

### The fix, and the trap it avoids

The obvious change is `match()` to `search()`, so the leading `-` stops blocking. That is what I did the first time, and it is wrong.

An array magnitude is formatted as one bracketed string containing several numbers. `search()` finds an exponent somewhere inside it, and then the substitution builds its replacement **once, from the first match**:

```python
exp = int(m.group(2) + m.group(3))
mstr = _EXP_PATTERN.sub(r"\1×10" + pretty_fmt_exponent(exp), mstr)
```

so every element is rewritten with the *first* element's exponent, and `[1e-16 1 1e+10]` renders as `[1×10⁻¹⁶ 1 1×10⁻¹⁶]`. Wrong numbers, not wrong styling.

So the sign goes in the pattern and the anchor stays:

```python
_EXP_PATTERN = re.compile(r"(-?[0-9]\.?[0-9]*)e(-?)\+?0*([0-9]+)")
```

A scalar matches at position 0 with the sign in group 1; a bracketed array still fails at position 0 and is left alone, exactly as today.

### Tests

Two, in `test_issues.py`:

- `test_negative_magnitude_pretty_exponent` covers the scalar case for `:P` and `:H`
- `test_negative_magnitude_pretty_exponent_leaves_arrays_alone` pins the trap above, so a future `search()` cannot pass review

Reverting just the pattern in both files turns the first red and leaves the second green, which is the right split: the array test is a guard, not the bug.

Full suite on Python 3.13: **2266 passed, 139 skipped, 11 xfailed**, benchmarks excluded (they need `pytest-codspeed`).

### DIFF
--- a/pint/delegates/formatter/html.py
+++ b/pint/delegates/formatter/html.py
@@ -41,10 +41,6 @@ if TYPE_CHECKING:
     from ...facets.plain import PlainQuantity, PlainUnit
 
 _EXP_PATTERN = re.compile(r"(-?[0-9]\.?[0-9]*)e(-?)\+?0*([0-9]+)")
-# The leading sign is part of the pattern rather than relaxing match() to search().
-# An array magnitude is one bracketed string of several numbers, and sub() below builds
-# its replacement once from the first match, so a search() would give every element the
-# first element's exponent: [1e-16 1 1e+10] would render as [1x10^-16 1 1x10^-16].
 
 
 class HTMLFormatter(BaseFormatter):
--- a/pint/delegates/formatter/html.py
+++ b/pint/delegates/formatter/html.py
@@ -40,7 +40,11 @@ if TYPE_CHECKING:
     from ...facets.measurement import Measurement
     from ...facets.plain import PlainQuantity, PlainUnit
 
-_EXP_PATTERN = re.compile(r"([0-9]\.?[0-9]*)e(-?)\+?0*([0-9]+)")
+_EXP_PATTERN = re.compile(r"(-?[0-9]\.?[0-9]*)e(-?)\+?0*([0-9]+)")
+# The leading sign is part of the pattern rather than relaxing match() to search().
+# An array magnitude is one bracketed string of several numbers, and sub() below builds
+# its replacement once from the first match, so a search() would give every element the
+# first element's exponent: [1e-16 1 1e+10] would render as [1x10^-16 1 1x10^-16].
 
 
 class HTMLFormatter(BaseFormatter):
--- a/pint/delegates/formatter/latex.py
+++ b/pint/delegates/formatter/latex.py
@@ -163,7 +163,7 @@ def siunitx_format_unit(
     return "".join(lpos) + "".join(lneg)
 
 
-_EXP_PATTERN = re.compile(r"([0-9]\.?[0-9]*)e(-?)\+?0*([0-9]+)")
+_EXP_PATTERN = re.compile(r"(-?[0-9]\.?[0-9]*)e(-?)\+?0*([0-9]+)")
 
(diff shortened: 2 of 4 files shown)


andrewgsavage:
does this need doing for latex?
this pattern exists there and is now different.
`_EXP_PATTERN = re.compile(r"([0-9]\.?[0-9]*)e(-?)\+?0*([0-9]+)")`

vineethsaivs:
Good question, and I checked rather than assumed: latex does not need it. On this branch, negative magnitudes already render correctly there.

```
                   spec    output
  scalar positive  latex   1.5\times 10^{-7}\ \mathrm{meter}
  scalar negative  latex   -1.5\times 10^{-7}\ \mathrm{meter}
  scalar positive  pretty  1.5×10⁻⁷ meter
  scalar negative  pretty  -1.5×10⁻⁷ meter
```

The reason is that latex uses the pattern differently, not that its copy happens to be luckier. `LatexFormatter.format_magnitude` calls `sub` straight away with a backreference replacement:

```python
mstr = _EXP_PATTERN.sub(r"\1\\times 10^{\2\3}", mstr)
```

There is no anchor, so for `-1.5e-07` the match starts at `1.5` and the leading `-` simply falls outside it and is left in place. `plain.py` and `html.py` instead call `_EXP_PATTERN.match(mstr)` first, which is anchored at position 0, so a leading `-` made the match fail outright and nothing was substituted at all. That anchor is the whole bug, and latex does not have it.

I also checked whether adding `-?` to latex's copy would change anything, in case it was worth doing just for consistency. It is a no-op for output: group 1 captures `-1.5` instead of the `-` sitting outside the match, and the rendered string is byte-identical. Nine values including `-1.5e-07`, `-6.022e+23`, `-0.0025` and `0.0`, zero differences.

The part of your comment I think is worth acting on is the divergence itself. There are three copies of this regex now (`plain.py:49`, `html.py:43`, `latex.py:166`) and after this PR mine differ from latex's, so the next person to read them has to work out whether that is deliberate. Worth noting before anyone unifies them: `pint/formatting.py` re-exports latex's copy under "Backwards compatiblity stuff", so `pint.formatting._EXP_PATTERN` is latex's specifically, which makes it the one with outside consumers.

So the options as I see them, and I would rather you pick than widen a bug fix on my own initiative:

1. Leave it. latex is correct, and its pattern is the one with a compatibility surface
2. Add `-?` to latex too, purely so the three read identically. No behaviour change, but it does change what `pint.formatting._EXP_PATTERN` matches for anyone importing it
3. Hoist one shared pattern somewhere common and have all three import it, keeping `pint.formatting._EXP_PATTERN` pointing at it. Cleanest, but it is a refactor rather than this fix, so I would rather do it as a follow-up unless you want it here

Happy to do 2 or 3 in this PR if you would prefer; just say which.

PR #2373, opened 2026-07-27
Improve typing for Quantity constructor and arithmetic with `timedelta`

RBerga06:
Typing improvements that reflect the new behaviour introduced in #2367.

---

**Edit**: Additional changes I made:
- Fixed a bug in `__pow__`'s type annotations (`Quantity(...) ** 2` previously resulted in an error, for example)
- Fixed a bug in `__add__`/`__iadd__` type annotations (`<Quantity[...]> + <datetime>` was previously annotated as returning `timedelta` instead of the correct `datetime`)
- Made `Quantity[_MagnitudeT_co]` inherit from `PlainQuantity[_MagnitudeT_co]` directly to reliably pass type information to the base class (this fixes `Quantity[<T>](...).magnitude` being `Any`/`Unknown` instead of the correct `<T>`)
- Other small annotation improvements

### DIFF
--- a/pint/facets/plain/quantity.py
+++ b/pint/facets/plain/quantity.py
@@ -943,10 +943,12 @@ class PlainQuantity(PrettyIPython, SharedRegistryObject, Generic[MagnitudeT_co])
 
         return self.__class__(magnitude, units)
 
+    # Quantity[float] += datetime -> becomes datetime
     @overload
     def __iadd__[T: int | float](
         self: PlainQuantity[T], other: datetime.datetime
-    ) -> datetime.timedelta: ...
+    ) -> datetime.datetime: ...
+    # General overload (should work in most cases)
     @overload
     def __iadd__[T: Magnitude, U: Magnitude](
         self: PlainQuantity[opt.CanIAdd[T, U]], other: PlainQuantity[T] | T
@@ -962,17 +964,12 @@ class PlainQuantity(PrettyIPython, SharedRegistryObject, Generic[MagnitudeT_co])
     @overload
     def __add__(
         self: PlainQuantity[int | float], other: datetime.datetime
-    ) -> datetime.timedelta: ...
+    ) -> datetime.datetime: ...
     # PlainQuantity[float | array[float]] + timedelta -> PlainQuantity[float | array[float]]
     @overload
-    def __add__(
-        self: PlainQuantity[float], other: datetime.timedelta | np.timedelta64
-    ) -> PlainQuantity[float]: ...
-    @overload
-    def __add__[X: np.floating, S: Shape](
-        self: PlainQuantity[npt.ArrayND[X, S]],
-        other: datetime.timedelta | np.timedelta64,
-    ) -> PlainQuantity[npt.ArrayND[X, S]]: ...
+    def __add__[T: int | float | npt.ArrayND[np.floating]](
+        self: PlainQuantity[T], other: datetime.timedelta | np.timedelta64
+    ) -> PlainQuantity[T]: ...
     # General overloads (should work in most cases)
     @overload
     def __add__[T: Magnitude, U: Magnitude](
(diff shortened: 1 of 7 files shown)


RBerga06:
I'll note that since the tests for the run-time part (#2367) use the operators explicitly, they would incidentally serve as tests for this PR as well.

pint/facets/plain/quantity.py
```diff
@@ -946,10 +958,22 @@ def __iadd__(self, other):
             return self._iadd_sub(other, operator.iadd)
         return self._add_sub(other, operator.add)
 
+    # PlainQuantity[float] + datetime -> datetime
     @overload
     def __add__(
-        self: PlainQuantity[int | float], other: datetime.datetime
```
andrewgsavage, line 951: why remove int here?

pint/facets/plain/quantity.py
```diff
@@ -932,7 +944,7 @@ def _add_sub(self, other, op):
         return self.__class__(magnitude, units)
 
     @overload
-    def __iadd__[T: int | float](
+    def __iadd__[T: float](
```
andrewgsavage: why remove int here?

pint/facets/plain/quantity.py
RBerga06: Ah sorry, that was meant to be a temporary change. Pyright treats `float` in type annotations as if it was `float | int` (so I instinctively thought it was superfluous), but this behaviour is not mandated by the typing spec so other type checkers might diverge: it's better to keep it.

pint/facets/plain/quantity.py
RBerga06: Same as above

RBerga06:
I've just found a way of making those overloads more concise. I'll update my PR accordingly.

**Edit:** I've made those changes and fixed a couple other typing bugs as well - I've updated the PR description.

pint/registry.py
```diff
@@ -130,28 +160,28 @@ def tolist[X: np.number](
...
             self: Quantity[int | float], other: datetime.datetime
-        ) -> datetime.timedelta: ...
+        ) -> datetime.datetime: ...
+        # General overload (should work in most cases)
         @overload
         def __iadd__[T: Magnitude, U: Magnitude](
             self: Quantity[opt.CanIAdd[T, U]], other: Quantity[T] | T
         ) -> Quantity[U]: ...
 
-        @overload
```
andrewgsavage, line 142: was this intentionally deleted?

RBerga06: Yes, because it was wrong:
```python
# just tested on master with Python 3.14.6 from conda-forge
>>> from pint import Quantity
>>> import datetime as dt
>>> q = Quantity(300.0, "s")
>>> q -= dt.datetime.now()
Traceback (most recent call last)
...
pint.errors.DimensionalityError: Cannot convert from 'second' to 'dimensionless'
# because pint tries to cast the datetime to `Quantity(datetime)` which is dimensionless,
#   but it's ok that this is an error because `[time]` quantities always represent time intervals
#   (they behave like timedelta)
>>> q -= dt.timedelta(seconds=3)
>>> q
Quantity(297.0, "s")
```
As you can see, `Quantity -= datetime` is prohibited (like `Quantity - datetime`) while `Quantity -= timedelta` is already covered typing-wise by the `__sub__` annotations - because IIRC when `__i<op>__` is not implemented Python also checks the corresponding `__<op>__` methods, so type checkers should be expected to understand this behaviour (although, again, this is underspecified by the spec).

In fact, as well as this overload I also removed the corresponding overload in `PlainQuantity.__isub__`, which made that method non-overloaded.

RBerga06:
@andrewgsavage Thank you very much for reviewing all these PRs, and more generally for dedicating your time to pint as a maintainer!

PR #2377, opened 2026-07-30
Fix DimensionalityError from floating-point noise in fractional exponents

### LINKED ISSUE #681: Unexpected DimensionalityError
Hello,

I use the Pint library for some time and I find it very useful and easy to use for computation involving physical quantities.

I'm facing an unexpected DimensionalityError in a computation using several quantities. I think this is a bug because the same computation with an intermediate and explicit conversion to meter works fine. See code below.

From the error, I guess the problem comes from the _[length] ** 1.11022e-16_ factor in the dimension:
> pint.errors.DimensionalityError: Cannot convert from 'meter ** 0.333333 * revolutions_per_minute * second * watt ** 0.333333 / kilogram ** 0.333333 / liter ** 0.333333' (1 / [length] ** 1.11022e-16 / [time] ** 1) to 'revolutions_per_minute' (1 / [time])

Configuration used:
- Python 3.6.4 on Windows 32b (problem reproduced on both W7 and W10)
- Pint 0.8.1
- numpy 1.14.0

Thanks for your help !

```
import pint
from numpy import sqrt

ureg = pint.UnitRegistry(system='mks')
Q = ureg.Quantity

n = Q(600, ureg.rpm)
q = Q(1000, ureg.watt) / (Q(42.0, 'kg/m**3') *  Q(43.0, 'm/s**2') * Q(100, ureg.meter) * 44.0)
f = Q(52.0, ureg.rpm) * 0.8 * q /  Q(10, 'l/s') * ureg.meter**3

# first computation of e. d is explicitly converted to meters. it works
print('\n', '*** With explicit conversion to meter ***', '\n')
d = ((f / n)**(1/3)).to(ureg.meter)
e = n * d / ureg.meter / sqrt(Q(100.0, ureg.meter)/ureg.meter)
e.to(ureg.rpm)
print(e)

# the same computation without converting d to meters fails
print('\n', '*** Without explicit conversion to meter ***', '\n')
d = (f / n)**(1/3)
e = n * d / ureg.meter / sqrt(Q(100.0, ureg.meter)/ureg.meter)
e.to(ureg.rpm)
print(e)
```

### LINKED ISSUE #1487: Specifying the floating precision of unit powers
Hello,

I am working with convection coefficients and they have some funky floating powers. Is there a way to specify precision of the units when reducing the units? For example I have the following equation

```python
h_c = 0.023 * V ** 0.8 * k ** 0.6 * (rho * cp) ** 0.4 / (D ** 0.2 * nu ** 0.4)
```

That I can properly reduced to base units

```python
h_c.to_base_units()
<Quantity(15710.5834, 'kilogram / kelvin ** 0.6 / meter ** 5.55112e-17 / second ** 2.6')>
````

But notice the `meter ** 5.55112e-17`. So that if I try to reduce to the units I want ("W/(m ** 2 * K)") or just `to_reduced_units()` I get the following error.

```python
(0.023 * (V ** 0.8 * k ** 0.6 * (rho * cp) ** 0.4) / (D ** 0.2 * nu ** 0.4)).to_reduced_units()
  File "<string>", line 1, in <module>
  File "/home/ijw/.local/lib/python3.8/site-packages/pint/quantity.py", line 738, in to_reduced_units
    newq.ito_reduced_units()
  File "/home/ijw/.local/lib/python3.8/site-packages/pint/quantity.py", line 728, in ito_reduced_units
    return self.ito(newunits)
  File "/home/ijw/.local/lib/python3.8/site-packages/pint/quantity.py", line 637, in ito
    self._magnitude = self._convert_magnitude(other, *contexts, **ctx_kwargs)
  File "/home/ijw/.local/lib/python3.8/site-packages/pint/quantity.py", line 616, in _convert_magnitude
    return self._REGISTRY.convert(
  File "/home/ijw/.local/lib/python3.8/site-packages/pint/registry.py", line 947, in convert
    return self._convert(value, src, dst, inplace)
  File "/home/ijw/.local/lib/python3.8/site-packages/pint/registry.py", line 1830, in _convert
    return super()._convert(value, src, dst, inplace)
  File "/home/ijw/.local/lib/python3.8/site-packages/pint/registry.py", line 1433, in _convert
    return super()._convert(value, src, dst, inplace)
  File "/home/ijw/.local/lib/python3.8/site-packages/pint/registry.py", line 980, in _convert
    raise DimensionalityError(src, dst, src_dim, dst_dim)
pint.errors.DimensionalityError: Cannot convert from 'kilogram ** 0.4 * watt ** 0.6 / centimeter ** 0.2 / kelvin ** 0.6 / meter / second ** 0.8' ([mass] / [length] ** 5.55112e-17 / [temperature] ** 0.6 / [time] ** 2.6) to 'kilogram ** 0.4 * watt ** 0.6 / centimeter ** 1.2 / kelvin ** 0.6 / second ** 0.8' ([mass] / [temperature] ** 0.6 / [time] ** 2.6)
```
Ultimately I think I just want to be able to truncate the powers of the units to 16 or so to be able to reduce. I tried a few different things and searched around, but could come up with anything clean.

Thanks so much! Love the library and I use it extensively.

andrewgsavage:
## Summary
- `UnitsContainer.add`/`__mul__`/`__truediv__` decided whether a cancelled-out exponent should be dropped with an exact `== 0` check. Repeated multiplication/division of fractional powers (e.g. `V**0.8 * k**0.6 * (rho*cp)**0.4 / (D**0.2 * nu**0.4)`) accumulates floating-point rounding error, so a unit that should cancel to exactly 0 was instead left with a residual exponent like `5.55e-17`. That leftover then made `to_base_units()`/`to_reduced_units()`/`to(...)` raise `DimensionalityError`, since the target dimensionality has no matching entry for it.
- The same noise also shows up as a *non-zero* exponent that's just off an integer, e.g. `[time] ** -0.9999999999999998` instead of exactly `[time] ** -1` (#681) — a plain "is this negligible/zero" check doesn't catch that case.
- `_get_root_units` and `_get_dimensionality` in `facets/plain/registry.py` have their own separate exponent accumulators with the same exact-comparison issue, so they needed the same fix.
- Added `_clean_exponent(value)` in `util.py`: a `float` gets snapped to the nearest integer when within `1e-9` of one (covering both the "should be 0" and "should be some other exact integer" cases); genuinely fractional exponents (like `0.333333`) and exact types (`int`, `Fraction`, `Decimal`) are left untouched.

Fixes #1487
Fixes #681

## Performance
Added a few hundred ns per `UnitsContainer` combine (`round()` + `math.isclose`), which is swamped by everything else a `Quantity` operation already does:

| operation | before | after | delta |
|---|---|---|---|
| `q1 * q2` | 9.05 µs | 9.35 µs | +3.3% |
| `q1 / q2` | 9.36 µs | 9.64 µs | +2.9% |
| `q1 ** 0.8` | 6.99 µs | 7.20 µs | +3.0% |
| full #1487 repro expression | 92.2 µs | 96.4 µs | +4.5% |
| `.to_base_units()` | 17.0 µs | 16.1 µs | -5.7% (noise) |
| `.to_reduced_units()` | 169.7 µs | 170.1 µs | +0.2% |

## Test plan
- [x] Added `test_issue1487` (original repro), `test_issue681` and `test_issue681_auto_reduce_dimensions` (both examples from that issue's thread, including the `auto_reduce_dimensions=True` case)
- [x] Full test suite passes (2369 passed)
- [x] `pre-commit run` clean
- [x] No new `pyright` errors (verified against the unmodified baseline)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,7 @@ Pint Changelog
 - Fix `DimensionalityError` from a leftover floating-point-noise exponent (e.g. `meter ** 5.55e-17` instead of exactly `meter ** 0`) after combining fractional unit powers, which broke `to_base_units`, `to_reduced_units`, and `to` (#1487)
 - Fix `DimensionalityError` from floating-point noise in combined fractional unit exponents (e.g. `meter ** 5.55e-17`) (#1487)
 - Fix `DimensionalityError` from floating-point noise in combined fractional unit exponents (e.g. `meter ** 5.55e-17` or `[time] ** -0.9999999999999998` instead of exactly `meter ** 0`/`[time] ** -1`) (#1487, #681)
+- Fix `DimensionalityError` from floating-point noise in combined fractional unit exponents (#1487, #681)
 - Fix `is_compatible_with` edge cases with dimensionless types and unit strings with a scaling factor (e.g. `"35000 ft"`) (#1190)
 - Allow string parsing of offset units (#386)
 - Remove the ambiguous `Nm` alias from `number_meter` (Textile group), since it's commonly expected to mean newton-meter (torque) (#1966)
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,7 @@ Pint Changelog
 - Switch CodSpeed CI benchmarks to `simulation` mode to fix a walltime environment warning.
 - Fix `DimensionalityError` from a leftover floating-point-noise exponent (e.g. `meter ** 5.55e-17` instead of exactly `meter ** 0`) after combining fractional unit powers, which broke `to_base_units`, `to_reduced_units`, and `to` (#1487)
 - Fix `DimensionalityError` from floating-point noise in combined fractional unit exponents (e.g. `meter ** 5.55e-17`) (#1487)
+- Fix `DimensionalityError` from floating-point noise in combined fractional unit exponents (e.g. `meter ** 5.55e-17` or `[time] ** -0.9999999999999998` instead of exactly `meter ** 0`/`[time] ** -1`) (#1487, #681)
 - Fix `is_compatible_with` edge cases with dimensionless types and unit strings with a scaling factor (e.g. `"35000 ft"`) (#1190)
 - Allow string parsing of offset units (#386)
 - Remove the ambiguous `Nm` alias from `number_meter` (Textile group), since it's commonly expected to mean newton-meter (torque) (#1966)
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@ Pint Changelog
 - Add a "Try Pint in your browser" page to the docs with an in-browser JupyterLite console and notebook, requiring no local install (#2372)
 - Switch CodSpeed CI benchmarks to `simulation` mode to fix a walltime environment warning.
 - Fix `DimensionalityError` from a leftover floating-point-noise exponent (e.g. `meter ** 5.55e-17` instead of exactly `meter ** 0`) after combining fractional unit powers, which broke `to_base_units`, `to_reduced_units`, and `to` (#1487)
+- Fix `DimensionalityError` from floating-point noise in combined fractional unit exponents (e.g. `meter ** 5.55e-17`) (#1487)
 - Fix `is_compatible_with` edge cases with dimensionless types and unit strings with a scaling factor (e.g. `"35000 ft"`) (#1190)
 - Allow string parsing of offset units (#386)
 - Remove the ambiguous `Nm` alias from `number_meter` (Textile group), since it's commonly expected to mean newton-meter (torque) (#1966)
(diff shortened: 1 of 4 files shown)


andrewgsavage:
There were some comments on previous issues about speed. I think a 3% difference is acceptable:

| operation | before | after | delta |
|---|---|---|---|
| `q1 * q2` | 8.92 µs | 9.20 µs | +3.1% |
| `q1 / q2` | 9.31 µs | 9.57 µs | +2.8% |
| `q1 ** 0.8` | 6.92 µs | 6.98 µs | +0.9% |
| full #1487 repro expression | 93.2 µs | 95.0 µs | +1.9% |
| `.to_base_units()` | 16.9 µs | 15.9 µs | -5.7% (faster) |
| `.to_reduced_units()` | 168.3 µs | 167.0 µs | -0.8% (faster) |

PR #2388, opened 2026-08-05
Fix RecursionError in find_connected_nodes for deeply chained graphs

lilu5458:
## Problem

`find_connected_nodes` in `pint/util.py` used unbounded recursion to traverse the context dimension graph. For each unvisited adjacent node it recursed once, so a linear chain longer than `sys.getrecursionlimit()` (default 1000) raised an uncaught `RecursionError`.

The function is reachable from the public API when contexts are enabled:

```
ureg.enable_contexts(...)
ureg.get_compatible_units(unit)
  -> ContextRegistry._get_compatible_units
  -> find_connected_nodes(self._active_ctx.graph, src_dim)
```

This is the same class of issue (CWE-674) that was fixed in `_build_eval_tree` (#2376).

## Fix

Replace the recursion with an iterative DFS using an explicit stack. The `visited` set still ensures each node is processed at most once, so the result is identical — only the call mechanism changes.

```python
stack = [start]
while stack:
    node = stack.pop()
    if node in visited:
        continue
    visited.add(node)
    stack.extend(graph[node] - visited)
```

## Verification

- PoC: a 2000-node linear chain that previously raised `RecursionError` now visits all 2001 nodes successfully.
- All 86 tests in `test_util.py` and `test_contexts.py` pass.
- Added a regression test `test_find_connected_nodes_deep_chain`.

## Files changed

- `pint/util.py` — recursive → iterative DFS (net +12 lines)
- `pint/testsuite/test_util.py` — regression test (+13 lines)
- `CHANGES` — changelog entry

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Pint Changelog
 0.26.0 (unreleased)
 -------------------
 
+- Fix `RecursionError` in `find_connected_nodes` for deeply chained context graphs by replacing recursion with an iterative DFS
 - Raise `DefinitionSyntaxError` instead of a bare `RecursionError` for deeply nested expressions
 - Add a "Try Pint in your browser" page to the docs with an in-browser JupyterLite console and notebook, requiring no local install (#2372)
 - Switch CodSpeed CI benchmarks to `simulation` mode to fix a walltime environment warning.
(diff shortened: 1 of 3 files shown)


lilu5458:
@andrewgsavage Thanks for cleaning up the comment! Is there anything else needed before this can be merged?

PR #2391, opened 2026-08-11
Fix ValueError when formatting units with Fraction exponents (#2386)

### LINKED ISSUE #2386: Formatting a Unit/Quantity with a Fraction exponent raises ValueError
## Description

When `non_int_type=fractions.Fraction` is used and a unit ends up with a `Fraction` exponent (e.g. from a fractional power, or from parsing a string exponent like `m**2.3`), formatting the `Unit`/`Quantity` (`str()`, `format()`, `repr` via default formatter) raises a `ValueError` instead of producing output.

## Reproduce

```python
import fractions, pint

ureg = pint.UnitRegistry(non_int_type=fractions.Fraction)
u = ureg.Unit("m**2.3")
str(u)
```

## Traceback

```
  File "<string>", line 6, in <module>
    print(str(u))
  File "pint\facets\plain\unit.py", line 75, in __str__
    return self._REGISTRY.formatter.format_unit(self)
  File "pint\delegates\formatter\full.py", line 138, in format_unit
    return self.get_formatter(uspec).format_unit(...)
  File "pint\delegates\formatter\plain.py", line 111, in format_unit
    return formatter(numerator, ...)
  File "pint\delegates\formatter\_format_helpers.py", line 226, in formatter
    pos_terms.append(power_fmt.format(key, fun(value)))
  File "pint\delegates\formatter\_format_helpers.py", line 217, in <lambda>
    fun = lambda x: exp_call(abs(x))
  File "fractions.py", line 580, in __format__
    raise ValueError(...)
ValueError: Invalid format specifier 'n' for object of type 'Fraction'
```

## Expected behavior

Formatting should either render the `Fraction` exponent (e.g. `m ** (23/10)`) or fall back to a sensible numeric representation, rather than raising.

## Cause

`_format_helpers.py`'s default exponent formatting path (around lines 217/226) assumes the exponent supports the `n` format spec, which `fractions.Fraction.__format__` does not accept. `str()`/`repr()` on the magnitude work fine — this is specific to `Unit`/`UnitsContainer` formatting.

## Related

Related to the broader exponent-type design discussion in #2379.

## Environment

- pint (current `master`/dev)
- Python 3.x, Windows

eeshsaxena:
Fixes #2386.

If you build a registry with `non_int_type=fractions.Fraction`, a unit can wind up with a `Fraction` exponent (from `Unit('m**2.3')`, a fractional power, etc). Formatting it blows up:

```python
import fractions, pint
ureg = pint.UnitRegistry(non_int_type=fractions.Fraction)
str(ureg.Unit('m**2.3'))
# ValueError: Invalid format specifier 'n' for object of type 'Fraction'
```

The exponent formatting path applies the `n` format spec (`f'{x:n}'`), which `Fraction.__format__` rejects. This hits every non-`/` spec: the default, compact (`~`) and pretty (`P`, `~P`) formatters all raise. The `/` pretty modifier already worked because it goes through `pretty_fmt_exponent_fraction`.

I added a small `_format_exponent_n` helper that applies the `n` spec but falls back to `str()` for types that don't accept it, and used it both for the default `exp_call` and inside `pretty_fmt_exponent`. Int and float exponents go through the exact same `f'{x:n}'` path as before, so nothing changes for them; only the previously-crashing types now render.

After the change:

```
''   -> 'meter ** 23/10'
'~'  -> 'm ** 23/10'
'P'  -> 'meter²³/¹⁰'
'~P' -> 'm²³/¹⁰'
```

Added two tests in `test_formatter.py`: a low-level `formatter()` check with a `Fraction` exponent, and an end-to-end check that all four specs render without raising. Both fail on master and pass with the fix.

Ran:
```
python -m pytest pint/testsuite/test_formatter.py pint/testsuite/test_formatting.py pint/testsuite/test_non_int.py
# 650 passed
```

Added a CHANGES entry too.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ Pint Changelog
 0.26.0 (unreleased)
 -------------------
 
+- Fix `ValueError` when formatting a `Unit`/`Quantity` whose exponent is a `fractions.Fraction` (e.g. registries created with `non_int_type=Fraction`), which does not support the `n` format spec (#2386)
 - Raise `DefinitionSyntaxError` instead of a bare `RecursionError` for deeply nested expressions
 - Add a "Try Pint in your browser" page to the docs with an in-browser JupyterLite console and notebook, requiring no local install (#2372)
 - Switch CodSpeed CI benchmarks to `simulation` mode to fix a walltime environment warning.
--- a/pint/delegates/formatter/_format_helpers.py
+++ b/pint/delegates/formatter/_format_helpers.py
@@ -92,10 +92,25 @@ def override_locale(
         setlocale(LC_NUMERIC, prev_locale_string)
 
 
+def _format_exponent_n(num: Number) -> str:
+    """Format an exponent using the locale-aware ``n`` spec, with a fallback.
+
+    Some numeric types do not accept the ``n`` presentation type and raise
+    ``ValueError`` when formatted with it - most notably ``fractions.Fraction``,
+    which is used as the exponent type when a registry is created with
+    ``non_int_type=fractions.Fraction``. For those, fall back to ``str`` so that
+    formatting produces output instead of raising.
+    """
+    try:
+        return f"{num:n}"
+    except (ValueError, TypeError):
+        return str(num)
+
+
 def pretty_fmt_exponent(num: Number) -> str:
     """Format an number into a pretty printed exponent."""
     # unicode dot operator (U+22C5) looks like a superscript decimal
-    ret = f"{num:n}".replace("-", "⁻").replace(".", "\u22c5")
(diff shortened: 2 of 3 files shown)


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Fix ValueError when formatting units with Fraction exponents (#2386)`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining pint!

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #2392, opened 2026-08-14
Raise DefinitionSyntaxError for empty and unbalanced parentheses

eeshsaxena:
Following the same thread as the recent missing-operand / deep-nesting hardening in the expression parser, I found two more malformed inputs that come back out of `parse_expression` as a Python-internal error rather than pint's own `DefinitionSyntaxError`:

- an empty parenthetical group such as `ureg('( )')` or `ureg('()')` hits `assert result is not None` in `_build_eval_tree`, so it raises a bare `AssertionError` with no message;
- an unclosed `(` such as `ureg('1 (')` or `ureg('(1 + 2')` makes Python's tokenizer raise `tokenize.TokenError` ("unexpected EOF in multi-line statement") when `build_eval_tree` materializes the token generator.

Both now raise `DefinitionSyntaxError`, consistent with how the parser already reports an unopened `)`, a missing operand, or a too-deeply-nested expression. I replaced the two `assert result is not None` lines with an explicit `DefinitionSyntaxError("empty parentheses in tokens")`, and wrapped the `list(tokens)` in `build_eval_tree` to convert `TokenError` into `DefinitionSyntaxError`.

Added parametrized tests alongside the existing missing-operand ones (both tokenizers). They fail on master (AssertionError / TokenError) and pass with the change; the full test_pint_eval and test_quantity suites still pass, and valid expressions including parenthetical groups and uncertainty notation like `1.0(2) meter` are unchanged. I found these by fuzzing `ureg.Quantity` with mutated valid expressions.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,8 @@ Pint Changelog
 -------------------
 
 - Raise `DefinitionSyntaxError` instead of a bare `RecursionError` for deeply nested expressions
+- Raise `DefinitionSyntaxError` instead of a bare `AssertionError` (empty parentheses) or
+  `tokenize.TokenError` (unbalanced parentheses) when parsing a malformed expression
 - Add a "Try Pint in your browser" page to the docs with an in-browser JupyterLite console and notebook, requiring no local install (#2372)
 - Switch CodSpeed CI benchmarks to `simulation` mode to fix a walltime environment warning.
 - Document that instantiating `Quantity` from a `(magnitude, unit)` pair is significantly faster than from a string or by multiplying by a unit (#1873)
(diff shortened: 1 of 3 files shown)


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Raise DefinitionSyntaxError for empty and unbalanced parentheses`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining pint!

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

Issue #2404, opened 2026-09-01
Typing: Ty reports `distance: Unknown = 1 * ureg.meter`


FabriceSalvaire:
I am aware there are many discussions about typing here...

But just to mention an important issue, actually Ty (LSP checker https://github.com/astral-sh/ty) reports:
```
distance«: Unknown» = 1 * ureg.meter
```
This is a real issue since Ty is not aware it is a `Pint.Quantity` !

zyv:
Did you try with `0.26.0.rc0`? I think this could have been fixed in #2302 and #2303 already. I'm currently looking into division and power, but multiplication seems to be working for me.

PR #2406, opened 2026-09-03
Improve type annotations for `Unit`'s `/`, `**` and copies

zyv:
`PlainUnit.__truediv__` is not annotated, so `ureg.meter / ureg.second` is inferred as `Unit | PlainQuantity[float] | Unknown` (pyright) or `Unknown` (ty), although the runtime returns a `Unit`.

`PlainUnit.__mul__` was typed in #2243, reverted in #2247, and typed again in #2373, but so far `PlainUnit.__truediv__` was omitted. This patch adds overloads for `__truediv__` mirroring `__mul__` and reuses `optype` like in `PlainQuantity.__truediv__`.

`PlainQuantity[float]` is used because `ureg.meter / 2` has a magnitude of `0.5`, and `__mul__`'s  `[T: Magnitude] -> PlainQuantity[T]` would make it an `int`. `__mul__`'s `str` overload was not copied, because unlike `Unit * "second"`, `Unit / "second"` raises `TypeError`.

`registry.Unit` gets the same overloads under `if TYPE_CHECKING` as `__mul__`, so that `/` also hands back a `Quantity` where `*` does. Without them `ureg.meter / 2` reads as `PlainQuantity[float]` while `ureg.meter * 2` reads as `Quantity[int]`.

`__pow__`, `__copy__` and `__deepcopy__` each build their result with `self.__class__(...)` but declare `PlainUnit`, so a subclass -- which is what every registry hands out -- is widened away to the plain base. `PlainQuantity.__copy__` and `__deepcopy__` already return `Self`.

---

- [x] Related issues/PRs:
  - #2302
  - #2303
  - #2404
  - #2243
  - #2247
  - #2373

---

This is what I did to get the typing errors in my codebase fixed and come up with a working upstreamable solution, but I'm otherwise very unfamiliar with this project, so I apologize if something in my PR breaks your customs.

### DIFF
--- a/pint/facets/plain/unit.py
+++ b/pint/facets/plain/unit.py
@@ -1,3 +1,5 @@
+# pyright: reportInvalidTypeArguments=warning
+
 """
 pint.facets.plain.unit
 ~~~~~~~~~~~~~~~~~~~~~
@@ -24,6 +26,7 @@ if TYPE_CHECKING:
     import datetime
 
     import numpy as np
+    import optype as opt
 
     from ..context import Context
     from .quantity import PlainQuantity
@@ -53,11 +56,11 @@ class PlainUnit(PrettyIPython, SharedRegistryObject):
                 )
             )
 
-    def __copy__(self) -> PlainUnit:
+    def __copy__(self) -> Self:
         ret = self.__class__(self._units)
         return ret
 
-    def __deepcopy__(self, memo) -> PlainUnit:
+    def __deepcopy__(self, memo) -> Self:
         ret = self.__class__(copy.deepcopy(self._units, memo))
         return ret
 
@@ -176,6 +179,21 @@ class PlainUnit(PrettyIPython, SharedRegistryObject):
 
     __rmul__ = __mul__
 
+    # PlainUnit / PlainUnit -> PlainUnit
+    @overload
+    def __truediv__(self, other: Self) -> Self: ...
+    # PlainUnit / timedelta -> PlainQuantity[float]
+    @overload
+    def __truediv__(
+        self, other: datetime.timedelta | np.timedelta64
+    ) -> PlainQuantity[float]: ...
+    # PlainUnit / <Magnitude> or PlainQuantity[<Magnitude>]
+    #   -> PlainQuantity[type of 1 / <Magnitude>]
+    @overload
+    def __truediv__[U: Magnitude](
+        self,
+        other: PlainQuantity[opt.CanRTruediv[int, U]] | opt.CanRTruediv[int, U],
+    ) -> PlainQuantity[U]: ...
     def __truediv__(self, other):
         if self._check(other):
             if isinstance(other, self.__class__):
@@ -200,7 +218,7 @@ class PlainUnit(PrettyIPython, SharedRegistryObject):
     __div__ = __truediv__
     __rdiv__ = __rtruediv__
 
-    def __pow__(self, other) -> PlainUnit:
+    def __pow__(self, other) -> Self:
         if isinstance(other, NUMERIC_TYPES):
             return self.__class__(self._units**other)
 
--- a/CHANGES
+++ b/CHANGES
@@ -22,6 +22,7 @@ Pint Changelog
   initializing a `Quantity` with a `datetime.timedelta` value.  (#1978)
 - Improve type annotations for `Quantity` (#2302, #2303)
 - Improve type annotations for `Quantity`'s `__new__` and document support for arithmetic with `datetime.timedelta` introduced in #2348 at the typing level (#2373)
+- Improve type annotations for `Unit`: `/` and `**` keep the unit type, and `__copy__`/`__deepcopy__` keep the registry subclass
 
 ### Formatting changes
 
--- a/pint/registry.py
+++ b/pint/registry.py
@@ -438,6 +438,22 @@ class Unit(
 
         __rmul__ = __mul__
 
+        # Unit / Unit -> Unit
+        @overload
+        def __truediv__(self, other: Self) -> Self: ...
+        # Unit / timedelta -> Quantity[float]
+        @overload
+        def __truediv__(
+            self, other: datetime.timedelta | np.timedelta64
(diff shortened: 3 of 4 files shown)


pint/facets/plain/unit.py
```diff
@@ -176,6 +177,22 @@ def __mul__(self, other):
...
+    @overload
+    def __truediv__(self, other: Self) -> Self: ...
+    # PlainUnit / timedelta -> PlainQuantity[float]
+    @overload
+    def __truediv__(
+        self, other: datetime.timedelta | np.timedelta64
+    ) -> PlainQuantity[float]: ...
+    # PlainUnit / PlainQuantity -> PlainQuantity
+    @overload
+    def __truediv__(self, other: PlainQuantity[Any]) -> PlainQuantity[Any]: ...
```
RBerga06: To your knowledge, is the quantity's type preserved here?

In that case, it would be ideal to type this overload as `def __truediv__[Q: Quantity](self, other: Q) -> Q`.

Otherwise, maybe the runtime behaviour is better described by `def __truediv__[U: MagnitudeT](self, other: PlainQuantity[opt.CanRTruediv[int, U]]) -> PlainQuantity[U]` or similar? (if *this* is the case, you can probably even merge this overload with the next one) Note that ty won't like `PlainQuantity[opt.CanRTruediv[int, U]]` – the correct signature would be something like `PlainQuantity[Intersection[Magnitude, opt.CanRTruediv[int, U]]`, but unfortunately only ty has `Intersection` and the latter won't work with any other type checker.

zyv: Hey, thank you very much for looking into it! Unfortunately, your questions reach far beyond my typing-foo, so I leaned on an LLM-assisted probe, which resulted into the following table:

> To your knowledge, is the quantity's type preserved here?

No, you are right, the magnitude is not preserved at runtime...

| `ureg.meter / q`                  | magnitude of `q` | result             |
|-----------------------------------|------------------|--------------------|
| `Quantity(2, "m")`                | `int`            | `0.5` `float`      |
| `Quantity(2.0, "m")`              | `float`          | `0.5` `float`      |
| `Quantity(2+0j, "m")`             | `complex`        | `complex`          |
| `Quantity(Fraction(2), "m")`      | `Fraction`       | `0.5` `float`      |
| `Quantity(np.array([1, 2]), "m")` | `ndarray[int64]` | `ndarray[float64]` |

> (if this is the case, you can probably even merge this overload with the next one)

... so I have merged this overload with the next one as you suggested and updated the PR.

pyright now reads `Unit / Quantity[int | float | complex]` as `Quantity[float | float | complex]`, where `PlainQuantity[Any]` threw the magnitude away. It's still not perfect, however:

- `CanRTruediv[int, U]` claims `Fraction` for `ureg.meter / Fraction(2)` (runtime gives `float`) and `Decimal` for `ureg.meter / Decimal(2)` (runtime raises `TypeError`)
- `CanRTruediv[float, U]` describes both correctly, but then `ureg.meter / 2` stops matching, since `int.__rtruediv__` takes an `int`

Not sure whether this is "good enough" or not.

RBerga06: That's strange, I would've expected, naively, that `<Unit> / <Fraction>` could retain `Fraction` as a magnitude, pretty much like if one had written `<Unit> * (1 / <Fraction>)`... I'm not a maintainer so I don't have an extensive knowledge, but I'll open an issue to clarify what's the intended runtime behaviour here.

Until that's resolved, I think what you chose here is ok. Again, I'm not a maintainer, so I won't be able to approve or merge your PR, but I wanted to have a look at it nevertheless because I had annotated `[Plain]Unit.__mul__` and I was curious about other potential approaches to the same problem.

**Edit**: opened #2413

zyv: > That's strange, I would've expected, naively, that `<Unit> / <Fraction>` could retain `Fraction` as a magnitude, pretty much like if one had written `<Unit> * (1 / <Fraction>)`... I'm not a maintainer so I don't have an extensive knowledge, but I'll open an issue to clarify what's the intended runtime behaviour here.

Thanks for the follow-up and taking care of this! If I understood you correctly, what you mean is that either the runtime can be adjusted to match the annotation as currently suggested in this PR in terms of Fraction handling (`CanRTruediv[int, U]`), or the annotation can be brought even closer to runtime behavior if it turns out that collapsing Fraction to float is preferable (extend my attempt with `CanRTruediv[float, U]`).

> Until that's resolved, I think what you chose here is ok. Again, I'm not a maintainer, so I won't be able to approve or merge your PR, but I wanted to have a look at it nevertheless because I had annotated `[Plain]Unit.__mul__` and I was curious about other potential approaches to the same problem.

I figured out that much, since I initially copy & pasted your `__mul__` annotations trying to get them to work for `__truediv__` :-) As I've already mentioned, I'm using Pint only very superficially, and have a poor understanding of the intricacies of the typing system, so all I wanted was to get the pesky red underlines to go away for me. Therefore, it's great that someone who has the knowledge and stakes in this took the time to look at my suggestion - thank you for that.

andrewgsavage:
is this ready for merging?

we can do the unit/fraction behaviour in another PR

zyv:
> is this ready for merging?
> 
> we can do the unit/fraction behaviour in another PR

I understand from your comment in #2413 that you concur with @RBerga06 that it would be nicer to keep Fractions. In this case, the annotations I'm offering here after integrating the suggestion by @RBerga06 already reflect this (future) behavior.

If you're planning to fix #2413 in the next development cycle anyways, I think the PR is ready for merge. If not, then I'm afraid I can't judge whether it's okay or not that the suggested annotations "lie" about current runtime behavior w.r.t. Fraction - I guess it's still better than the state before the PR, but that's my opinion.

RBerga06:
> If you're planning to fix #2413 in the next development cycle anyways, I think the PR is ready for merge.

I think I can work on it this week. **Edit**: opened PR #2427.

andrewgsavage:
nice one thanks

zyv:
> nice one thanks

Thank you for maintaining pint!

Issue #2306, opened 2026-05-09
UnitRegistry typing


andrewgsavage:
Following from https://github.com/hgrecco/pint/pull/2303 , there was some discussion on UnitRegistry typing. This issue is to discuss how best to type UnitRegistry.

>> However, I don't understand why does UnitRegistry need to be generic over MagnitudeT: Magnitude. Isn't the same UnitRegistry used for all associated quantities?

> What do you mean by this? It is generic over `Quantity` and `Unit`

Ah yes, sorry, that's right. What I meant is, we technically wouldn't need unit registries to be generic over the complete (fully specialized) quantity type, because with the same registry one can construct different quantities with different magnitude types (e.g. `ureg.Quantity(32, "m") -> Quantity[int]`, `ureg.Quantity(np.array([3.2, 3.4, 3.5])) -> Quantity[np.ndarray[...]]`, etc.).

Because of the `ureg.Quantity()` constructor, what we'd *actually* need to be generic over is a higher kinded type `QuantityT[~MagnitudeT: Magnitude]: PlainQuantity[MagnitudeT]` (`~` meaning invariant). Sadly, we don't have HKTs in Python yet (https://github.com/python/typing/issues/548), so that doesn't look feasible for now.

The problem with the current approach is that the type aliases `GenericUnitRegistry[QuantityT, UnitT].Quantity = QuantityT` are not generic, which means that, for example, `SystemRegistry.Quantity = SystemRegistry[SystemQuantity[Any], SystemUnit].Quantity = SystemQuantity[Any]` is considered (at least by pyright) as already specialized, which triggers a fair amount of errors in the `class Quantity[MagnitudeT: Magnitude](facets.SystemRegistry.Quantity[MagnitudeT], ...)` definition.

**Edit**: In fact, pyright reports an error in the `test_quantity.py`:`TestGenericQuantityTyping.test_generic_type_use` test.

I think the simplest approach would be to switch to generic type aliases, like:
```python
# pint/facets/plain/registry.py
class GenericUnitRegistry[UnitT: PlainUnit]:
    type Quantity[MagnitudeT] = PlainQuantity[MagnitudeT]

    def parse_expression(...) -> Quantity[Any]
```
This unfortunately requires re-annotating `parse_expression` in subclasses (in a potential bright future, the [associated types proposal](https://discuss.python.org/t/associated-types/105401/6) I saw on Discourse would fix that).

[...]

PR #2407, opened 2026-09-03
Give `UnitRegistry`'s magnitude parameter a default

### LINKED ISSUE #2245: Typing regression in 0.25.1
Just tried upgrading to the new 0.25.1 version and got several dozens typing errors. Minimal example, basically the first few lines of the tutorial: 
```python
from pint import UnitRegistry

ureg = UnitRegistry()
reveal_type(ureg.meter)
distance = 24.0 * ureg.meter
reveal_type(distance)
reveal_type(24.0 * ureg.meter)
```

- 0.25.0:
```
$ mypy test_pint.py

test_pint.py:4: note: Revealed type is "pint.registry.Unit"
test_pint.py:6: note: Revealed type is "Any"
test_pint.py:7: note: Revealed type is "Any"
Success: no issues found in 1 source file
```
- 0.25.1:
```
$ mypy test_pint.py

test_pint.py:3: error: Need type annotation for "ureg"  [var-annotated]
test_pint.py:4: note: Revealed type is "pint.registry.Unit"
test_pint.py:5: error: Need type annotation for "distance"  [var-annotated]
test_pint.py:6: note: Revealed type is "Any"
test_pint.py:7: note: Revealed type is "Never"
Found 2 errors in 1 file (checked 1 source file)
```

Annotating the first line as `ureg: UnitRegistry[float] = UnitRegistry()` and the third as `distance: Quantity[float] = 24.0 * ureg.meter` fixes the two reported errors but the final expression is still inferred as "Never".

This looks like a serious regression, in a patch version no less; it basically renders pint unusable in any project using mypy (didn't check other typecheckers but I expect similar).

zyv:
`ureg = UnitRegistry()` leaves `MagnitudeT` unsolved, so mypy asks for an annotation on the first line of the tutorial. The 0.25.1 revert cleared the second error of the #2245 reproducer, but not this one, which is still there on 0.25.3 and 0.26.0rc0.

`MagnitudeT_co` in the same file is already declared with `default="Any"`, so this patch spells the registry's parameter the same way. A PEP 696 default in the PEP 695 parameter list would need Python 3.13.

- [x] Closes #2245, relates to #2306

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@ Pint Changelog
 -----------------
 
 - Fix `UnitRegistry(system="imperial")` returning US units instead of imperial ones for shared names like `gallon` or `pint` (#2408)
+- Give `UnitRegistry`'s magnitude type parameter a default, so `ureg = UnitRegistry()` needs no annotation (#2245)
 
 
 0.26.0 (2026-09-10)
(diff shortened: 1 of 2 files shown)


andrewgsavage:
nice one thanks

zyv:
> nice one thanks

Thanks for maintaining Pint!

PR #2410, opened 2026-09-06
Fix np.sum/np.nansum not converting a Quantity initial value

### LINKED ISSUE #2400: [bug] np.sum ignores the units of a Quantity `initial`
`numpy.sum` and `numpy.nansum` strip a Quantity `initial` without converting it,
so it is added as a bare magnitude in the input's units. A compatible unit gives
a silently wrong number, and an incompatible one is accepted instead of raising.

```python
import numpy as np
import pint

u = pint.UnitRegistry()
q = [1, 2] * u.m

print(np.sum(q, initial=100 * u.cm))  # 103 meter, expected 4 meter
print(np.sum(q, initial=1 * u.s))     # 4 meter, expected DimensionalityError
```

The positional form, `np.sum(q, None, None, None, False, 100 * u.cm)`, behaves
the same way.

`np.max` and `np.amax` already convert `initial`, so `np.max(q, initial=100 * u.cm)`
is correct and only the `sum` family is affected. They are registered through
`implement_consistent_units_by_argument` with `initial` named as a unit argument,
while `sum` and `nansum` go through `implement_func` with `input_units=None`,
which strips every argument without converting any of them.

Tested on 0.25.3 and on master (6fc0533) with numpy 2.5.2.

I can open a PR if that is helpful.

afonsojanu:
Closes #2400.

`np.max`/`np.min` (and `np.amax`/`np.amin`) already convert a Quantity `initial` argument to the input array's units before running the reduction, so `np.max(q, initial=100 * ureg.cm)` works correctly. `np.sum` and `np.nansum` go through a different registration path that strips units without converting anything, so the same call silently treats `initial` as a bare magnitude in whatever unit it happens to carry:

```python
q = [1, 2] * ureg.m
np.sum(q, initial=100 * ureg.cm)   # 103 meter, should be 4 meter
np.sum(q, initial=1 * ureg.s)      # 4 meter, should raise DimensionalityError
```

The fix converts a Quantity `initial` kwarg to the array's units right where `first_input_units` is already computed, mirroring what the max/min path does, and only touches this one narrow case (functions registered with `input_units=None`, gated on `initial` actually being a Quantity kwarg). Everything else that goes through the same generic dispatch keeps working exactly as before, verified by running the full test suite with and without the change.

As a smaller, known follow-up: the positional form mentioned in the issue (`np.sum(q, None, None, None, False, 100 * ureg.cm)`) isn't covered by this fix, since that would need binding the call against `numpy.sum`'s actual signature rather than a narrow kwarg check, which felt like more surface area than this fix needed. Happy to look at that separately if it's wanted.

Added two regression tests (`test_sum_with_initial_arg`, `test_nansum_with_initial_arg`) that fail on master and pass with the fix.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -63,6 +63,7 @@ Pint Changelog
 - Fix `Unit.__eq__` when comparing with strings using unit aliases (e.g. `Unit('metre') == 'meter'`) (#634)
 - Fix `@alias` raising a bare `KeyError` on prefixed unit targets (#1076, #1447)
 - Fix `parse_units` cache never being hit for symbol and prefixed spellings (e.g. `kg`, `kWh`, `ms`), making those very common lookups re-parse on every call (#2320)
+- Fix `numpy.sum`/`numpy.nansum` treating a Quantity `initial` as a bare magnitude instead of converting it to the input's units, unlike `numpy.max`/`numpy.min` which already convert it correctly (#2400)
 - Fix NumPy ufunc mul/div paths allowing offset units to bypass `OffsetUnitCalculusError` (#1335)
 
 
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ Pint Changelog
 0.26.2 (unreleased)
 -------------------
 
-- No changes yet.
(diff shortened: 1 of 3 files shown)


andrewgsavage:
cheers

PR #2422, opened 2026-09-14
Fix wraps conversion for variadic positional arguments

### LINKED ISSUE #2254: ureg.wraps does not work with var-positional arguments

Doribelove:
Closes #2254.

`UnitRegistry.wraps()` currently treats a `*args` parameter as one positional value. Later values reach the wrapped function without conversion, empty calls can raise `KeyError`, and keyword-only parameters can be confused with variadic values.

For example, a sum wrapped with `ureg.wraps("=A", "=A")` cannot combine `1 * ureg.meter` and `200 * ureg.centimeter`. With this change it receives the magnitudes `1` and `2.0`, and returns `Quantity(3.0, "meter")`.

The converter now binds variadic calls using their Python signature, applies defaults, and maps each declared unit specification to its actual values. A reference defined by `*args` takes its unit from the first element; all subsequent elements are converted to it. An empty variadic reference is dimensionless, so wrapping `math.gcd` also supports `gcd()`. Fixed reference parameters and keyword-only reference anchors retain their units even when there are no variadic values.

Keyword-only arguments keep their own conversion slots, arbitrary `**kwargs` pass through, and the existing trailing `None` placeholder for `**kwargs` remains supported. Non-variadic calls keep their existing argument-packing path. The documentation explains the empty-reference behavior and includes executable examples. It also corrects an existing strict-mode error example to include the already-supported string inputs.

Related prior work: #2336 handles `**kwargs`, whereas #2326 was closed under the assumption that #2336 might also address `*args`. The reporter subsequently confirmed that #2254 still reproduced. This change covers the remaining positional-variadic behavior, including mixed signatures.

### Validation

Base: `e1efeda32ea2066d4a385f48e7a2aea949f8118c`, Python 3.12.14, NumPy 2.5.3 and SciPy 1.18.1 on Linux x86_64.

- New public-API regressions against unchanged master: **29 failed, 13 passed**. With the fix: **42 passed**.
- Full local suite (`pytest --benchmark-skip`): **2,461 passed, 471 skipped, 12 xfailed**, 298 subtests passed. The unchanged baseline has 2,419 passes with the same skips, expected failures, subtests and 46 warnings.
- All **38 examples** in the wrapping documentation pass with its declared test setup.
- All repository pre-commit hooks pass via `pixi run -e lint lint --all-files`; `git diff --check` passes. An independent 384-case signature/argument matrix also passes.
- Wheel and source distribution builds pass.

The regressions include incompatible units, strict strings and bare values, reference resets between calls, empty input, positional-only and keyword-only parameters, defaults, extra keywords, arrays, multiple returns, and signature metadata. Different-unit conversions are not forced to produce integers; `math.gcd` tests use same-unit integer magnitudes.

Optional dependencies not installed for the local suite and the full Python/platform matrix are left to CI. Local fixed-argument timing samples range from about 1.8% faster to 7.4% slower; the latter is 0.056 microseconds per call. These are small local samples, not a performance guarantee.

- [x] Closes #2254
- [x] The change is covered by automated unit tests

AI assistance: the implementation, tests, documentation and this description were prepared with Codex. Validation commands were executed locally.


pint/registry_helpers.py
```diff
@@ -117,56 +117,106 @@ def _parse_wrap_args(args, registry=None):
...
-                _replace_units(args_as_uc[ndx][0], values_by_name),
-            )
+            for pos in indices[ndx]:
+                value = values[pos]
+                assert _replace_units(args_as_uc[ndx][0], values_by_name) is not None
+                values[pos] = ureg._convert(
+                    getattr(value, "_magnitude", value),
+                    getattr(value, "_units", UnitsContainer({})),
+                    _replace_units(args_as_uc[ndx][0], values_by_name),
+                )
```
matthiasschabel: The dependent target is constant across this parameter’s variadic elements. Resolve it once rather than twice per element. Keep empty groups skipped, preserving the current behavior.

```suggestion
            positions = indices[ndx]
            if not positions:
                continue
            target_units = _replace_units(args_as_uc[ndx][0], values_by_name)
            assert target_units is not None
            for pos in positions:
                value = values[pos]
                values[pos] = ureg._convert(
                    getattr(value, "_magnitude", value),
                    getattr(value, "_units", UnitsContainer({})),
                    target_units,
                )
```

pint/testsuite/test_wraps_varargs.py
```diff
@@ -0,0 +1,287 @@
...
+        return sum(areas)
+
+    values = [ureg.Quantity(10000, "cm**2"), ureg.Quantity(2, "m**2")][:count]
+    result = total(*values, length=ureg.Quantity(3, "m"))
+
+    assert result.units == ureg.meter**2
+    assert result.magnitude == sum([1, 2][:count])
+
+
+def test_none_varargs_between_converted_parameters(ureg):
```
matthiasschabel, line 213: Could we also pin down the complementary empty-reference case? When *args defines A and is empty, A is dimensionless; a later keyword-only dependency must not redefine it. The implementation already does this correctly.

```suggestion
def test_empty_varargs_reference_does_not_rebind_to_keyword_dependency(ureg):
    @ureg.wraps(None, ("=A", "=A"))
    def collect(*values, tail):
        return values, tail

    with pytest.raises(DimensionalityError):
        collect(tail=ureg.Quantity(1, "m"))


def test_none_varargs_between_converted_parameters(ureg):
```

matthiasschabel (commented):
This PR correctly fixes the issue AFAICT. Codex identifies two optional refinements below. With both applied locally, test_wraps_varargs.py and test_unit.py pass: 187 passed, 1 xfailed, 38 subtests passed (Python 3.12.12, NumPy 2.5.3, macOS arm64).

Doribelove:
Thanks for reviewing and testing this. Both refinements are now in 8714d6a6985451abbf27c20dc0dc62c800840be7.

The dependent target is resolved once per nonempty parameter group, with empty groups still skipped. The added regression checks that an empty `*args` defining `A` remains dimensionless and rejects a unit-bearing keyword-only dependency instead of rebinding `A`.

Validation on Linux with Python 3.12.14:

- `test_wraps_varargs.py` and `test_unit.py`: 187 passed, 1 xfailed, 38 subtests passed.
- Independent signature/argument matrix: all 384 cases passed.
- All seven repository pre-commit hooks and `git diff --check`: passed.

These follow-up changes and checks were prepared with Codex. The PR is ready for another look.

Doribelove:
Merged current upstream master (`e4042bb`) in `f4c3bcd` to bring the review branch up to date. Both refinements from the review remain intact.

Revalidated the merged tree: `test_wraps_varargs.py` and `test_unit.py` report 187 passed, 1 xfailed and 38 passing subtests; the independent 384-case signature matrix and all seven repository pre-commit hooks pass. GitHub now reports the PR as mergeable. The new ReadTheDocs build is pending.

Doribelove:
@matthiasschabel, both refinements from your review are in the current branch (`f4c3bcd`, updated onto upstream `e4042bb`), and the focused tests, independent signature matrix, and pre-commit checks pass. Read the Docs has now passed as well. I updated the PR description to distinguish the original full-suite run from validation on the current head.

Could you take another look when convenient? The three GitHub Actions workflows are still marked `action_required` and need approval from someone with repository access before remote tests can run. Thank you.

matthiasschabel:
> @matthiasschabel, both refinements 
> ...
> Could you take another look when convenient? The three GitHub Actions workflows are still marked `action_required` and need approval from someone with repository access before remote tests can run. Thank you.

@Doribelove unfortunately, I'm not a maintainer. Maybe @andrewgsavage can have a look?

Issue #2346, opened 2026-07-14
Integer quantities support status


RBerga06:
The docs [mention](https://pint.readthedocs.io/en/stable/advanced/typing.html) that it makes sense to write `Quantity[int]` (or `Quantity[np.ndarray[<shape>, <integer dtype>]]`).

However, I think this poses some problems, because then a lot of operations can break this type invariance: for example, `Quantity(3, 'cm').to_root_units()` returns `Quantity(0.03, "meter")`.

The real problem comes when we consider inplace operations:

```python
def foo(x: Quantity[int]):
    reveal_type(x.magnitude)  # int (good)
	x.ito_root_units()
	reveal_type(x.magnitude)  # type checker thinks it's int, but might be float at runtime
```

I don't think there's any way of telling type checkers that an in-place operation might've changed the type of a variable.

Is this expected behaviour? Should the docs be updated with a note about this?

---

**Note:** since type checkers think `int <: float` ([and it's still an open problem](https://github.com/python/typing/issues/1746)), as long as `MagnitudeT` (in `Quantity`'s definition) remains covariant (which simplifies a lot arithmetic method overloads) it's not possible to make the `Quantity[int]` annotation an error – the best we can do is document it.

However, in the case of numpy arrays, integer and floating dtypes are not in a subtype-supertype relationship, which means we can simply accept `floating` and `complex` arrays while the boolean/integer ones will be disallowed by default in the annotations.

---

**Edit:** apparently `x.ito_root_units()` **fails at runtime** if the magnitude is an integer array. The error is:
```numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'multiply' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'```.
This happens because we're essentially doing `self.magnitude *= <float>` and numpy correctly raises because this operation cannot preserve the LHS type.

However, `x.to_root_units()` *works correctly* and returns a float array (following the [data type promotion rules](https://numpy.org/doc/stable/reference/arrays.promotion.html#numerical-promotion)).

andrewgsavage:
I hadn't thought of this but it's a great point.

Adding a note or recommendation to avoid using int would be helpful

Could you expand on the numpy case with an example?

RBerga06:
See the numpy dtype hierarchy [here](https://numpy.org/doc/stable/reference/arrays.scalars.html#id4).
[...]

PR #2423, opened 2026-09-14
Fix `**=` changing units when the in-place magnitude update fails

matthiasschabel:

`PlainQuantity.__ipow__` sets the new units before it raises the magnitude to the power in place. If that step fails, the quantity ends up with its old magnitude and the new units:

```python
import numpy as np
import pint

u = pint.UnitRegistry()
q = u.Quantity(np.array([4, 9]), "meter")
try:
    q **= 0.5  # UFuncTypeError: can't cast float64 output to int64
except TypeError:
    pass
print(q)  # [4 9] meter ** 0.5, expected [4 9] meter
```

This computes the new units first and assigns them only once the magnitude update succeeds. NumPy's error is still raised. `*=`, `/=`, `+=` and `ito` already leave the quantity unchanged on the same failure.


RBerga06:
Related to #2346.

Issue #1505, opened 2022-04-08
Decimal registry parses unit expression without scaling factor using float instead of decimal


Flix6x:
I found the following unexpected difference between float type and decimal type registries:

```
import decimal
import pint

ur = pint.UnitRegistry(non_int_type=decimal.Decimal)

pint.Quantity("m/s").to("km/h")  # <Quantity(3.6, 'kilometer / hour')>
ur.Quantity("m/s").to("km/h")  # fails with TypeError: unsupported operand type(s) for *: 'float' and 'decimal.Decimal'
```
If I add an explicit scaling factor, it works as expected:
```
pint.Quantity("1m/s").to("km/h")  # <Quantity(3.6, 'kilometer / hour')>
ur.Quantity("1m/s").to("km/h")  # <Quantity(3.60000, 'kilometer / hour')>
```

I think that underlying this there's a problem with the parsing of the input string, but I could use some help figuring out where the float is being introduced.

Here's what a relevant test could look like:
```
def test_decimal_magnitude():
    ur = pint.UnitRegistry(non_int_type=decimal.Decimal)

    assert isinstance(ur.Quantity("1m/s").magnitude, decimal.Decimal)
    assert not isinstance(ur.Quantity("m/s").magnitude, float)  # unexpected success (magnitude should not be a float)
    assert isinstance(ur.Quantity("m/s").magnitude, decimal.Decimal)  # unexpected fail (magnitude should be a decimal)
```

Both `pint==0.18` and `pint==0.19.1` have this problem.

Flix6x:
`pint==0.20.1` still has this problem. Does anyone know where the float is introduced? I'd love to solve this issue.

To clarify:
```
>>> ur.Quantity("m/s").magnitude
1.0
>>> type(ur.Quantity("m/s").magnitude)
float
```
Above, the float is unexpected, given that we set up the `ur` unit registry to work with decimals. Adding an explicit magnitude works as expected:
```
>>> type(ur.Quantity("1m/s").magnitude)
decimal.Decimal
>>> type(ur.Quantity("1.0m/s").magnitude)
decimal.Decimal
```

dalito:
I wonder if the root of the problem is the same as here https://github.com/hgrecco/pint/issues/1735

Flix6x:
[...]

PR #2427, opened 2026-09-15
More robust arithmetic operations for `Quantity` and `Unit`

### LINKED ISSUE #2413: Clarifying `Unit.__truediv__`'s semantics
Hi everyone, thanks for this great project!

Since I had annotated `Unit.__mul__`, when I saw a PR annotating `Unit.__truediv__` (#2406 by @zyv) I decided to have a look at it. The PR author has pointed out that, at runtime, `Unit / Fraction -> Quantity[float]` and not `Quantity[Fraction]`, which to me seems a bit counter-intuitive.

In fact, I would naively expect that, in general, given `u: Unit` and `x` a numeric scalar (e.g. `float`, `Fraction`, `np.float64`, `optype.numpy.NDArray[np.float64, ...]`), the expression `u / x` would be equivalent to `u * (1 / x)`.

Hence, if `x` is `Fraction`, `1 / x` is also a `Fraction` which in turn would mean `u / x` should be `Quantity[Fraction]`. The good thing about keeping `Fraction`s around is that they are always exact (i.e. they represent their value exactly) and this `u / x` operation needlessly gives up this invariant.

**Side note**: the current behaviour is consistent with `Quantity(1, u) / x`, which unfortunately also returns `Quantity[float]`. In #2346 I argued treating `Quantity(1, u)` as `Quantity(1.0, u)` could be a good thing, because `Quantity` allows in-place operations. However, the case of `u / x` is different, because it's a single operation and there is no way for the user to call `.ito_root_units()` between the creation of the `Quantity` and the division itself: I think this justifies preferring the `u * (1/x)` behaviour over the status quo (`(u*1)/x`).

I'd like to know if this distinction was already discussed (I couldn't find an issue for this), and whether the status quo is preferred for technical reasons.

In case a decision is made, I am willing to open a PR describing it in the docs and, if necessary, updating runtime behaviour and type annotations as well (although I might have time to work on it only next week).

RBerga06:
### Additional changes
- Made the behaviour of `__mul__`, `__truediv__`, `__rtruediv__` (in both `PlainQuantity` and `PlainUnit`) more consistent and resilient;
- Improved type annotations for `PlainUnit.__mul__`, `PlainUnit.__truediv__`, `PlainUnit.__rtruediv__` (and corresponding `Unit` methods);
- Refactored `pint.compat` removing two confusing/redundant (public but undocumented) symbols (`check_upcast_type` in favour of `is_upcast_type` and `upcast_type_names` in favour of `upcast_type_map`);
- Improved type annotation for `UnitRegistry.__init__`'s `file` parameter;
- Removed some code duplication in `PlainQuantity.__new__` and `PlainRegistry.__new__` by moving the common unit conversion code into a new `GenericPlainRegistry` method. For consistency, also moved the magnitude conversion code from `PlainQuantity.__new__` into `GenericPlainRegistry` (I have left ideas for additional refactoring in the code as `# TODO` comments, although some of them depend on the answer to a question I've left in a comment below);
- Improved a handful of docstrings and added more comments throughout the code while I was reading it to understand it better (and help future contributors);

### Notes
~Depends on #2406~ (has been merged).

---
- [x] Closes #2413


RBerga06:
This is still missing type annotations but I'd like to wait for #2406 to be merged in order to avoid git conflicts - I'd like to very slightly refine those annotations, although I'm not sure about how to handle upcast types just yet (I don't think they're currently supported in annotations, for the time being, so this can probably be left for future PRs)

andrewgsavage:
> This is still missing type annotations but I'd like to wait for #2406 to be merged in order to avoid git conflicts - I'd like to very slightly refine those annotations, although I'm not sure about how to handle upcast types just yet (I don't think they're currently supported in annotations, for the time being, so this can probably be left for future PRs)

alright merged that for you

RBerga06:
> alright merged that for you

Thank you!

pint/registry.py
```diff
@@ -433,6 +435,9 @@ def __mul__(
         # Unit * <Magnitude> -> Quantity[<Magnitude>]
         @overload
         def __mul__[T: Magnitude](self, other: T) -> Quantity[T]: ...
+        # Unit * <PlainQuantity> -> <PlainQuantity>
+        @overload
+        def __mul__[Q: _PlainQuantity](self, other: Q) -> Q: ...
```
andrewgsavage: Is this needed to be PlainQuantity and not Quantity?

andrewgsavage:
I had claude review it and it found some issues, these are the ones worth addressing. The fix at the end looks worth looking into.
## 1. `Unit / <list or tuple>` now raises `TypeError` instead of returning an array `Quantity`

**`pint/facets/plain/unit.py`, line 214** (`PlainUnit.__truediv__`)

The new fallback computes `1 / other` in plain Python *before* wrapping the result in `Quantity`, bypassing `_to_magnitude`'s coercion. `_to_magnitude` (`pint/compat.py`) unconditionally converts `list`/`tuple` to `np.asarray(...)` â€” the old code always routed through `self._REGISTRY.Quantity(1, self._units) / other`, which reached that coercion via `Quantity.__truediv__` â†’ `_mul_div` â†’ `_to_magnitude`. Lists/tuples aren't "upcast types" (`is_upcast_type`), so they now hit `1 / other` directly, and `int`/`list` don't support that operation.

**Repro (verified on both branches):**
```python
import pint
u = pint.UnitRegistry().Unit("m")
u / [1, 2, 3]
```
- On `master`: returns `Quantity([1.0, 0.5, 0.333...], 'meter')`
- On this branch: raises `TypeError: unsupported operand type(s) for /: 'int' and 'list'`

No test in `test_unit.py`/`test_quantity.py` covers `Unit / list` or `Unit / tuple`, so this slipped past CI.

---

## 2. `Unit / bool` is now silently mis-handled instead of raising

**`pint/facets/plain/unit.py`, line 214** (`PlainUnit.__truediv__`)

Same root cause as #1: the new fallback skips `_to_magnitude`'s explicit `bool` rejection (`isinstance(value, (dict, bool))` â†’ `TypeError`).

**Repro (verified on both branches):**
```python
import pint
u = pint.UnitRegistry().Unit("m")
u / True   # silently "succeeds" now
u / False  # raises a different, less informative error now
```
- On `master`: `u / True` raises `TypeError: unsupported operand type(s) for /: 'Quantity' and 'bool'`; `u / False` raises the same kind of `TypeError`.
- On this branch: `u / True` silently returns `Quantity(1.0, 'meter')` (since `1 / True == 1.0` in plain Python, and this happens before any magnitude validation); `u / False` raises `ZeroDivisionError: division by zero` instead.

This is inconsistent with the rest of pint, which still rejects `bool` as a magnitude everywhere else (`Quantity(True, "m")` and `u * True` both still raise `TypeError` on this branch).

---

## 3. `Unit / ""` now raises the wrong exception type

**`pint/facets/plain/unit.py`, line 214** (`PlainUnit.__truediv__`)

Same fallback, same root cause: `_to_magnitude` raises a specific `ValueError("Quantity magnitude cannot be an empty string.")` for an empty string, but the new code never reaches it.

**Repro (verified on both branches):**
```python
import pint
u = pint.UnitRegistry().Unit("m")
u / ""
```
- On `master`: raises `ValueError: Quantity magnitude cannot be an empty string.`
- On this branch: raises `TypeError: unsupported operand type(s) for /: 'int' and 'str'`

Any caller catching `ValueError` specifically for this case will no longer catch it.

---

**Suggested fix for all three:** rather than special-casing `is_upcast_type(type(other))` to decide whether to route through `Quantity(1, self._units) / other`, consider fixing the root cause of #2413 at the source â€” `PlainQuantity._truedivide_cast_int` (`pint/facets/plain/quantity.py`), which casts an `int` magnitude to `float`/`non_int_type` before dividing regardless of what `other` is. A fix there (e.g. skip the intâ†’float cast when `other` is `Fraction`/`Decimal`) would fix `Unit / Fraction` *and* `Quantity / Fraction` consistently, without changing behavior for `list`/`tuple`/`bool`/`""` at all. (Worth noting: as currently written, `Quantity(1, "m") / Fraction(3,4)` still loses precision and `Quantity(1, "m") / Decimal("0.1")` still raises `TypeError` on this branch â€” the fix only reaches the `Unit` path, not `Quantity`.)

RBerga06:
Thanks for the review.

Yes, I clearly didn't think about the different conversions the `Quantity` constructor does behind the curtain, in particular list -> numpy array (but I'm sure it's not the only one that needs to be taken care of here). I explicitely handled the typecast-type special case because there was a comment pointing to it as the specific reason behind why the implementation was `Quantity(1, self) / x` instead of `Quantity(1/x, self)` - I guess that comment was maybe somewhat outdated.

I'll try and see what's the best option here in terms of implementation; Claude's suggestion looks sensible, although I think maybe something a little more general might be useful here.

I guess it's clear we also need a lot more tests...

RBerga06:
I have a question, though, about the role of `PlainQuantity` vs `NumpyQuantity`.

To me, it looks like `NumpyQuantity` is there simply to provide support for numpy protocols (i.e. it implements dunder methods for numpy support). However, conversion to numpy arrays and similar things should sit in `PlainQuantity`, right?

I'm asking because I've just noticed that `np.timedelta64 -> Quantity` conversions are implemented in `NumpyQuantity` instead...

**Edit**: Nope, `_to_magnitude` is implemented in `pint.compat`, while `to_list()` and `from_list(...)` in `PlainQuantity`.

RBerga06:
Also, regarding 3, it's very odd that pint checks if the rhs is an empty string, but still rejects every other string anyway in the operation.

Also, the results & error messages one gets in all the other cases are a little strange:
```python
# tested on current master branch
>>> from pint import Quantity

>>> Quantity(3, "m") + "2 m"  # same with "2 s", "2", "s" or ""
>>> Quantity(3, "m") - "2 m"  # same with "2 s", "2", "s" or ""
>>> Quantity(3) + "2"
TypeError: unsupported operand type(s) for +: 'int' and 'str'

>>> Quantity(3, "m") * ""
ValueError: Quantity magnitude cannot be an empty string.
>>> Quantity(3, "m") * "1"
Quantity(111, "meter")  # note the missing ''s here, magnitude is an integer
>>> Quantity(3, "m") * "s"
Quantity('sss', "meter")
>>> Quantity(3, "m") * "2 s"
Quantity('2 s2 s2 s', "meter")

>>> Quantity(3, "m") / ""
ValueError: Quantity magnitude cannot be an empty string.
>>> Quantity(3, "m") / "2 s"
TypeError: unsupported operand type(s) for /: 'float' and 'str'
```
What's the intended behaviour here? Is pint supposed to parse and autoconvert strings as well in arithmetic operations, or not?

RBerga06:
Given the discussion in #1505, I think the correct check was not `isinstance(self._magnitude, int) or isinstance(other_magnitude, int)` but instead `isinstance(self._magnitude, int) and isinstance(other_magnitude, int)` (this is pseudocode, you can see the removed lines in the diff): this seems to nicely fix division between quantity and other types (Fraction, Decimal, etc.) while preserving all of the conversions.

I've taken the liberty to move the customized `truediv` (previously known as `_truedivide_cast_int`) next to the definition of `non_int_type`, in the hope it improves maintenance.

RBerga06:
I'm sorry, this PR has come out quite big. Most of it is actually just doc improvements, type annotations and tests, then there's a bit of refactoring and the important runtime change is the one I mentioned in the previous comment.

One last thing I don't understand is why does `Unit.__mul__` special-case `other` being a `Number`, and it having value 1 (even before potential conversions, e.g. `_to_magnitude`, can be applied). Is it for performance reasons?

